PR #1, opened 2022-02-20
add distance metrics

cigrainger:
This:
- Adds `ex_doc` with math support
- Creates a `Scholar.Spatial` module which should be similar to `scipy.spatial`
- Drops required Elixir to 1.13
- Adds some simple distance metrics in `Scholar.Spatial.Distances` (which should be similar to `scipy.spatial.distances`
- Steals the `assert_shape!` helper from `Axon.Shared` and puts it in `Scholar.Shared`

I don't think we should copy `scipy` all the time re: organisation of code but in this case i think they've got this spot on. An alternative is what `sklearn` does, which is to put distance functions in `metrics`. I think that makes less sense, and it's because the focus there is on `pairwise` (see [here](https://scikit-learn.org/stable/modules/metrics.html#metrics)).

I need to double check these before we merge, but given that this first PR brings a fair amount of baggage with it I thought I'd put it up and start the discussion.


lib/scholar/spatial/distances.ex
```diff
@@ -0,0 +1,271 @@
...
+      iex> Scholar.Spatial.Distances.manhattan(x, y)
+      ** (ArgumentError) expected input shapes to be equal, got {2} != {3}
+  """
+  @spec manhattan(Nx.t(), Nx.t()) :: Nx.t()
+  defn manhattan(x, y) do
+    assert_shape!(x, y)
+
+    x
+    |> Nx.subtract(y)
+    |> Nx.abs()
```
polvalente, line 132: We have Nx.LinAlg.norm(x, 1) or something like that

lib/scholar/spatial/distances.ex
```diff
@@ -0,0 +1,271 @@
...
+        0.0
+      >
+
+      iex> x = Nx.tensor([1, 2])
+      iex> y = Nx.tensor([1, 2, 3])
+      iex> Scholar.Spatial.Distances.chebyshev(x, y)
+      ** (ArgumentError) expected input shapes to be equal, got {2} != {3}
+  """
+  @spec chebyshev(Nx.t(), Nx.t()) :: Nx.t()
+  defn chebyshev(x, y) do
```
polvalente, line 168: There's also a variant of Nx.LinAlg.norm for this and the one below IIRC

lib/scholar/spatial/distances.ex
```diff
@@ -0,0 +1,271 @@
...
+        0.0
+
+      norm_x == 0.0 or norm_y == 0.0 ->
+        1.0
+
+      true ->
+        numerator = Nx.dot(x, y)
+        denominator = norm_x |> Nx.dot(norm_y)
+        result = 1.0 - numerator / denominator
+        Nx.as_type(result, {:f, 32})
```
polvalente, line 268: I don't think this explicit cast is needed because 1.0 above should force the cast to float already

lib/scholar/spatial/distances.ex
```diff
@@ -0,0 +1,271 @@
...
+    cond do
+      norm_x == 0.0 and norm_y == 0.0 ->
+        0.0
+
+      norm_x == 0.0 or norm_y == 0.0 ->
+        1.0
+
+      true ->
+        numerator = Nx.dot(x, y)
+        denominator = norm_x |> Nx.dot(norm_y)
```
polvalente, line 266:
```suggestion
        denominator = norm_x  * norm_y
```
Aren't those scalars?

lib/scholar/spatial/distances.ex
cigrainger: I actually don't think this is a vector norm?

lib/scholar/spatial/distances.ex
cigrainger: Yep, `Nx.LinAlg.norm(ord: :inf)` works here.

lib/scholar/spatial/distances.ex
polvalente: It's the L1 norm

polvalente: > I actually don't think this is a vector norm?

Could you elaborate on this?

cigrainger: Sorry I got confused by where the review comment was. :facepalm:

cigrainger: Replacing `Nx.abs()` and `Nx.sum()` with `Nx.LinAlg.norm(ord: 1)` is throwing for me though when `x == y`. I'm guessing it's a divide by zero error but opaque:

```elixir
     stacktrace:
       (nx 0.1.0) lib/nx/binary_backend.ex:654: Nx.BinaryBackend.element_divide/3
       (nx 0.1.0) lib/nx/binary_backend.ex:612: Nx.BinaryBackend."-element_wise_bin_op/4-lbc$^17/2-9-"/7
       (nx 0.1.0) lib/nx/binary_backend.ex:611: Nx.BinaryBackend.element_wise_bin_op/4
       (nx 0.1.0) lib/nx/defn/evaluator.ex:149: Nx.Defn.Evaluator.eval_apply/4
       (nx 0.1.0) lib/nx/defn/evaluator.ex:63: Nx.Defn.Evaluator.eval/3
       (elixir 1.13.3) lib/enum.ex:1715: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (nx 0.1.0) lib/nx/defn/evaluator.ex:130: Nx.Defn.Evaluator.eval_apply/4
       (nx 0.1.0) lib/nx/defn/evaluator.ex:63: Nx.Defn.Evaluator.eval/3
       (elixir 1.13.3) lib/enum.ex:1715: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (nx 0.1.0) lib/nx/defn/evaluator.ex:130: Nx.Defn.Evaluator.eval_apply/4
       (nx 0.1.0) lib/nx/defn/evaluator.ex:63: Nx.Defn.Evaluator.eval/3
       (nx 0.1.0) lib/nx/defn/evaluator.ex:30: Nx.Defn.Evaluator.__jit__/4
       (for doctest at) lib/scholar/spatial/distances.ex:115: (test)
```

lib/scholar/shared.ex
```diff
@@ -0,0 +1,28 @@
...
+  # Collection of private helper functions and
+  # macros for enforcing shape/type constraints,
+  # and doing shape calculations.
+
+  import Nx.Defn
+
+  @doc """
+  Asserts `lhs` has same shape as `rhs`.
+  """
+  defn assert_shape!(lhs, rhs) do
```
josevalim, line 13: We already have `assert_shape` in Kernel, so my suggestion would be:

```suggestion
  defn assert_same_shape!(lhs, rhs) do
```

lib/scholar/spatial/distances.ex
```diff
@@ -0,0 +1,269 @@
...
+  defn minkowski(x, y, p \\ 2) do
+    assert_shape!(x, y)
+
+    x
+    |> Nx.subtract(y)
+    |> Nx.abs()
+    |> Nx.power(p)
+    |> Nx.sum()
+    |> Nx.power(1.0 / p)
+    |> Nx.as_type({:f, 32})
```
josevalim, line 217: This will downcast f64 to f32. Is it necessary? I think `Nx.power(..., float)` will always return a float already!

josevalim:
> I don't think we should copy scipy all the time re: organisation of code but in this case i think they've got this spot on. An alternative is what sklearn does, which is to put distance functions in metrics. I think that makes less sense, and it's because the focus t[here](https://scikit-learn.org/stable/modules/metrics.html#metrics) is on pairwise (see here).

My concern is that, looking beyond the distance functionality, we will definitely have the other functionality in `sklearn.metrics` but we don't share much with the other spatial functionality. For example, where would we put `sklearn.metrics.pairwise.kernel_metrics`? Or the regression metrics?

Even though I agree spatial are a perfect fit for distances, I don't think we will be hosting much functionality there. So my $0.02 would be to go with `Scholar.Metrics.Distance`.

cigrainger:
I think you're right on that. I would see `Spatial` as containing things like knn (and ann) but you're right it's relatively limited. And nothing is stopping us from using the `Scholar.Metrics.Distance` functions there. Perfectly happy to put `Distance` under `Metrics` as I think it's very unlikely we have too many non-metric distances to support.

josevalim (approved):
And so it begins!!!

PR #2, opened 2022-02-21
add standard scaler

tiagodavi:
This is a WIP PR. 

I would like to validate a few things:

1 - Is it ok to create a new module to store functions like standard_scaler, mix_max_scaler, robust_scaler etc? 

2 - I noticed the tagged NX version 0.1.0 doesn't have useful functions such as variance/standard_deviation yet. Are e we planning to release a new tag version or should change git here to download directly from Nx/main ?

3 - I don't think we need arguments such as axes, ddof here, do we? 

https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html


lib/scholar/scaler.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Scholar.Scaler do
+  @moduledoc """
+  Set of functions to change data scale.
+  """
+
+  @doc """
+  Standardizes the tensor by removing the mean and scaling to unit variance.
+  Formula: `z = (x - u) / s`
```
josevalim, line 8:
```suggestion

  Formula: `z = (x - u) / s`

```

josevalim:
1. Yes, I would call it `Scholar.Preprocessing.StandardScaler`. :)

2. Let's use git for now.

3. Up to you, we can call them later.

I am mostly curious with how we are going to implement the `fit`/`train` contract. :)

josevalim:
> Is the goal of this library being something alike sklearn? I mean, with algorithms such as decision tree, svm and so on?

Yes!

> To respect a contract like fit/transform. I think we can create a Scaler structure, return it after fitting and apply transformation after transforming. This helps us create something like inverse_transform as well.

So what if we flip it around? And we do this:

1. We define a module name `Sklearn.Preprocessing.Scaler`
2. We define many functions in it such as `standard`. They return a %Scaler{}.
3. We also have a `transform` function that receives the scaler and the data

Are there other things that follow the `fit/transform` pattern besides scalers?

lib/scholar/preprocessing/scaler.ex
```diff
@@ -0,0 +1,63 @@
...
+  def fit([]), do: []
+
+  def fit(data),
+    do: raise(ArgumentError, "expected a list with a least two elements got #{inspect(data)}")
+
+  @spec transform(scaler :: t()) :: t()
+  def transform(%__MODULE__{data: data} = scaler) do
+    tensor = Nx.to_tensor(data)
+
+    transformation =
```
tiagodavi, line 50: I would like some ideas on this formula if possible because it should return the same sklearn returns, but it's different despite the math formula is ok? 

```python
from sklearn.preprocessing import StandardScaler
data = [[1,2,3]]
scaler = StandardScaler()
scaler.fit(data)
print(scaler.transform(data)) # [[0. 0. 0.]]
```

tiagodavi:
@josevalim I pushed a couple of new changes, but I could use some help on the formula and validate how to improve the fit function a little bit more. 

I'll investigate more about the formula. sklearn code is kind of hard to understand.

seanmor5:
Just jumping in to discuss API - what is the value of stateful vs stateless transformations in this case? E.g. what is the value of introducing a new data structure which tracks state, versus just implementing functional transformations: https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.scale.html#sklearn.preprocessing.scale ? 

In the event it's valuable to go forth with the data structure, I would maybe go even more general than just `Scaler` and introduce a `Scholar.State` which would be a shared state abstraction across all estimators / preprocessors. We might have to build out a few more estimators before converging on what that contract looks like or if it makes sense, but if we can settle on a shared representation I think it could simplify things moving forward

josevalim:
@seanmor5 I may have misunderstood. My understand was that `fit` receives a training set and `transform` receives an actual set. However, if `fit` does not do any preparation of the data (which seems to be the case), then I agree with you a simple `standard_scale` function would work better.

Thoughts @tiagodavi?

tiagodavi:
@josevalim @seanmor5 

Yeah. I agree those could be simple functions. I think sklearn uses fit/transform to keep consistency with its API, but it's not required for that case, on the other hand there are cases where fit/predict would make sense, for example, to apply an algorithm to a training set then apply predict against the testing set with the previously learned params. We could call them train/predict instead.

Anyway a help is welcome on the standard_scaler formula/function as well as suggestions for the API. 

Thanks.

josevalim:
@seanmor5 @cigrainger for the API, let's go with `Scholar.Preprocessing.standard_scaler` for now? Once that's in I can take a look at the implementation too. :)

cigrainger:
I agree on `Scholar.Preprocessing.standard_scaler` for now. And I agree with @seanmor5 that it's best to stick to functional transformations. I could see something like the R package [`recipes`](https://recipes.tidymodels.org/articles/recipes.html) for building functional pipelines. We should have a think about that because i think Elixir is particularly well suited to an ergonomic experience here. But for now, the above doesn't preclude that at all.

tiagodavi:
@josevalim done! I changed the code again, but I am still thinking about the formula because:

```python
from sklearn.preprocessing import StandardScaler
import numpy as np
data = [[ 1, -1,  2], [ 2,  0,  0], [ 0,  1, -1]]
        
scaler = StandardScaler()
print(np.mean(data))
print(np.std(data))
print(scaler.fit_transform(data))

0.4444444444444444
1.0657403385139377
[[ 0.         -1.22474487  1.33630621]
 [ 1.22474487  0.         -0.26726124]
 [-1.22474487  1.22474487 -1.06904497]]
```

```python
data = [[ 1, -1,  2], [ 2,  0,  0], [ 0,  1, -1]]
mean = np.mean(data)
std = np.std(data)
data = (data-mean)/std
print(mean)
print(std)
data

0.4444444444444444
1.0657403385139377
array([[ 0.52128604, -1.35534369,  1.4596009 ],
       [ 1.4596009 , -0.41702883, -0.41702883],
       [-0.41702883,  0.52128604, -1.35534369]])
```

```elixir
iex> Scholar.Preprocessing.standard_scaler(Nx.tensor([[1, -1, 2], [2, 0, 0], [0, 1, -1]]))
        #Nx.Tensor<
          f32[3][3]
          [
            [0.5212860703468323, -1.3553436994552612, 1.4596009254455566],
            [1.4596009254455566, -0.4170288145542145, -0.4170288145542145],
            [-0.4170288145542145, 0.5212860703468323, -1.3553436994552612]
          ]
        >
```

I think there's something special happening inside sklearn class despite we implement the same formula. 
https://kenzotakahashi.github.io/scikit-learns-useful-tools-from-scratch.html

josevalim:
@tiagodavi it seems that the code in sklearn is quite more complex. Unfortunately I am not familiar enough to know from where the differences are coming from: https://github.com/scikit-learn/scikit-learn/blob/7e1e6d09b/sklearn/preprocessing/_data.py#L632 :/

tiagodavi:
It appears sklearn apply a "scale factor" somehow change the final result, but maybe it's not so important since we are applying the correct mathematical formula here.

lib/scholar/preprocessing.ex
```diff
@@ -0,0 +1,54 @@
...
+        >
+
+        iex> Scholar.Preprocessing.standard_scaler(42)
+        #Nx.Tensor<
+          s64
+          42
+        >
+  """
+  @spec standard_scaler(tensor :: Nx.Tensor.t()) :: Nx.Tensor.t()
+  def standard_scaler(%Nx.Tensor{} = tensor) do
```
josevalim, line 40: Can you please convert this to `defn`?

tiagodavi: Yes, sure.

tiagodavi: done @josevalim

test/scholar/preprocessing_test.exs
```diff
@@ -0,0 +1,26 @@
+defmodule Scholar.PreprocessingTest do
+  use ExUnit.Case, async: true
+
+  alias Scholar.Preprocessing
+
+  describe "standard_scaler/1" do
+    test "should apply standard scaler when data is a tensor" do
```
josevalim, line 7:
```suggestion
    test "applies standard scaling to data" do
```

test/scholar/preprocessing_test.exs
```diff
@@ -0,0 +1,26 @@
...
+        Nx.tensor([
+          [0.5212860703468323, -1.3553436994552612, 1.4596009254455566],
+          [1.4596009254455566, -0.4170288145542145, -0.4170288145542145],
+          [-0.4170288145542145, 0.5212860703468323, -1.3553436994552612]
+        ])
+
+      assert expected == Preprocessing.standard_scaler(data)
+    end
+
+    test "should leave data as it is when variance is zero" do
```
josevalim, line 20:
```suggestion
    test "leaves data as it is when variance is zero" do
```

josevalim:

PR #4, opened 2022-03-15
Start linreg

seanmor5:
WIP. Wanted to get the ball rolling on some actual estimators. @polvalente please correct my poor Linear Algebra :)

Open for discussion about the API. w.r.t fit / train, I don't know if train is the right word for this and possibly some other implementations. But I'm open to change :)


lib/scholar/linear_regression.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Scholar.LinearRegression do
+
+  defstruct :coefficients
```
polvalente, line 3: Do I smell a `Scholar.Summary` protocol coming up? haha

lib/scholar/linear_regression.ex
```diff
@@ -0,0 +1,42 @@
...
+      u
+      |> Nx.transpose()
+      |> Nx.dot(y)
+
+    s_inv = Nx.new_axis(1 / s, -1)
+
+    x =
+      vt
+      |> Nx.transpose()
+      |> Nx.dot(Nx.multiply(s_inv, uty))
```
polvalente, line 25: The overall linalg seems right! Although I had never seen this formulation, going through `At . A = At.b` with `A = U.S.Vt` yielded the same result.

I got to this other implementation which has some parts which are a little bit more direct (without needing to use Nx.new_axis). Perhaps there's something there you like?

```elixir
iex(9)> a                                                                                           
#Nx.Tensor<
  s64[3][3] 
  Torchx.Backend(cpu)
  [
    [1, 1, 1],
    [0, 2, 1],
    [0, 0, 3]
  ]
>
iex(10)> b
#Nx.Tensor<
  s64[3]
  Torchx.Backend(cpu)
  [3, 5, 9]
>
iex(11)> Nx.LinAlg.solve(a, b)
#Nx.Tensor<
  f32[3]
  Torchx.Backend(cpu)
  [-1.0, 1.0, 3.0]
>
iex(12)> {u, s, vt} = Nx.LinAlg.svd(a)                                                               
{#Nx.Tensor<
   f32[3][3]
   Torchx.Backend(cpu)
   [
     [0.4022642970085144, 0.3421006500720978, 0.8492057919502258],
     [0.4796755611896515, 0.711304783821106, -0.5137672424316406],
     [0.7798042893409729, -0.6140136122703552, -0.12203514575958252]
   ]
 >,
 #Nx.Tensor<
   f32[3]
   Torchx.Backend(cpu)
   [3.520359516143799, 1.9629522562026978, 0.868269681930542]
 >,
 #Nx.Tensor<
   f32[3][3]
   Torchx.Backend(cpu)
   [
     [0.11426796764135361, 0.3867832124233246, 0.915063738822937],
     [0.1742786318063736, 0.8990083336830139, -0.4017597734928131],
     [0.9780439138412476, -0.20538432896137238, -0.03531980887055397]
   ]
 >}
iex(13)> vt |> Nx.transpose() |> Nx.multiply(Nx.divide(1, s)) |> Nx.dot(Nx.transpose(u)) |> Nx.dot(b)
#Nx.Tensor<
  f32[3]
  Torchx.Backend(cpu)
  [-0.9999991655349731, 0.9999992847442627, 2.999999761581421]
>
```

PS: In retrospect, the final formulation is just the same as using x = inv(A).b where inv(A) is approximated through SVD :thinking:

lib/scholar/linear_regression.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Scholar.LinearRegression do
+
+  defstruct :coefficients
+
+  @doc """
+  Fits a linear regression model to the given dataset.
+  """
+  def fit(x, y) do
```
polvalente, line 8: I like `fit` better than `train` because it's more general and can encompass cases such as this one where there's no iterative learning

cigrainger: Agreed @polvalente!

lib/scholar/linear_regression.ex
cigrainger: Yesssssssssssssss

cigrainger:
As a general point, it's worth having a look at options in [`statsmodels`](https://www.statsmodels.org/stable/regression.html), [R's `lm`](https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/lm) _and_ [`sklearn`](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html) for options. For example, we'll definitely want to handle the case where we omit the intercept. Sample weights too. I think those are table stakes.

Are there any checks we need to make on the data to ensure OLS will compute without errors here?

lib/scholar/linear_regression.ex
polvalente: @cigrainger w.r.t checking the data, we don't really need to check the contents of the matrices. We could support both tall and wide a matrices (which neither code seems to support right now). Brace for incoming linear algebra :sweat_smile: 

Let's suppose we have A{n, m} and b{n, k}, which yields x{m, k}.

For the A matrix with shape {n, m}, Nx.LinAlg.svd returns:
U: {n, n}
S: {m}
Vt: {m, m}

We would need to transform the S vector into an {n, m} diagonal matrix before applying the `iex(13)` line above, and it would need to change its Nx.multiply into Nx.dot with the transposed S matrix. Something like:

`vt |> Nx.transpose() |> Nx.dot(Nx.divide(1, s_as_diagonal(s, n, m))) |> Nx.dot(Nx.transpose(u)) |> Nx.dot(b)`

with:

```elixir
def s_as_diagonal(t, n, m) do
  # Note that this returns the transposed shape to be used directly where we need it
  [t, Nx.broadcast(0, {abs(m - n)})]
  |> Nx.concatenate()
  |> Nx.make_diagonal()
  |> Nx.slice([0, 0], [m, n])
end
```

polvalente: PS: Some NaNs/Infs can appear due to `Nx.divide(1, s)` if s contains 0-valued singular values. This doesn't crash the code, but will return an unusable result. I don't think there's too much we can do though

cigrainger: Makes sense! I think it's a good idea to warn in that case?

polvalente: Do you mean we would print something or raise? 
Perhaps we could return a flag which indicates if the result is numerically valid together with the data of interest?

We would need to add something like `Nx.has_nan` and `Nx.has_inf` or even `Nx.numerically_valid` though, to make this check work.

polvalente: Actually, nevermind, we just need to `Nx.any(Nx.less_equal(s, eps))` and operate from there

cigrainger: Print something I think -- e.g. [`IO.warn/1`](https://hexdocs.pm/elixir/1.12/IO.html#warn/1).

lib/scholar/linear_regression.ex
josevalim, line 8:
```suggestion
  defn fit(x, y) do
```

?

lib/scholar/linear_regression.ex
```diff
@@ -0,0 +1,42 @@
...
+      |> Nx.transpose()
+      |> Nx.dot(Nx.multiply(s_inv, uty))
+
+    Nx.transpose(x)
+  end
+
+  defimpl Scholar do
+    import Nx.Defn
+
+    def predict(%LinearRegression{coefficients: coef}, x) do
```
josevalim, line 33: I think we can have `predict` actually defined in `Scholar.LinearRegression` and then we just delegate to it. We should probably define `Nx.Container` for the struct too?

lib/scholar/linear_regression.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Scholar.LinearRegression do
```
josevalim, line 1: Do we want to namespace this inside `Scholar.Stats` or similar?

lib/scholar/linear_regression.ex
seanmor5: I was using the least squares implementation that JAX has which is somewhat what I've implemented here. As far as ensuring s is safe, we'll just need a fuzz value:

```elixir
mask = ...some check that s is safe...
safe_s = Nx.select(mask, s, eps)
```

Where eps is some small value

lib/scholar/linear_regression.ex
seanmor5: For sklearn this is namespaced under `linear_model` with LogisticRegression and a couple of others. We can go that route. I think it depends how we want to structure various models / algorithms

josevalim: I think we have three options here:

1. Have a single module with a bunch of functions. All of those functions reuse the same struct that would have be like `Scholar.Linear{op: :linear_regression, data: ...}`

2. Have one module per operation, each with their own struct. Then I think we should namespace because the collection is too large

3. Have one module per operation with a shared struct

Looking at sklearn, the parameters of linear structs do not actually match. LinearRegression has different parameters than ElasticNet, so it makes me think 2 is the best option here.

However, I am wondering if we should go the protocol route. The current approach means that, for each model, we will have to define at least 3 modules: the model module, the impl for Scholar, and the impl for Nx.Container. Likely others in the future. So I am wondering if instead we should make `Scholar` a behaviour and merge the model module with the impl module. In any case, it can be done in a future PR.

TL;DR - yes, let's namespace it.

seanmor5:
@polvalente I cleaned up the implementation and added rescaling according to what sklearn does. Unfortunately not passing any tests yet :(

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,65 @@
...
+    {Nx.dot(x, scale), Nx.dot(y, scale)}
+  end
+
+  # Implements ordinary least-squares by estimating the
+  # solution A to the equation A.X = b.
+  defnp lstsq(x, y) do
+    {u, s, vt} = Nx.LinAlg.svd(x)
+
+    mask = Nx.not_equal(s, 0)
+    safe_s = Nx.select(mask, s, 1)
```
polvalente, line 56:
```suggestion
    safe_s = Nx.select(mask, s, 1.0e-10)
```

Shouldn't it be eps instead of 1 here?

lib/scholar/metrics.ex
```diff
@@ -0,0 +1,418 @@
+defmodule Scholar.Metrics do
```
polvalente, line 1: Why are these changes from #5 showing up as diff from this PR? :confused:

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,65 @@
+defmodule Scholar.Linear.LinearRegression do
+  @moduledoc """
+  Ordinary least squares linear regression.
+  """
+  import Nx.Defn
+  alias __MODULE__, as: LinearRegression
```
polvalente, line 6:
```suggestion
  alias __MODULE__
```

seanmor5:
This is stale, and I won't be able to get back to working on it anytime soon :)

msluszniak:
I'd like to continue this PR, but I'll be grateful if you write me about what needs to be added and changed.

msluszniak:
@seanmor5 @polvalente

seanmor5:
@msluszniak I don't think much really needs to change---from what I could tell from the sklearn LinReg implementation this one implemented more or less the same thing. But I let this get stale so I ended up closing, feel free to take this work to another branch :)

PR #5, opened 2022-03-18
feat(metrics): bring Axon.Metrics to Scholar.Metrics

polvalente:
Brings the Axon.Metrics module over, with some changes to the moduledoc (which had some motivations regarding training and NN specific terms)

### DIFF
--- a/test/scholar/metrics_test.exs
+++ b/test/scholar/metrics_test.exs
@@ -1,0 +1,5 @@
+defmodule Scholar.MetricsTest do
+  use ExUnit.Case
+
+  doctest Scholar.Metrics
+end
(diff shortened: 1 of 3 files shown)


test/scholar/metrics_test.exs
```diff
@@ -0,0 +1,5 @@
+defmodule Scholar.MetricsTest do
+  use ExUnit.Case
```
josevalim, line 2:
```suggestion
  use ExUnit.Case, async: true
```

josevalim (approved):
LGTM! We could also have a separate library to keep only metrics but I am fine either way.

PR #8, opened 2022-05-03
calculate confusion matrix

cocoa-xu:


lib/scholar/metrics.ex
```diff
@@ -331,6 +331,76 @@ defmodule Scholar.Metrics do
...
+    * `y_true` - $\(d_0, d_1, ..., d_n\)$
+    * `y_pred` - $\(d_0, d_1, ..., d_n\)$
+
+  ## Examples
+
+      iex> y_true = Nx.tensor([1.0, 1.0, 2.0, 2.0, 3.0, 3.0], type: {:f, 32})
+      iex> y_pred = Nx.tensor([1.0, 2.0, 1.0, 3.0, 3.0, 3.0], type: {:f, 32})
+      iex> Scholar.Metrics.confusion_matrix(y_true, y_pred)
+  """
+  def confusion_matrix(y_true, y_pred) do
```
jonatanklosko, line 348: We want to use `defn` whenever possible. I think it makes sense to require `y_true` and `y_pred` to be integer tensors, with class numbers starting from 0. Then we need a required option like `:num_classes`, because we need to know the confusion matrix size upfront.

Given that, one way to implement this would be creating a zeroed confusion matrix, then then concatenate `y_true` and `y_pred` to create indices into the matrix, and finally do `indexed_add` with those indices and ones as values.

Related: [tf.math.confusion_matrix](https://www.tensorflow.org/api_docs/python/tf/math/confusion_matrix), [sklearn.metrics.confusion_matrix](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.confusion_matrix.html).

jonatanklosko, line 348: We want to use `defn` whenever possible. I think it makes sense to require `y_true` and `y_pred` to be integer tensors, with class numbers starting from 0. Then we need a required option like `:num_classes`, because we need to know the confusion matrix size upfront.

Given that, one way to implement this would be creating a zeroed confusion matrix, then then concatenate `y_true` and `y_pred` to create indices into the matrix, and finally do `indexed_add` with those indices and ones as values.

Related: [tf.math.confusion_matrix](https://www.tensorflow.org/api_docs/python/tf/math/confusion_matrix), [sklearn.metrics.confusion_matrix](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.confusion_matrix.html).

test/scholar/metrics_test.exs
```diff
@@ -1,5 +1,22 @@
...
+  @y_true Nx.tensor([1.0, 1.0, 2.0, 2.0, 3.0, 3.0], type: {:f, 32})
+  @y_pred Nx.tensor([1.0, 2.0, 1.0, 3.0, 3.0, 3.0], type: {:f, 32})
+  @cm Nx.tensor(
+        [
+          [1, 1, 0],
+          [1, 0, 1],
+          [0, 0, 2]
+        ],
+        type: {:s, 64}
+      )
```
jonatanklosko, line 16: You can cover this with a doctest by adding the result below

```
iex> Scholar.Metrics.confusion_matrix(y_true, y_pred)
```

in function docs. See the other metrics :)

jonatanklosko, line 16: You can cover this with a doctest by adding the result below

```
iex> Scholar.Metrics.confusion_matrix(y_true, y_pred)
```

in function docs. See the other metrics :)

cocoa-xu: Oh, I forgot to mention that I tried to write the test here, but I encountered some weird errors. It works well if I only return a single Nx tensor, but it failed to parse the result when a tuple is returned. Not sure about the reason...

```
1) doctest Scholar.Metrics.confusion_matrix/3 (2) (Scholar.MetricsTest)
     test/scholar/metrics_test.exs:4
     Doctest did not compile, got: (SyntaxError) lib/scholar/metrics.ex:346:3: syntax error before: '['
          |
      346 |   [
          |   ^
     doctest:
       iex> y_true = Nx.tensor([1.0, 1.0, 2.0, 2.0, 3.0, 3.0], type: {:f, 32})
       iex> y_pred = Nx.tensor([1.0, 2.0, 1.0, 3.0, 3.0, 3.0], type: {:f, 32})
       iex> Scholar.Metrics.confusion_matrix(y_true, y_pred)
       {#Nx.Tensor<
         s64[3][3]
         [
           [1, 1, 0],
           [1, 0, 1],
           [0, 0, 2]
         ]
       >, 
       #Nx.Tensor<
         f32[3]
         [1.0, 2.0, 3.0]
       >}
```

But since we're going to ask the user to provide `:num_classes` in opts, this will not be a problem. :)

lib/scholar/metrics.ex
```diff
@@ -331,6 +331,76 @@ defmodule Scholar.Metrics do
...
+    {unique_label_atoms, unique_labels, num_classes} =
+      [y_true, y_pred]
+      |> Nx.concatenate()
+      |> Nx.to_flat_list()
+      |> Enum.reduce(%{}, fn label, acc ->
+        Map.update(acc, String.to_atom("#{label}"), label, fn _ -> label end)
```
polvalente, line 354: This probably will be removed upon refactoring to defn, but `String.to_atom` is a no-go because of atom table exhaustion

polvalente: I would also avoid any to_flat_list calls because they take data away from the device

lib/scholar/metrics.ex
cocoa-xu: Cool! But I'm not sure about how to create a zero tensor with a specific shape. Is there any function to achieve that, like `Nx.zeros/1` or `Nx.ones/1`?

polvalente: You want Nx.broadcast(0, shape)

lib/scholar/metrics.ex
```diff
@@ -331,6 +331,49 @@ defmodule Scholar.Metrics do
...
+          [0, 0, 2]
+        ]
+      >
+  """
+  def confusion_matrix(y_true, y_pred, opts \\ []) do
+    num_classes = opts[:num_classes]
+
+    cm =
+      Nx.iota({num_classes, num_classes})
+      |> then(&Nx.subtract(&1, &1))
```
polvalente:
```suggestion
    cm = Nx.broadcast(0, {num_classes, num_classes})
```

polvalente: We should probably also pass the backend from either y_true or y_pred. Not exactly sure if defn will pick up on that

lib/scholar/metrics.ex
```diff
@@ -331,6 +331,44 @@ defmodule Scholar.Metrics do
...
+
+    indices =
+      Enum.to_list(0..(Nx.size(y_true) - 1))
+      |> Enum.map(&Nx.tensor(&1))
+      |> Enum.map(&[Nx.take(y_true, &1), Nx.take(y_pred, &1)])
+      |> Nx.tensor()
+
+    Nx.indexed_add(
+      Nx.broadcast(0, {num_classes, num_classes}),
+      indices,
```
polvalente, line 367: Your should use Nx.new_axis + Nx.concatenate instead

lib/scholar/metrics.ex
```diff
@@ -331,6 +331,44 @@ defmodule Scholar.Metrics do
...
+      Nx.broadcast(1, {Nx.size(y_true)})
+    )
+  end
+
   @doc ~S"""
```
polvalente, line 372:
```suggestion
      Nx.broadcast(1, {Nx.size(y_true)})
```

lib/scholar/metrics.ex
```diff
@@ -331,6 +331,44 @@ defmodule Scholar.Metrics do
...
+      >
+  """
+  defn confusion_matrix(y_true, y_pred, opts \\ []) do
+    num_classes = opts[:num_classes]
+
+    indices =
+      Nx.iota({Nx.size(y_true)})
+      |> then(&[Nx.take(y_true, &1), Nx.take(y_pred, &1)])
+      |> Nx.stack()
+      |> Nx.transpose()
```
polvalente, line 363: Please see my comment on Nx.new_axis + Nx.concatenate

Nx.transpose is costly and if you concatenate along the correct axis beforehand it will be unnecessary.

Torchx.backend uses this quite frequently, IIRC

cocoa-xu: Sorry I didn't see that before the previous commit. 😅 I think it should be okay now (except the backend thing), I wonder if there is anything else to improve?

polvalente: I think it's just the backend thing. I'm not at the PC right now, but if you cant find how to get it and pass it along, I can deal with this later :)

cocoa-xu: I was looking at the docs wishing to find something like `Nx.tensor_backend/1` that gives me the backend of a tensor, but I didn't find one that looks like the correct function...I guess I'll leave that here, sorry.

polvalente: let's leave this as a TO-DO then. Worst case scenario the defn will use the default backend with default options to allocate the new tensors and that's probably the same env in which other tensors live anyway

PR #10, opened 2022-05-04
calculates f1 scores

cocoa-xu:


lib/scholar/metrics.ex
```diff
@@ -369,6 +369,93 @@ defmodule Scholar.Metrics do
...
+  ## Options
+
+    * `:num_classes` - required. Number of classes contained in the input tensors
+    * `:average` - optional. This determines the type of averaging performed on the data.
+      * `nil`. Default value. In this case, the scores for each class are returned.
+      * `:macro`. Calculate metrics for each label, and find their unweighted mean.
+        This does not take label imbalance into account.
+      * `:weighted`. Calculate metrics for each label, and find their average weighted by
+        support (the number of true instances for each label).
+      * `:micro`. Calculate metrics globally by counting the total true positives,
```
polvalente, line 385: Any references for the naming here?

lib/scholar/metrics.ex
```diff
@@ -369,6 +369,93 @@ defmodule Scholar.Metrics do
...
+      transform(opts[:average], fn average ->
+        average || nil
+      end)
+
+    cm = confusion_matrix(y_true, y_pred, num_classes: num_classes)
+    true_positive = Nx.take_diagonal(cm)
+    false_positive = Nx.subtract(Nx.sum(cm, axes: [0]), true_positive)
+    false_negative = Nx.subtract(Nx.sum(cm, axes: [1]), true_positive)
+    precision = Nx.divide(true_positive, Nx.add(true_positive, false_positive))
+    recall = Nx.divide(true_positive, Nx.add(true_positive, false_negative))
```
polvalente, line 434: Aren't there functions to ger precision, recall and so on already?

lib/scholar/metrics.ex
cocoa-xu: I was referencing to sklearn when I wrote this, [sklearn.metrics.f1_score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.f1_score.html). The example data comes from [here](https://towardsdatascience.com/micro-macro-weighted-averages-of-f1-score-clearly-explained-b603420b292f).

polvalente: Do jax or pytorch have a similar function?

lib/scholar/metrics.ex
cocoa-xu: Existing functions only works for binary case (the label is either 0 or 1). The code here works for any number of classes.

polvalente: Shouldn't they work for other cases as well?
Perhaps it should ne changed as part of a separate PR on which this one depends

cocoa-xu: I think that could be changed in a separate PR, however, the approach here would require to calculate the confusion matrix first (which should be pretty fast for binary labels). If that's okay, then I can send a separate PR. :)

jonatanklosko: > the approach here would require to calculate the confusion matrix first (which should be pretty fast for binary labels)

We can use a different implementation when `num_classes: 2`.

The primary difference is that the current implementations accept a `:threshold`, and so they are semantically different. But maybe it's better to apply such normalization (probabilities -> classes) before using the metrics. I believe the implementations come from Axon, @seanmor5 thoughts?

cocoa-xu: Another thing to note, `0` stands for `negative` in binary-class cases, however, in multi-class cases `0` should be treated as any other label. 

In simple words, even if `num_classes` is 2, in binary cases `1` and `0` represent `true` and `false` respectively, whereas the multi-class cases `1` and `0` could represent `cat` and `dog`. 

Therefore, when considering true positives, we only count the number of samples whose y_pred == 1 and y_true == 1. However, for multi-class, true positives should be the sum of samples whose (y_pred == 0 and y_true == 0) **and** (y_pred == 1 and y_true == 1).

cocoa-xu: Say we have

```
y_true = Nx.tensor([1, 0, 1, 1, 0, 1, 0])
y_pred = Nx.tensor([1, 1, 0, 0, 1, 0, 0])
```

then the number of true positives should be 1 in binary-class case, while it should be 2 in multi-class case.

Binary-class:
```
y_true = Nx.tensor([1, 0, 1, 1, 0, 1, 0])
                    ^
y_pred = Nx.tensor([1, 1, 0, 0, 1, 0, 0])
                    ^
```

Multi-class:
```
y_true = Nx.tensor([1, 0, 1, 1, 0, 1, 0])
                    ^                 ^
y_pred = Nx.tensor([1, 1, 0, 0, 1, 0, 0])
                    ^                 ^
```

cocoa-xu: Maybe we should set `num_classes` to 1 to denote the former case (1/0 denotes true/false). This should make sense because the question for the former case could be 'is this a cat' (only 1 class), while the latter case is asking 'is this cat or dog' (2 classes).

jonatanklosko: Looking at [sklearn.metrics.recall_score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.recall_score.html), they support `average="binary"` and return the value only for the true class (configurable with `pos_label`). For binary input they make it the default behaviour:

```
>>> y_true = [1, 0, 1, 1, 0, 1, 0]
>>> y_pred = [1, 1, 0, 0, 1, 0, 0]
>>> recall_score(y_true, y_pred)
0.25
>>> recall_score(y_true, y_pred, average=None)
array([0.33333333, 0.25      ])
```

cocoa-xu: yeah, that's true. I think this also depends on the perspective, if we use the One-vs-Rest strategies for multi-class cases, then binary interpretation seems to be more general.

lib/scholar/metrics.ex
cocoa-xu: I didn't find it in jax or pytorch, but F1 score is a useful measure for imbalanced classification problems.

lib/scholar/metrics.ex
jonatanklosko: It seems like the binary-class version is semantically a separate function. Again, looking at sklearn, the `:average` option doesn't make sense for binary (maybe that's why they use `average="binary"`, which reads pretty weird to be honest). On the other hand, the `pos_label` option doesn't make sense for multi-class.

There is a huge number of metrics, so I can see striving the decision to prefer a single parameterizable function, but at the same time, having `binary_recall_score` and `recall_score` may just be less confusing, especially that we want to special case the implementation for them anyway.

I think `num_classes: 1` can be confusing, an alternative approach would be `num_classes: :binary` (maybe no longer `num_`). Something along this lines sounds better than `average: :binary`, but I'm not convinced if that's the best we can do.

Could definitely use other opinions on this :D

(Sidenote: having confusion matrix, maybe we no longer need to expose the individual `tp` `fp` `tn` `tn` functions. sklearn doesn't)

cocoa-xu: I like the idea of having two sets of functions too (one for binary-class and the other for multi-class). It seems to be less confusing and easier to maintain. Besides that, there's probably no way to combine them in a single function without some sort of if-statement inside.

lib/scholar/metrics.ex
cigrainger: It's in [`torchmetrics`](https://torchmetrics.readthedocs.io/en/stable/classification/f1_score.html).

lib/scholar/metrics.ex
```diff
@@ -336,6 +336,93 @@ defmodule Scholar.Metrics do
...
+          Nx.mean(per_class_f1)
+
+        :weighted ->
+          support = Nx.sum(Nx.equal(y_true, Nx.iota({num_classes, 1})), axes: [1])
+          Nx.sum(Nx.multiply(per_class_f1, Nx.divide(support, Nx.sum(support) + 1.0e-16)))
+
+        :micro ->
+          y_pred
+          |> Nx.equal(y_true)
+          |> Nx.mean()
```
jonatanklosko, line 421: This is just accuracy, is this correct?

cocoa-xu: yes, but I got errors when I tried to use `if`, no matter inside a transform or not... Not sure how to do this correctly. 

btw, using `case` works, is that what I should use?

### `case` works
```
  defn f1_score(y_true, y_pred, opts \\ []) do
    opts = keyword!(opts, [:num_classes, :average])

    assert_shape_pattern(y_true, {_})
    assert_shape(y_pred, Nx.shape(y_true))

    transform(opts[:average], fn average ->
      case average do
        :micro ->
          accuracy(y_true, y_pred)

        _ ->
          # ...
          case average do
            nil ->
              # ...
            :macro ->
              # ...
            :weighted ->
              # ...
          end
      end
    end)
  end
```

### `if` does not work
```
  defn f1_score(y_true, y_pred, opts \\ []) do
    opts = keyword!(opts, [:num_classes, :average])

    assert_shape_pattern(y_true, {_})
    assert_shape(y_pred, Nx.shape(y_true))

    if opts[:average] == :micro do
        # ...
    else
        # ...
    end
  end

(ArgumentError) expected a %Nx.Tensor{} or a number, got: nil
```

```
  defn f1_score(y_true, y_pred, opts \\ []) do
    opts = keyword!(opts, [:num_classes, :average])

    assert_shape_pattern(y_true, {_})
    assert_shape(y_pred, Nx.shape(y_true))

    transform(opts[:average], fn average ->
      if average == :micro do
          # ...
      else
          # ...
      end
    end)
  end

(ArgumentError) expected a %Nx.Tensor{} or a number, got: nil
```

```
  defn f1_score(y_true, y_pred, opts \\ []) do
    opts = keyword!(opts, [:num_classes, :average])

    assert_shape_pattern(y_true, {_})
    assert_shape(y_pred, Nx.shape(y_true))

    average = transform(opts[:average], &(&1))
    if average == :micro do
        # ...
    else
        # ...
    end
  end

(ArgumentError) expected a %Nx.Tensor{} or a number, got: nil
```

jonatanklosko: @cocoa-xu my question was rather whether f1 score `:micro` is just accuracy (?).

As for the error, inside transform the `==` is still the tensor-aware equality, you can do `Elixir.Kernel.==(average, :micro)` in those cases.

cocoa-xu: yes, micro f1 is just the accuracy. (however, if we'd like to also implement [F-beta scores](https://en.wikipedia.org/wiki/F-score), then it's different.)

jonatanklosko: Ah, I see, it's the same for single-label case! `fbeta` should be separate.

lib/scholar/metrics.ex
```diff
@@ -336,6 +336,92 @@ defmodule Scholar.Metrics do
...
+    assert_shape(y_pred, Nx.shape(y_true))
+
+    transform(opts[:average], fn average ->
+      if Elixir.Kernel.==(average, :micro) do
+        accuracy(y_true, y_pred)
+      else
+        num_classes =
+          transform(opts[:num_classes], fn num_classes ->
+            num_classes || raise ArgumentError, "missing option :num_classes"
+          end)
```
jonatanklosko, line 393: I think it makes sense to require it for consistency, so let's move it to the top.

cocoa-xu: sure!

jonatanklosko (approved):
Thanks :)

PR #12, opened 2022-05-04
Multi-class support

cocoa-xu:
WIP.  ~Only implemented for `true_positives/3` for now.~

~When `num_classes` is 1, then this is a binary classification problem; when `num_classes` is greater than 1, then this is a multi-class problem.~

~We can also follow sklearn which uses `average='binary'` to denote a binary classification problem.~

Functions with `binary_*` prefix will be used for binary classification problems, and the rest are for multi-class classification problems.


lib/scholar/metrics.ex
```diff
@@ -29,34 +54,65 @@ defmodule Scholar.Metrics do
...
+      iex> y_pred = Nx.tensor([[1, 0, 0, 0], [0, 0, 1, 0], [0, 1, 0, 0], [0, 0, 0, 1]])
+      iex> Scholar.Metrics.onehot_accuracy(y_true, y_pred)
+      #Nx.Tensor<
+        f32
+        0.5
+      >
 
   """
-  defn accuracy(y_true, y_pred) do
+  defn onehot_accuracy(y_true, y_pred) do
```
jonatanklosko, line 71: I don't think we need one-hot versions of the functions, we can always convert one-hot to class numbers beforehand.

lib/scholar/metrics.ex
```diff
@@ -29,34 +54,65 @@ defmodule Scholar.Metrics do
...
+    opts = keyword!(opts, [:num_classes])
+    assert_shape_pattern(y_true, {_})
+    assert_shape(y_pred, Nx.shape(y_true))
+
+    cm = confusion_matrix(y_true, y_pred, opts)
+
+    true_positive = Nx.take_diagonal(cm)
+    false_positive = Nx.subtract(Nx.sum(cm, axes: [0]), true_positive)
+
+    Nx.divide(true_positive, true_positive + false_positive + 1.0e-16)
```
jonatanklosko, line 110: This is precision :)

lib/scholar/metrics.ex
```diff
@@ -153,9 +282,36 @@ defmodule Scholar.Metrics do
...
+
+      iex> y_true = Nx.tensor([0, 1, 1, 1, 1, 0, 2, 1, 0, 1], type: {:u, 32})
+      iex> y_pred = Nx.tensor([0, 2, 1, 1, 2, 2, 2, 0, 0, 1], type: {:u, 32})
+      iex> Scholar.Metrics.true_positives(y_true, y_pred, num_classes: 3)
+      #Nx.Tensor<
+        s64[3]
+        [2, 3, 1]
+      >
+  """
+  defn true_positives(y_true, y_pred, opts \\ []) do
```
jonatanklosko, line 303: Honestly, I think exposing tp, tn, fp, fn is too granular given that we have `confusion_matrix`.

lib/scholar/metrics.ex
cocoa-xu: Ah, actually, for multi-class classification problems, precision = recall = accuracy = micro f1 score.<sup>*</sup>

<sup>*</sup>: when each observation/input only associates with a single label. Otherwise, it should be multi-label multi-class classification problems (which we need to handle in a different way).

lib/scholar/metrics.ex
cocoa-xu: Yup, they also seem to be somewhat redundant/duplicated in terms of some calculations, and the calculated confusion matrix cannot easily be reused between them. It's simpler to derive tp, tn, fp, fn from the confusion matrix.

lib/scholar/metrics.ex
jonatanklosko: We should return a single value, no? That's what [sklearn.metrics.accuracy_score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.accuracy_score.html) does.

cocoa-xu: sorry brain not working for a moment 😂 should be fixed now

lib/scholar/metrics.ex
```diff
@@ -103,42 +162,79 @@ defmodule Scholar.Metrics do
...
       iex> y_pred = Nx.tensor([0.8, 0.6, 0.4, 0.2, 0.8, 0.2, 0.2])
-      iex> Scholar.Metrics.true_positives(y_true, y_pred)
+      iex> Scholar.Metrics.binary_true_positives(y_true, y_pred)
       #Nx.Tensor<
         u64
         1
       >
   """
-  defn true_positives(y_true, y_pred, opts \\ []) do
+  defn binary_true_positives(y_true, y_pred, opts \\ []) do
```
jonatanklosko, line 237: We don't need these either right? We can keep them as private if they are useful.

We may want to benchmark functions like binary precision/recall using this implementation vs using confusion matrix, but we can do that separately.

Also, I wonder if accepting `:threshold` everywhere makes sense, given that we always just compute `thresholded_preds`. This could very well be a separate function, like one-hot conversion etc.

lib/scholar/metrics.ex
```diff
@@ -29,34 +26,60 @@ defmodule Scholar.Metrics do
...
+
+    cm = confusion_matrix(y_true, y_pred, opts)
+
+    true_positive = Nx.take_diagonal(cm)
+    false_positive = Nx.sum(Nx.subtract(Nx.sum(cm, axes: [0]), true_positive))
+    false_negative = Nx.sum(Nx.subtract(Nx.sum(cm, axes: [1]), true_positive))
+    true_positive = Nx.sum(true_positive)
+
+    Nx.divide(true_positive, true_positive + 0.5 * (false_positive + false_negative) + 1.0e-16)
+  end
```
jonatanklosko, line 78: Can't we do pretty much the same as the binary implementation?

```elixir
y_pred
|> Nx.equal(y_true)
|> Nx.mean()
```

jonatanklosko: Also, the only difference is the thresholding we do in binary accuracy, but as I mentioned in the other comment, I'm not convinced we should do that. If we work with class numbers only, then both implementations are the same.

cocoa-xu: you're right, the results are the same. Shall we expect the user pass normalised `y_pred` for all functions here? I think it's pretty easy to do `thresholded_preds` or convert one-hot encoded `y_pred`. They can be done using one single existing Nx APIs, i.e., `Nx.greater` or `Nx.argmax`. We probably don't need to have separate functions for them?

jonatanklosko: Exactly, I think expecting normalized classes makes sense! We can totally have functions in some other module for converting back and forth to make the user code more readable, I just don't think this should be a concern of the metric calculation.

cc @seanmor5 for an opinion :)

lib/scholar/metrics.ex
cocoa-xu: I think for binary cases, current implementation should be faster because `confusion_matrix` will first construct an `{n, 2}`-shape indices tensor and then update the confusion matrix `n` times.

jonatanklosko: > then update the confusion matrix n times

Pretty sure `indexed_add` is more optimized than that! Each individual tensor operation is closer to "all at once", and XLA optimizes across operations via fusion and whatnot, so it's best to measure various options.

cocoa-xu: Oh I didn't take parallelised operation into consideration! Then `indexed_add` should be fairly fast. Let's see the benchmark data and decide. :)

josevalim:
+1 for separating binary from non-binary ones. I think though, for accuracy, a single function for both will suffice.

My only question is if the `binary` functions should exist in a separate module but it feels many of metrics function can actually work on both (such as accuracy).

lib/scholar/metrics.ex
jonatanklosko: I created #13 for separating thresholding, we can handle it separately. We can still merge accuracy into a single function, and skip the `0.5` thresholding for binary :)

cocoa-xu:
> +1 for separating binary from non-binary ones. I think though, for accuracy, a single function for both will suffice.
> 
> My only question is if the `binary` functions should exist in a separate module but it feels many of metrics function can actually work on both (such as accuracy).

yeah, I also think many of them should work for both cases, and the confusion matrix can be used to derive most metrics we need.

lib/scholar/metrics.ex
cocoa-xu: I pushed a new commit that all functions expect normalized inputs. It was done last night but was waiting to see some other options. I agree a preprocess module could be used to do all the normalizing things instead of doing them in the metrics module.

lib/scholar/metrics.ex
```diff
@@ -98,197 +87,221 @@ defmodule Scholar.Metrics do
...
-      iex> Scholar.Metrics.true_positives(y_true, y_pred)
+      iex> y_pred = Nx.greater(y_pred, threshold)
+      iex> Scholar.Metrics.binary_true_positives(y_true, y_pred)
       #Nx.Tensor<
         u64
         1
       >
   """
-  defn true_positives(y_true, y_pred, opts \\ []) do
+  defn binary_true_positives(y_true, y_pred) do
```
jonatanklosko, line 194: Let's make these `defnp` :)

cocoa-xu: oops, didn't notice that they break the `running_sum/1` doctest. I changed the doctest `running_sum/1` from using  `binary_false_positives/2` to `mean_absolute_error/2`. (although it seems not quite make sense to have a sum of mean absolute error...)

lib/scholar/metrics.ex
```diff
@@ -266,71 +246,101 @@ defmodule Scholar.Metrics do
...
   end
 
+  @doc ~S"""
+  Computes the specificity of the given predictions with respect
+  to the given targets for multi-class classification problems.
+
+  ## Argument Shapes
+
+    * `y_true` - $\(d_0, d_1, ..., d_n\)$
+    * `y_pred` - $\(d_0, d_1, ..., d_n\)$
```
jonatanklosko, line 313: These are always 1-d tensors, no? We need to adjust in other places too.

cocoa-xu: Yeah, they should be 1-d tensors as we are only dealing with single-label problems. For multilabel problems, we probably should have a different set of functions?

jonatanklosko: Yeah, I'm not sure about multi labels, we can investigate that separately.

lib/scholar/metrics.ex
```diff
@@ -266,71 +246,101 @@ defmodule Scholar.Metrics do
...
+      |> Nx.logical_and(Nx.equal(y_pred, 0))
       |> Nx.sum()
 
     false_positives =
-      thresholded_preds
+      y_pred
       |> Nx.not_equal(y_true)
-      |> Nx.logical_and(Nx.equal(thresholded_preds, 1))
+      |> Nx.logical_and(Nx.equal(y_pred, 1))
       |> Nx.sum()
```
jonatanklosko, line 301: We can use the private functions :D

josevalim (approved):
Up to you to merge it @jonatanklosko! Thanks @cocoa-xu!

lib/scholar/metrics.ex
```diff
@@ -29,67 +26,61 @@ defmodule Scholar.Metrics do
...
   @doc ~S"""
-  Computes the precision of the given predictions with
-  respect to the given targets.
+  Computes the precision of the given predictions with respect to
+  the given targets for binary classification problems.
 
   ## Argument Shapes
 
     * `y_true` - $\(d_0, d_1, ..., d_n\)$
     * `y_pred` - $\(d_0, d_1, ..., d_n\)$
```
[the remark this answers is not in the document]
cocoa-xu: I think this should be fine? as `d_i` should be a scalar value, maybe we need to explicitly write that down, something like

```
`y_true` - $\(d_0, d_1, ..., d_n\)$ where $d_i$ is a scalar.
```

what do you think? shall we just remove them or make it more clear? 🤔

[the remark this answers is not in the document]
jonatanklosko: Isn't `(d_0, d_1, ..., d_n)` here the same as `{d_0, d_1, ..., d_n}` in defn world?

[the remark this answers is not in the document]
cocoa-xu: I didn't quite understand the defn world, but here it goes!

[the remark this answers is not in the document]
jonatanklosko: What I meant is that `y_true - (d_0, d_1, ..., d_n)` means `y_true` is a n-dimensional tensor, while we require it to be a 1-dimensional tensor.

jonatanklosko (approved):
Awesome, thanks!

lib/scholar/metrics.ex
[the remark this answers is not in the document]
cocoa-xu: Oh, now I see what you mean! 😂 d_i represents the element in the tensor's shape instead of the tensor itself.

PR #18, opened 2022-08-03
Add logistic regression

msluszniak:


lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,192 @@
+defmodule Scholar.Linear.LogisticRegression do
```
polvalente, line 1: I'm not sure about the `Linear` subdomain. What does it mean?
The logistic function isn't a linear transformation

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,192 @@
...
+  @moduledoc """
+  Logistic regression in both binary and multinomial variants
+  """
+  import Nx.Defn
+  alias __MODULE__
```
polvalente, line 6: I would rather use `__MODULE__` wherever `LogisticRegression` is needed here

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,192 @@
...
+
+  @derive {Nx.Container, containers: [:coefficients, :bias], keep: [:mode]}
+  defstruct [:coefficients, :bias, :mode]
+
+  @doc """
+  Fits a logistic regression model for sample inputs `x` and
+  sample targets `y`. Depending on number of classes the function chooses
+  either binary or multinomial logistic regression.
+  """
+  defn(fit(x, y, opts \\ []), do: fit_verify(x, y, opts))
```
polvalente, line 16:
```suggestion
  defn fit(x, y, opts \\ []), do: fit_verify(x, y, opts)
```

polvalente: If this doesn't work, we need to fix the .formatter.exs to use import_deps like EXLA and Torchx do

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,192 @@
...
+    if Nx.rank(x.shape) != 2 do
+      raise ArgumentError, "Training vector must be two-dimensional (n_samples, n_features)"
+    end
+
+    if (Nx.rank(y.shape) != 1 and !opts[:one_hot]) or (Nx.rank(y.shape) != 2 and opts[:one_hot]) do
+      raise ArgumentError,
+            "Target vector must be one-dimensional (n_samples) or two-dimensional if :one_hot set to true"
+    end
+
+    if is_number(opts[:lr]) and opts[:lr] < 0 do
```
polvalente, line 34: this is asserting on "non-negative" instead of positive. We need to either fix the message below or the comparison with 0 here

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,192 @@
...
+      raise ArgumentError, "Learning rate must be a positive number"
+    end
+
+    if is_integer(opts[:iterations]) and opts[:iterations] <= 0 do
```
polvalente, line 38:
```suggestion
    unless is_integer(opts[:iterations]) and opts[:iterations] > 0 do
```

lib/scholar/linear/logistic_regression.ex
polvalente, line 34:
```suggestion
    unless is_number(opts[:lr]) and opts[:lr] >= 0 do
```

polvalente: `[lr: :atom]` passes the current check without raising. Same issue with `iterations` below

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,192 @@
...
+  Fits a logistic regression model for sample inputs `x` and
+  sample targets `y`. Depending on number of classes the function chooses
+  either binary or multinomial logistic regression.
+  """
+  defn(fit(x, y, opts \\ []), do: fit_verify(x, y, opts))
+
+  # Function checks validity of the provided data
+
+  deftransformp fit_verify(x, y, opts) do
+    if !is_integer(opts[:num_classes]) or opts[:num_classes] < 1 do
```
polvalente, line 21: you should use `keyword!` to assert on all the valid options here

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,192 @@
...
+    if is_number(opts[:lr]) and opts[:lr] < 0 do
+      raise ArgumentError, "Learning rate must be a positive number"
+    end
+
+    if is_integer(opts[:iterations]) and opts[:iterations] <= 0 do
+      raise ArgumentError, "Number of iterations must be a positive integer"
+    end
+
+    if opts[:num_classes] < 3, do: fit_binary(x, y, opts), else: fit_multinomial(x, y, opts)
+  end
```
polvalente, line 43: you need to call `fit_binary` and `fit_multinomial` outside of this deftransform, in the body for `fit`, to minimize how much time we spend inside transforms

lib/scholar/linear/logistic_regression.ex
msluszniak: I don't agree with the statement that logistic regression is not a linear model. https://scikit-learn.org/stable/modules/linear_model.html#logistic-regression and https://stats.stackexchange.com/questions/93569/why-is-logistic-regression-a-linear-classifier.

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,192 @@
...
+      |> then(&(1 / &1))
+
+    diff =
+      logit
+      |> Nx.subtract(Nx.transpose(y))
+      |> Nx.reshape({m})
+
+    coeff_diff =
+      x
+      |> Nx.transpose()
```
polvalente, line 130: I think it would be nice to try and avoid these transposes. Perhaps there's a way to get `x` in the desired form beforehand?

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,192 @@
...
+      x
+      |> Nx.dot(coeff)
+
+    prob = softmax(dot_prod)
+
+    diff =
+      one_hot
+      |> Nx.subtract(prob)
+      |> then(&Nx.dot(Nx.transpose(x), &1))
+      |> Nx.multiply(-1)
```
polvalente, line 160: It seems to me that you can avoid this `|> multiply(-1)` by reversing the args for subtract:

(transpose(x) . (one_hot - prob)) * -1 == transpose(x) . (prob - one_hot)

lib/scholar/linear/logistic_regression.ex
msluszniak: Yes, you're right. I'll set this to non-negative.

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,192 @@
...
+      |> Nx.divide(m)
```
polvalente, line 161: either that or absorb the negation into `-m` here

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,192 @@
...
+    logit =
+      coeff
+      |> Nx.dot(Nx.transpose(x))
+      |> Nx.add(bias)
+      |> Nx.multiply(-1)
+      |> Nx.exp()
+      |> Nx.add(1)
+      |> then(&(1 / &1))
+
+    Nx.select(Nx.greater(logit, 0.5), 1, 0)
```
polvalente, line 184:
```suggestion
    Nx.greater(logit, 0.5)
```

polvalente: The results are already 0 and 1 if you just return the selector

lib/scholar/linear/logistic_regression.ex
msluszniak: Ok, I'll leave only validity checks inside transforms.

mix.exs
```diff
@@ -21,7 +21,9 @@ defmodule Scholar.MixProject do
   defp deps do
     [
       {:ex_doc, "~> 0.27", only: :dev, runtime: false},
-      {:nx, "~> 0.2.0"}
+      {:nx, github: "elixir-nx/nx", sparse: "nx"},
+      {:scidata, "~> 0.1.9"},
+      {:explorer, "~> 0.2.0"}
```
polvalente, line 26:
```suggestion
      {:scidata, "~> 0.1.9", only: [:test, :dev]},
      {:explorer, "~> 0.2.0", only: [:test, :dev]}
```

The code itself doesn't depend on them, does it?

polvalente: It would be nice if we could not depend on them at all. Perhaps the tests can use a subset of the data parsed manually?

lib/scholar/linear/logistic_regression.ex
msluszniak: Nice

mix.exs
msluszniak: Yes, only in the testing module at this point

lib/scholar/linear/logistic_regression.ex
polvalente: Thanks for the reference. The function is indeed not linear, but since it is reversible they make for Generalized Linear Models: https://scikit-learn.org/stable/modules/linear_model.html#generalized-linear-regression

What I meant by it not being linear is that for `f(x) = logistic(x)`, `f(x + y) != f(x) + f(y)`, which is one of the conditions for linearity (the other one where `f(k*x) = k*f(x)` is also violated)

polvalente: I'm still not totally comfortable with calling this a Linear model, but since it's already called that by SciKit, we can move forward with this namespace :)

polvalente:
Nice work with the PR!

lib/scholar/linear/logistic_regression.ex
msluszniak: I totally agree in the sense that this function does not satisfy the conditions of a linear transformation, but as you mentioned it has an inverse function and we can reformulate the problem.

lib/scholar/linear/logistic_regression.ex
msluszniak: Indeed, the formatter adds these parentheses.

lib/scholar/linear/logistic_regression.ex
msluszniak: I decided to precalculate transpositions once and use them later on.

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,195 @@
...
+
+  # Function checks validity of the provided data
+
+  deftransformp fit_verify(x, y, opts) do
+    if !is_integer(opts[:num_classes]) or opts[:num_classes] < 1 do
+      raise ArgumentError, "The number of classes must be a positive integer"
+    end
+
+    if Nx.rank(x.shape) != 2 do
+      raise ArgumentError, "Training vector must be two-dimensional (n_samples, n_features)"
```
jonatanklosko, line 29:
```suggestion
      raise ArgumentError, "expected x to have shape {n_samples, n_features}, got tensor with shape: #{inspect(Nx.shape(x))}"
```

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,195 @@
...
+    end
+
+    if (Nx.rank(y.shape) != 1 and !opts[:one_hot]) or (Nx.rank(y.shape) != 2 and opts[:one_hot]) do
+      raise ArgumentError,
+            "Target vector must be one-dimensional (n_samples) or two-dimensional if :one_hot set to true"
+    end
```
jonatanklosko, line 35: Maybe instead of requiring the user pass `:one_hot` we could determine it based on whether `y` is rank-1 or rank-2?

Also, is there a good reason for supporting one-hot here? The reason why the user would have the data in one-hot is if they encoded as such in the first place, or is there another reason? Is this something `scikit` does?

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,195 @@
...
+
+    if (!is_nil(opts[:lr]) and !is_number(opts[:lr])) or (is_number(opts[:lr]) and opts[:lr] <= 0) do
```
jonatanklosko, line 37:
```suggestion
    unless is_number(opts[:lr]) and opts[:lr] > 0 do
```

Note that we don't need to check nil, because we should call `keyword!` beforehand and it will set a default.

lib/scholar/linear/logistic_regression.ex
msluszniak: Well, I tried to test my algorithms on data with a text-categorical label. In Explorer, at this point, I cannot find something like OrdinalEncoder from SciKit. But Explorer got dummies, so I use them. Frankly, when I think about this now, I don't think we need to offer this option, and I can get rid of it.

msluszniak: But this ordinalEncoder is just dummies + argmax

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,193 @@
...
+      |> Nx.dot(diff)
+      |> Nx.divide(m)
+
+    bias_diff =
+      diff
+      |> Nx.sum()
+      |> Nx.divide(m)
+
+    new_coeff = Nx.subtract(coeff, Nx.multiply(coeff_diff, lr))
+    new_bias = Nx.subtract(bias, Nx.multiply(bias_diff, lr))
```
jonatanklosko, line 142:
```suggestion
    new_coeff = coeff - coeff_diff * lr
    new_bias = bias - bias_diff * lr
```

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,193 @@
...
+    coeff = Nx.broadcast(Nx.tensor(0, type: {:f, 32}), {n})
+
+    {_, _, _, _, _, _, final_coeff, final_bias} =
+      while {iter = 0, x, lr, iterations, x_t, y_t, coeff, bias = Nx.tensor(0, type: {:f, 32})},
+            Nx.less(iter, iterations) do
+        {coeff, bias} = update_coefficients(x, x_t, y_t, {coeff, bias}, lr)
+        {iter + 1, x, lr, iterations, x_t, y_t, coeff, bias}
+      end
+
+    %__MODULE__{coefficients: final_coeff, bias: final_bias, mode: 0}
```
jonatanklosko, line 75: `0 | 1` -> `:binary | :multinomial` :)

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,193 @@
...
+  targets `y`.
+
+  Depending on number of classes the function chooses either binary
+  or multinomial logistic regression.
+  """
+  defn fit(x, y, opts \\ []) do
+    opts =
+      keyword!(opts,
+        iterations: 1000,
+        lr: 0.01,
```
jonatanklosko, line 21: Let's change `:lr` to `:learning_rate` for clarity :)

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,193 @@
...
+        num_classes: 1
+      )
```
jonatanklosko, line 23: We should document the options.

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,193 @@
...
+
+  Depending on number of classes the function chooses either binary
+  or multinomial logistic regression.
+  """
+  defn fit(x, y, opts \\ []) do
+    opts =
+      keyword!(opts,
+        iterations: 1000,
+        lr: 0.01,
+        num_classes: 1
```
jonatanklosko, line 22: Maybe we should require the user to pass `:num_classes` as in metrics, otherwise the default only works for specific `y`s, wdyt?

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,193 @@
...
+      x
+      |> Nx.dot(coeff)
+
+    prob = softmax(dot_prod)
+
+    diff =
+      prob
+      |> Nx.subtract(one_hot)
+      |> then(&Nx.dot(x_t, &1))
+      |> Nx.divide(m)
```
jonatanklosko, line 162:
```suggestion
    diff = Nx.dot(x_t, prob - one_hot) / m
```

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,193 @@
...
+
+    Nx.subtract(coeff, Nx.multiply(diff, lr))
```
jonatanklosko, line 164:
```suggestion
    coeff - diff * lr
```

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,193 @@
...
+
+  # Binary logistic regression
+
+  defnp fit_binary(x, y, opts \\ []) do
+    iterations = opts[:iterations]
+    lr = opts[:lr]
+    x_t = Nx.transpose(x)
+    y_t = Nx.transpose(y)
+
+    {_m, n} = x.shape
```
jonatanklosko, line 65:
```suggestion
    {_m, n} = Nx.shape(x)
```

better to rely on the public function :)

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,193 @@
...
+      end
+
+    %__MODULE__{coefficients: final_coeff, bias: Nx.tensor(0, type: {:f, 32}), mode: 1}
+  end
+
+  # Normalized softmax
+
+  defnp softmax(t) do
+    normalized = t - Nx.reduce_max(t, axes: [1], keep_axes: true)
+    Nx.transpose(Nx.transpose(Nx.exp(normalized)) / Nx.sum(Nx.exp(normalized), axes: [1]))
```
jonatanklosko, line 109: Let's extract `Nx.exp(normalized)` to a separate variable. I'm not sure if defn compiler would optimise it out, but we would definitely compute it twice when doing the operations individually with an Nx backend.

jonatanklosko: Also, can we transpose/reshape the divisor, instead of transposing twice?

jonatanklosko (commented):
:cat2:

lib/scholar/linear/logistic_regression.ex
msluszniak: Sure, but how to require a field in the opts? ```:num_classes``` needs to be passed by the opts

jonatanklosko: Same as in `Metrics`, we just raise if not present.

lib/scholar/linear/logistic_regression.ex
msluszniak: How to compare atoms in defn predict then?

jonatanklosko: We can always use a transform, but recently `case` was added as `defn` primitive and it supports atoms, so this should work:

```elixir
case mode do
  :binary -> predict_binary(model, x)
  :multinomial -> predict_multinomial(model, x)
end
```

lib/scholar/linear/logistic_regression.ex
msluszniak: I think this will be ok
```elixir 
 defnp softmax(t) do
    normalized_exp = t - Nx.reduce_max(t, axes: [0], keep_axes: true) |> Nx.exp()
    normalized_exp / Nx.sum(normalized_exp, axes: [0])
  end
```

lib/scholar/linear/logistic_regression.ex
msluszniak: Is this documentation sufficient @jonatanklosko?

lib/scholar/linear/logistic_regression.ex
```diff
@@ -0,0 +1,197 @@
...
+  Fits a logistic regression model for sample inputs `x` and sample
+  targets `y`.
+
+  Depending on number of classes the function chooses either binary
+  or multinomial logistic regression.
+
+  ## Options
+    * `:num_classes` - Number of classes contained in the input tensors
+    * `:learning_rate` - Learning rate used in logistic regression
+    * `:iterations` - Number of iterations of gradient descent performed inside logistic regression
```
jonatanklosko, line 20:
```suggestion
  ## Options
  
    * `:num_classes` - number of classes contained in the input tensors. Required

    * `:learning_rate` - learning rate used by gradient descent. Defaults to `0.01`

    * `:iterations` - number of iterations of gradient descent performed inside logistic
      regression. Defaults to `1000`

```

test/scholar/linear/logistic_regression_test.exs
```diff
@@ -0,0 +1,168 @@
...
+    test "Number of classes is not a positive integer II" do
+      x = Nx.tensor([[1, 2], [3, 4]])
+      y = Nx.tensor([1, 2])
+
+      assert_raise ArgumentError,
+                   "expected :num_classes to be a positive integer, got: 2.0",
+                   fn ->
+                     Scholar.Linear.LogisticRegression.fit(x, y, num_classes: 2.0)
+                   end
+    end
```
jonatanklosko, line 114:
```suggestion
    test "raises on invalid :num_classes" do
      x = Nx.tensor([[1, 2], [3, 4]])
      y = Nx.tensor([1, 2])

      assert_raise ArgumentError, "expected :num_classes to be a positive integer, got: -3", fn ->
        Scholar.Linear.LogisticRegression.fit(x, y, num_classes: -3)
      end

      assert_raise ArgumentError,
                   "expected :num_classes to be a positive integer, got: 2.0",
                   fn ->
                     Scholar.Linear.LogisticRegression.fit(x, y, num_classes: 2.0)
                   end
    end
```

test/scholar/linear/logistic_regression_test.exs
```diff
@@ -0,0 +1,168 @@
+defmodule Scholar.Linear.LinearRegressionTest do
+  use ExUnit.Case, async: true
+
+  describe "Logistic Regression" do
```
jonatanklosko, line 4: We don't need this block, since everything here is about logistic regression :D

polvalente: The module name is incorrect, by the way
LinearRegression -> LogisticRegression

jonatanklosko (approved):
Great!

@polvalente feel free to merge whenever you're happy :)

polvalente (approved):

PR #20, opened 2022-08-17
Add jaccard similarity metric

tiagodavi:
Calculate Jaccard similarity. I'll create a PR to calculate dissimilarity as well.


lib/scholar/metrics/similarity.ex
```diff
@@ -0,0 +1,61 @@
...
+        0.25
+      >
+
+      iex> x = Nx.tensor([1, 2])
+      iex> y = Nx.tensor([1, 2, 3])
+      iex> Scholar.Metrics.Similarity.jaccard x, y
+      ** (ArgumentError) expected input shapes to be equal, got {2} != {3}
+  """
+  defn jaccard(x, y) do
+    assert_same_shape!(x, y)
```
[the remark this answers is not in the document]
polvalente: Perhaps it would be a good idea to support rank N tensors where the similarity is calculated over the last axis?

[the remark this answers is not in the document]
tiagodavi: I can take a look during the week.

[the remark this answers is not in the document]
polvalente: With this proposal, the shape would need to be the same except on the last axis, I think

lib/scholar/metrics/similarity.ex
```diff
@@ -0,0 +1,61 @@
+defmodule Scholar.Metrics.Similarity do
+  @moduledoc """
+  Similarity metrics between 1-D tensors.
+  """
+
+  import Nx.Defn
+  import Scholar.Shared
+
+  @doc """
+  The Jaccard similarity (also known as Jaccard similarity coefficient, or Jaccard index)
```
jonatanklosko, line 10: As a convention we always keep the first paragraph a single sentence. So I lets change to

```
Calculates Jaccard similarity (also known as ...).

Jaccard similarity is a statistic ...

File: lib/scholar/metrics/similarity.ex [hunk 3]
```diff
@@ -0,0 +1,61 @@
...
+  is a statistic used to measure similarities between two sets. Mathematically, the calculation
+  of Jaccard similarity is the ratio of set intersection over set union.
+
+  J(A, B) = |A∩B| / |A∪B|
```
jonatanklosko, line 14: You can write this between a pair of `$$` and use the LaTeX syntax :)

lib/scholar/metrics/similarity.ex
```diff
@@ -0,0 +1,61 @@
...
+      iex> y = Nx.tensor([5.0, 2.5, 3.1, 9.0])
+      iex> Scholar.Metrics.Similarity.jaccard(x, y)
+      #Nx.Tensor<
+        f32
+        0.1428571492433548
+      >
+
+      iex> x = Nx.tensor([1,2,3,5,7])
+      iex> y = Nx.tensor([1,2,4,8,9])
+      iex> Scholar.Metrics.Similarity.jaccard x, y
```
jonatanklosko, line 28:
```suggestion
      iex> Scholar.Metrics.Similarity.jaccard(x, y)
```

Same below.

lib/scholar/metrics/similarity.ex
```diff
@@ -0,0 +1,61 @@
...
+      iex> x = Nx.tensor([1.0, 5.0, 3.0, 6.7])
+      iex> y = Nx.tensor([5.0, 2.5, 3.1, 9.0])
+      iex> Scholar.Metrics.Similarity.jaccard(x, y)
+      #Nx.Tensor<
+        f32
+        0.1428571492433548
+      >
+
+      iex> x = Nx.tensor([1,2,3,5,7])
+      iex> y = Nx.tensor([1,2,4,8,9])
```
jonatanklosko, line 27:
```suggestion
      iex> y = Nx.tensor([1, 2, 4, 8, 9])
```

lib/scholar/metrics/similarity.ex
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  defnp unique_size(tensor) do
+    sorted = Nx.sort(tensor)
+
+    [
+      Nx.not_equal(sorted[0..-2//1], sorted[1..-1//1]),
+      Nx.tensor([1])
+    ]
+    |> Nx.concatenate()
```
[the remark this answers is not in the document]
tiagodavi: What I did didn't work. What's am I missing? 

```
 [
      Nx.not_equal(sorted[0..-2//1], sorted[1..-1//1]),
      Nx.tensor([1])
    ]
    |> Nx.add(1)
```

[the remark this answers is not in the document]
polvalente: You need to remove the list and the Nx.tensor([1]) call as well. 

You also need to keep the Nx.sum before Nx.add(1)

lib/scholar/metrics/similarity.ex
```diff
@@ -0,0 +1,61 @@
...
+      ** (ArgumentError) expected input shapes to be equal, got {2} != {3}
+  """
+  defn jaccard(x, y) do
+    assert_same_shape!(x, y)
+
+    x_size = unique_size(x)
+    y_size = unique_size(y)
+
+    union_size = unique_size(Nx.concatenate([x, y]))
+    intersection_size = Nx.add(x_size, y_size) - union_size
```
jonatanklosko, line 46:
```suggestion
    intersection_size = (x_size + y_size) - union_size
```

lib/scholar/metrics/similarity.ex
```diff
@@ -0,0 +1,61 @@
...
+
+    Nx.divide(intersection_size, union_size)
```
jonatanklosko, line 48:
```suggestion
    intersection_size / union_size
```

lib/scholar/metrics/similarity.ex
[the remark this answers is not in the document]
tiagodavi: Not sure how to do it here. What's your suggestion @polvalente ?

lib/scholar/metrics/similarity.ex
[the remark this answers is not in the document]
tiagodavi: done

lib/scholar/metrics/similarity.ex
tiagodavi: done

lib/scholar/metrics/similarity.ex
tiagodavi: done

lib/scholar/metrics/similarity.ex
tiagodavi: done

lib/scholar/metrics/similarity.ex
tiagodavi: done

lib/scholar/metrics/similarity.ex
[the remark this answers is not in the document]
tiagodavi: It appears same shape is required here: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.jaccard_score.html

but it allows multi dimensions.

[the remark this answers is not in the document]
polvalente: We can keep requiring the same shape. Please add a comment similar to: "We're requiring the same shape because usual use cases will have the same shape. The last axis could in theory be different on both sides"

lib/scholar/metrics/similarity.ex
```diff
@@ -0,0 +1,61 @@
+defmodule Scholar.Metrics.Similarity do
+  @moduledoc """
+  Similarity metrics between 1-D tensors.
+  """
+
+  import Nx.Defn
+  import Scholar.Shared
+
+  @doc """
```
jonatanklosko, line 9:
```suggestion
  @doc ~S"""
```

otherwise the LaTeX back slashes will be interpreted as string escapes.

lib/scholar/metrics/similarity.ex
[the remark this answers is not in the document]
tiagodavi: comment added

lib/scholar/metrics/similarity.ex
tiagodavi: fixed

tiagodavi:
@polvalente @jonatanklosko It's missing a case for binary attributes (the formula would be different for this case) .. What do you think of adding it as option?

type: :binary or something.. 

```
iex(1)> x = Nx.tensor [1,0,0,1,1,1]
#Nx.Tensor<
  s64[6]
  [1, 0, 0, 1, 1, 1]
>
iex(2)> y = Nx.tensor [0,0,1,1,1,0]
#Nx.Tensor<
  s64[6]
  [0, 0, 1, 1, 1, 0]
>
iex(3)> Scholar.Metrics.Similarity.jaccard(x, y)
#Nx.Tensor<
  f32
  1.0
>
```

It's was supposed to return 0.4 instead..

polvalente:
@tiagodavi I think adding a `:type` option is fine. What would be the correct formula for this case?

jonatanklosko:
The binary case has a different semantics and implementation, so I would have it separately as `binary_jaccard`. I believe this is similar to what we already do in `Scholar.Metrics` :)

lib/scholar/metrics/similarity.ex
```diff
@@ -0,0 +1,63 @@
...
+    union_size = unique_size(Nx.concatenate([x, y]))
+    intersection_size = x_size + y_size - union_size
+
+    intersection_size / union_size
+  end
+
+  defnp unique_size(tensor) do
+    sorted = Nx.sort(tensor)
+
+    Nx.not_equal(sorted[0..-2//1], sorted[1..-1//1])
```
jonatanklosko, line 59: This is going to fail for tensors with length 1, we should match on the shape and for `{1}` we return `1` :)

jonatanklosko: Also worth a test!

tiagodavi: To be honest I don't even know what this piece of code is doing.. 

`Nx.not_equal(sorted[0..-2//1], sorted[1..-1//1]) `??

jonatanklosko: My point is that we are accessing index 1, which is not valid for length-1 tensor.

This code creates a binary mask meaning "is an element at this position different from its successor". For example given `[3, 2, 1]` we want to compare `[3, 2]` with their successors `[2, 1]` (the last element has no successor), so we need all-but-last and all-but-first slices of the original tensor.

tiagodavi: I was referring to: `0..-2//1`. this range is odd to me.

tiagodavi: @jonatanklosko changes added

polvalente: @tiagodavi negative indices are a way of indexing from right to left instead of left to right. -2 is the second position from the right (a.k.a second to last item). For a tensor of size 5, index -2 is the same as index 3, index -1 is the same as index 4, which is the last index.

polvalente: So 0..-2//1 is a way of saying "take all items from the first until the second to last, inclusive"

polvalente (approved):

polvalente:
I'll leave this open for a last pass on Jonatan's part

lib/scholar/metrics/similarity.ex
tiagodavi: It’s similar to list index, however with double slash and a number at the end.

tiagodavi: Thanks

test/scholar/metrics/similarity_test.exs
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    test "returns 20% of similarity" do
+      x = Nx.tensor([1, 2, 3])
+      y = Nx.tensor([3, 4, 5])
+
+      assert Similarity.jaccard(x, y) == Nx.tensor(0.20)
+    end
+
+    test "returns similarity when tensor has single shape" do
```
jonatanklosko, line 35:
```suggestion
    test "returns similarity when tensors have a single element" do
```

lib/scholar/metrics/similarity.ex
```diff
@@ -0,0 +1,75 @@
...
+    intersection_size = x_size + y_size - union_size
+
+    intersection_size / union_size
+  end
+
+  defnp unique_size(%Nx.Tensor{shape: shape} = tensor) do
+    transform(
+      shape,
+      fn
+        {} ->
```
jonatanklosko, line 60: You should be able to use `case` directly, like:

```elixir
case Nx.shape(tensor) do
  {1} ->
    ...
end
```

jonatanklosko:
Looks good, just a couple final comments!

lib/scholar/metrics/similarity.ex
tiagodavi: it does not work..

tiagodavi:
@jonatanklosko added

lib/scholar/metrics/similarity.ex
tiagodavi:
```
== Compilation error in file lib/scholar/metrics/similarity.ex ==
** (CompileError) lib/scholar/metrics/similarity.ex:57: invalid numerical expression:

    case Nx.shape(tensor) do
      {} ->
        raise "expected input shape of at least {1}, got: {}"
```

tiagodavi: @polvalente it works! thank you

lib/scholar/metrics/similarity.ex
```diff
@@ -0,0 +1,74 @@
...
+    intersection_size / union_size
+  end
+
+  defnp unique_size(%Nx.Tensor{shape: shape} = tensor) do
+    case shape do
+      {} ->
+        raise "expected input shape of at least {1}, got: {}"
+
+      {1} ->
+        Nx.tensor(1)
```
jonatanklosko, line 62:
```suggestion
        1
```

jonatanklosko (approved):
:shipit:

tiagodavi:
@polvalente @jonatanklosko Can I have my name Assigned to the PR?

PR #21, opened 2022-08-17
Implement K-means clustering

msluszniak:


lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,325 @@
...
+  @derive {Nx.Container, containers: [:num_iterations, :clusters, :inertia, :labels]}
+  defstruct [:num_iterations, :clusters, :inertia, :labels]
+
+  @doc """
+  Fits a K-Means model for sample inputs `x`
+
+
+  ## Options
+
+    * `:num_clusters` - The number of clusters to form as well as the number of centroids to generate. Defaults to `8`.
```
polvalente, line 17: I would make `num_clusters` required. Also, perhaps this doc could emphasize this is the `K` in `K-means`

lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,325 @@
...
+        max_iterations: 300,
+        num_runs: 10,
+        tol: 1.0e-4,
+        random_state: nil,
+        init: :k_means_plus_plus,
+        weights: nil
+      )
+
+    verify(x, opts)
+    x = x |> Nx.as_type({:f, 32})
```
polvalente, line 62: Do we need to forcibly upcast here?

polvalente: Actually, this could downcast if the type is already f64. I would try to avoid this `as_type` here

lib/scholar/cluster/k_means.ex
msluszniak: Yes, I've been thinking about requiring num_clusters. In sci-kit learn 8 is the default, but the required parameter also makes sense.

lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,325 @@
...
+  defnp k_means_plus_plus(x, num_clusters, num_runs) do
+    inf = Nx.Constants.infinity()
+    {num_samples, num_features} = Nx.shape(x)
+    centroids = Nx.broadcast(inf, {num_runs, num_clusters, num_features})
+    inertia = Nx.broadcast(0.0, {num_runs, num_samples})
+    centroid_mask = Nx.broadcast(1, {num_runs, num_samples})
+
+    first_centroid_idx = Nx.random_uniform({num_runs}, 0, num_samples - 1, type: {:u, 32})
+
+    first_centroid = Nx.take(x, first_centroid_idx) |> Nx.reshape({:auto})
```
polvalente, line 187:
```suggestion
    first_centroid = Nx.take(x, first_centroid_idx) |> Nx.flatten()
```

polvalente: there are similar occurences below as well

lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,325 @@
...
+
+    indices_centroids =
+      Nx.concatenate([
+        Nx.new_axis(Nx.reshape(Nx.iota({num_runs, num_features}, axis: 0), {:auto}), 0),
+        Nx.new_axis(Nx.broadcast(0, {num_runs * num_features}), 0),
+        Nx.new_axis(Nx.tile(Nx.iota({num_features}), [num_runs]), 0)
+      ])
```
polvalente, line 194: Can't this be Nx.stack instead?

Also, if we can avoid the transpose it would be better

lib/scholar/cluster/utils.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Scholar.Cluster.Utils do
+  import Nx.Defn
+  import Scholar.Shared
+
+  @spec squared_euclidean(Nx.t(), Nx.t()) :: Nx.t()
+  defn squared_euclidean(x, y, opts \\ []) do
```
polvalente, line 6: We already have this function under Scholar.Metrics.Distance

msluszniak: But not with axes specification

polvalente: You can add it there instead

msluszniak: I guess that we should add this option to all the other metrics to keep API consistent then, but it's an idea.

polvalente: I agree. Perhaps it's better that this is done in a separate PR first

lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,322 @@
...
+  Fits a K-Means model for sample inputs `x`
+
+
+  ## Options
+
+    * `:num_clusters` - The number of clusters to form as well as the number of centroids to generate. Defaults to `8`.
+
+    * `:max_iterations` - Maximum number of iterations of the k-means algorithm for a single run. Defaults to `300`.
+
+    * `:num_runs` - Number of time the k-means algorithm will be run with different centroid seeds. The final results will be the best output of num_runs runs in terms of inertia. Defaults to `10`.
```
josevalim, line 21: Sorry for the nitpick but can you please trim the descriptions so the lines do not exceed 100 chars?

lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,322 @@
...
+      raise ArgumentError,
+            "expected :random_state to be an integer or nil, got: #{inspect(opts[:random_state])}"
+    end
+
+    if opts[:random_state] != nil do
+      :rand.seed(:exsss, opts[:random_state])
+    end
+  end
+
+  #
```
josevalim, line 278:
```suggestion
```

lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,322 @@
...
+  @doc """
+  Makes predictions with the given model on inputs `x`. It returns a tensor with clusters corresponding to the input.
```
josevalim, line 280:
```suggestion
  Makes predictions with the given model on inputs `x`.

  It returns a tensor with clusters corresponding to the input.
```

test/scholar/cluster/k_means_test.exs
```diff
@@ -0,0 +1,152 @@
+defmodule Scholar.Cluster.KMeansTest do
+  use ExUnit.Case, async: true
+
+  test "Simple test without weights" do
```
josevalim, line 4:
```suggestion
  test "without weights" do
```

lib/scholar/cluster/utils.ex
josevalim: @msluszniak please change `Scholar.Metrics.Distance` to support the `:axes` option. :)

lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,322 @@
+defmodule Scholar.Cluster.KMeans do
+  @moduledoc """
+  K-Means algorithm
```
josevalim, line 3:
```suggestion
  K-Means algorithm.
```

josevalim (approved):
This looks great @msluszniak!

Do you think you can also add a notebook that shows how to use this API, including some plottings? :)

msluszniak:
> 

Sure, that's a great idea! :)

lib/scholar/cluster/k_means.ex
```diff
@@ -14,21 +14,28 @@ defmodule Scholar.Cluster.KMeans do
 
   ## Options
 
-    * `:num_clusters` - The number of clusters to form as well as the number of centroids to generate. Defaults to `8`.
+    * `:num_clusters` - The number of clusters to form as well as the number of centroids to generate. Required.
```
josevalim, line 17: Btw, why did you make them required? If 8 is a random number, perhaps a better default is 2? Or does Scipy make them required?

polvalente: I suggested making the number of clusters required mainly because since any choice we provide as default is arbitrary, I didn't see a reason for it being optional.

josevalim: I think the least arbitrary is 2 but I am fine with it being required or following Scipy.

lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,337 @@
...
+            "expected :tol to be a non-negative number, got: #{inspect(opts[:tol])}"
+    end
+
+    unless is_integer(opts[:random_state]) or is_nil(opts[:random_state]) do
+      raise ArgumentError,
+            "expected :random_state to be an integer or nil, got: #{inspect(opts[:random_state])}"
+    end
+
+    if opts[:random_state] != nil do
+      :rand.seed(:exsss, opts[:random_state])
```
jonatanklosko, line 288: This will only impact the BinaryBackend and this happens at defn compile time, which we don't want. I think we need to wait for https://github.com/elixir-nx/nx/pull/846 to do this sort of things.

As for tests that rely on this, maybe we could use data points that guarantee convergence, regardless the initialization. And/or have some property tests if there are any we can reasonably have.

lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,337 @@
...
+
+    * `:tol` - Relative tolerance with regards to Frobenius norm of the difference in the cluster centers of two
+      consecutive iterations to declare convergence. Defaults to `1e-4`.
+
+    * `:random_state` - Determines random number generation for centroid initialization.
+      Use an int to make the randomness deterministic.
+      The argument `nil` is special and means that the seed is not set. Defaults to `nil`.
+
+    * `:weights` - The weights for each observation in x. If equals to `nil`, all observations
+      are assigned equal weight.
```
jonatanklosko, line 32: It seems like this should be a tensor argument, as we convert it to a tensor anyway. Since it is one weight per each data point in `x`, the weights may be huge and we don't want to embed all of that as a constant into the computation.

One option is:

```elixir
defn fit(x, opts \\ []) do
  weights = Nx.broadcast(1, ...)
  fit(x, weights, opts)
end

defn fit(x, weights, opts) do
  ...
end
```

@polvalente @josevalim thoughts?

lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,337 @@
...
+
+    * `:init` - Method for initialization (Defaults to `:k_means_plus_plus`):
+
+        * `:k_means_plus_plus` -  selects initial cluster centroids using sampling based on an empirical probability
+         distribution of the points’ contribution to the overall inertia.
+         This technique speeds up convergence, and is theoretically proven to be O(log(k))-optimal.
+
+        * `:random` - choose :num_clusters observations (rows) at random from data for the initial centroids.
```
jonatanklosko, line 40:
```suggestion
    * `:init` - Method for centroid initialization, either of:

        * `:k_means_plus_plus` (default) -  selects initial cluster centroids using sampling based on
          an empirical probability distribution of the points’ contribution to the overall inertia.
         This technique speeds up convergence, and is theoretically proven to be O(log(k))-optimal.

        * `:random` - choose `:num_clusters` observations (rows) at random from data for the initial centroids.
```

lib/scholar/cluster/k_means.ex
jonatanklosko: FTR we can support tensors in options by having `fit/2` as a transform and let it call the `fit/3` defn. We will worry about this later.

lib/scholar/cluster/k_means.ex
msluszniak: The problem will be only with the labels parameter. Depending on the initialization of centroids the resultant labels might be i. ex. [1, 0, 1, 0], but also [0, 1, 0, 1].

jonatanklosko: You can normalize he classes for comparison.

```elixir
defmodule Scholar.Utils do
  import Nx.Defn

  @doc """
  Returns labels that unambiguously identify the given grouping.

  This function effectively renumbers the classes, such that the
  first occurrences of each classes are in increasing order.
  
  ## Examples

      iex> Scholar.Utils.normalize_labels(Nx.tensor([1, 0, 1, 0]), num_classes: 2)
      #Nx.Tensor<
        s64[4]
        [0, 1, 0, 1]
      >

      iex> Scholar.Utils.normalize_labels(Nx.tensor([1, 2, 2, 0, 1, 2, 0]), num_classes: 3)
      #Nx.Tensor<
        s64[7]
        [0, 1, 1, 2, 0, 1, 2]
      >

  """
  defn normalize_labels(labels, opts \\ []) do
    opts = keyword!(opts, [:num_classes])

    num_classes = opts[:num_classes]

    classes = Nx.iota({num_classes})

    class_min_idx =
      Nx.equal(
        Nx.new_axis(classes, 1),
        Nx.new_axis(labels, 0)
      )
      |> Nx.argmax(axis: 1)

    mapping = to_consecutive(class_min_idx)

    Nx.take(mapping, labels)
  end

  # This is effectively a double argsort, just optimised
  defnp to_consecutive(x) do
    x = Nx.argsort(x)
    Nx.indexed_add(Nx.broadcast(0, x), Nx.new_axis(x, -1), Nx.iota(x))
  end
end
```

test/scholar/cluster/k_means_test.exs
```diff
@@ -0,0 +1,171 @@
...
+  test "with weights" do
+    model =
+      Scholar.Cluster.KMeans.fit(Nx.tensor([[1, 2], [2, 4.25], [1, 3], [2, 5]]),
+        num_clusters: 2,
+        weights: [1, 2, 3, 4]
+      )
+
+    {oredered_clusters, ordered_labels} = sort_clusters(model.clusters, model.labels)
+    model = %{model | :labels => ordered_labels}
+    model = %{model | :clusters => oredered_clusters}
```
jonatanklosko, line 45:
```suggestion
    model = %{model | labels: ordered_labels, clusters: oredered_clusters}
```

Also, let's do that in `sort_clusters`, so it's shared across tests :)

lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,323 @@
...
+
+    centroids = initialize_centroids(x, opts)
+    distance = Nx.broadcast(inf, {num_runs})
+
+    {i, _, _, _, _, _, _, _, final_centroids, nearest_centroids} =
+      while {i = 0, x, previous_iteration_centroids = Nx.broadcast(inf, centroids), distance,
+             inertia = Nx.broadcast(0.0, {num_runs}), weights, broadcast_weights, broadcast_x,
+             centroids, nearest_centroids = Nx.broadcast(-1, {num_runs, num_samples})},
+            i < opts[:max_iterations] and
+              Nx.all(distance > opts[:tol]) do
```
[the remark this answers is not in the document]
msluszniak: I missed that, let me fix this

msluszniak:
I added a livebook with a few examples of KMeans usage. In the livebook, I implemented the Silhouette module which, in my opinion, might be added to metrics in scholar, what do you think?

josevalim:
> In the livebook, I implemented the Silhouette module which, in my opinion, might be added to metrics in scholar, what do you think?

Do we have similar metrics in sklearn? If we do, we should go ahead and add them here as well. :+1:

msluszniak:
> 

https://scikit-learn.org/stable/modules/generated/sklearn.metrics.silhouette_score.html and https://scikit-learn.org/stable/modules/generated/sklearn.metrics.silhouette_samples.html#sklearn.metrics.silhouette_samples so yes, they have these functionalities

josevalim:
@msluszniak fantastic, so let’s add them too!!!

notebooks/k_means.livemd
```diff
@@ -0,0 +1,559 @@
...
+
+```elixir
+Mix.install(
+  [
+    {:stb_image, "~> 0.5.2"},
+    {:kino, github: "livebook-dev/kino", override: true},
+    {:nx, github: "elixir-nx/nx", sparse: "nx", override: true},
+    {:exla, "~> 0.3"},
+    {:explorer, "~> 0.2.0"},
+    {:scholar, path: "/home/mateusz/Dashbit/scholar"},
```
hugobarauna, line 11: I was trying to run this notebook on my machine and got a problem with this path.

Since we're using path option for a mix dependency, assuming that the person running this notebook is running it within the scholar git repo, maybe we could change that hardcoded path to a relative one, like this:

```elixir
{:scholar, path: Path.join(__DIR__, "..")},
```

notebooks/k_means.livemd
```diff
@@ -0,0 +1,559 @@
...
+## Clustering of pixels' colors
+
+The other interesting use case of KMeans clustering is pixel clustering. This technique replaces all pixels with similar colors (similar in terms of euclidean distance between RGB) with a centroid related to them.
+
+<!-- livebook:{"break_markdown":true} -->
+
+Let us start with loading the referral image.
+
+```elixir
+{:ok, img} = StbImage.read_file("../scholar/test/data/taj_mahal.jpg")
```
hugobarauna, line 223:
```suggestion
{:ok, img} = StbImage.read_file(Path.join(__DIR__, "../test/data/taj_mahal.jpg"))
```

jonatanklosko: @msluszniak are there reliable locations we can download the image and the csv file from? Ideally we would download them with `req`, so the notebook runs without depending on local files.

msluszniak: The CSV data is available [here](https://cam-orl.co.uk/facedatabase.html), with the image I think I should change it, cause I cannot find the usage rights.

notebooks/k_means.livemd
```diff
@@ -0,0 +1,559 @@
...
+## Clustering Face Images from Olivetti Dataset
+
+The last example is the clustering problem on Olivetti Dataset. The dataset consists of four hundred images 64 by 64 pixels. They present forty people with ten images per each one. Let's dive into this clustering problem.
+
+<!-- livebook:{"break_markdown":true} -->
+
+Firstly, load the data and cast it into Nx tensors.
+
+```elixir
+olivetti_data = Explorer.DataFrame.from_csv!("../scholar/test/data/olivetti_data.csv")
```
hugobarauna, line 303:
```suggestion
olivetti_data = Explorer.DataFrame.from_csv!(Path.join(__DIR__, "../test/data/olivetti_data.csv"))
```

notebooks/k_means.livemd
```diff
@@ -0,0 +1,559 @@
...
+olivetti_target = Explorer.DataFrame.from_csv!("../scholar/test/data/olivetti_target.csv")
```
hugobarauna, line 304:
```suggestion
olivetti_target = Explorer.DataFrame.from_csv!(Path.join(__DIR__, "../test/data/olivetti_target.csv"))
```

notebooks/k_means.livemd
msluszniak: @jonatanklosko maybe it will be better to add the Olivetti dataset to [Scidata](https://github.com/elixir-nx/scidata) module. What do you think?

jonatanklosko: Sounds good to me!

josevalim: @msluszniak yes, please do send a PR to Scidata. It also has some image datasets if you would prefer to use that instead. :)

msluszniak: @josevalim unfortunately there is a problem with Olivetti. The data source (sci-kit uses it; [link](https://github.com/scikit-learn/scikit-learn/blob/36958fb24/sklearn/datasets/_olivetti_faces.py#L32)) is saved as a .mat file. It would be problematic to parse this file into CSV. I think that I use fashionmnist instead.

josevalim: Ah, if fashionmnist can work that's great!

lib/scholar/cluster/k_means.ex
```diff
@@ -0,0 +1,308 @@
...
+  """
+  import Nx.Defn
+  import Scholar.Shared
+
+  @derive {Nx.Container, containers: [:num_iterations, :clusters, :inertia, :labels]}
+  defstruct [:num_iterations, :clusters, :inertia, :labels]
+
+  @doc """
+  Fits a K-Means model for sample inputs `x`
+
```
josevalim, line 13:
```suggestion
  Fits a K-Means model for sample inputs `x`.
```

josevalim (approved):
This is awesome @msluszniak, excellent job! :)

I have approved and someone else can merge it if they also agree. :)

jonatanklosko (approved):
Beautiful work with the notebook examples! I've just pushed some small changes :)

PR #22, opened 2022-08-21
Support batching in distance metrics

msluszniak:
As we need to use distances with aggregation over axes I propose changes to the current implementation.


cigrainger (commented):
I like this API a lot. Certainly beats calling a completely separate function for pairwise distances.

lib/scholar/metrics/distance.ex
```diff
@@ -35,16 +40,31 @@ defmodule Scholar.Metrics.Distance do
       iex> y = Nx.tensor([1, 2, 3])
       iex> Scholar.Metrics.Distance.euclidean(x, y)
       ** (ArgumentError) expected input shapes to be equal, got {2} != {3}
+
+      iex> x = Nx.tensor([[1,2], [3,4]])
+      iex> y = Nx.tensor([[8,3], [2,5]])
+      iex> Scholar.Metrics.Distance.euclidean(x, y, axes: [0])
+      #Nx.Tensor<
```
polvalente, line 47: Could you use a {3,2}-shaped example to make it more clear what `axes: [0]` means?

Also, a multi-axis example could be useful as well.

polvalente: This is also relevant to the other functions as well

lib/scholar/metrics/distance.ex
```diff
@@ -13,6 +13,11 @@ defmodule Scholar.Metrics.Distance do
   D(x, y) = \\sqrt{\\sum_i (x_i - y_i)^2}
   $$
 
+  ## Options
+
+    * `:axes` - Axes to aggregate distance over. If `:axes` set to `nil` then function does not aggregate distances.
```
polvalente, line 18: What does it mean to not aggregate distances?

polvalente: Perhaps an example with multi-dimensional inputs and `axes: nil` could be useful here

lib/scholar/metrics/distance.ex
```diff
@@ -198,17 +284,54 @@ defmodule Scholar.Metrics.Distance do
...
+      iex> y = Nx.tensor([[8,3], [2,5]])
+      iex> Scholar.Metrics.Distance.minkowski(x, y, p: 2.5, axes: [0])
+      #Nx.Tensor<
+        f32[2]
+        [7.021548271179199, 1.3195079565048218]
+      >
   """
-  @spec minkowski(Nx.t(), Nx.t(), integer()) :: Nx.t()
-  defn minkowski(x, y, p \\ 2) do
+  @spec minkowski(Nx.t(), Nx.t()) :: Nx.t()
```
polvalente, line 296:
```suggestion
  @spec minkowski(Nx.t(), Nx.t(), keyword()) :: Nx.t()
```

lib/scholar/metrics/distance.ex
```diff
@@ -159,14 +222,30 @@ defmodule Scholar.Metrics.Distance do
...
+
+      iex> x = Nx.tensor([[1,2], [3,4]])
+      iex> y = Nx.tensor([[8,3], [2,5]])
+      iex> Scholar.Metrics.Distance.chebyshev(x, y, axes: [1])
+      #Nx.Tensor<
+        f32[2]
+        [7.0, 1.0]
+      >
   """
   @spec chebyshev(Nx.t(), Nx.t()) :: Nx.t()
```
polvalente, line 234:
```suggestion
  @spec chebyshev(Nx.t(), Nx.t(), keyword()) :: Nx.t()
```

lib/scholar/metrics/distance.ex
```diff
@@ -118,15 +162,29 @@ defmodule Scholar.Metrics.Distance do
...
+
+      iex> x = Nx.tensor([[1,2], [3,4]])
+      iex> y = Nx.tensor([[8,3], [2,5]])
+      iex> Scholar.Metrics.Distance.manhattan(x, y, axes: [0])
+      #Nx.Tensor<
+        f32[2]
+        [8.0, 2.0]
+      >
   """
   @spec manhattan(Nx.t(), Nx.t()) :: Nx.t()
```
polvalente, line 174:
```suggestion
  @spec manhattan(Nx.t(), Nx.t(), keyword()) :: Nx.t()
```

lib/scholar/metrics/distance.ex
```diff
@@ -77,15 +102,29 @@ defmodule Scholar.Metrics.Distance do
...
+
+      iex> x = Nx.tensor([[1,2], [3,4]])
+      iex> y = Nx.tensor([[8,3], [2,5]])
+      iex> Scholar.Metrics.Distance.squared_euclidean(x, y, axes: [0])
+      #Nx.Tensor<
+        f32[2]
+        [50.0, 2.0]
+      >
   """
   @spec squared_euclidean(Nx.t(), Nx.t()) :: Nx.t()
```
polvalente, line 114:
```suggestion
  @spec squared_euclidean(Nx.t(), Nx.t(), keyword()) :: Nx.t()
```

lib/scholar/metrics/distance.ex
```diff
@@ -35,16 +40,31 @@ defmodule Scholar.Metrics.Distance do
...
+
+      iex> x = Nx.tensor([[1,2], [3,4]])
+      iex> y = Nx.tensor([[8,3], [2,5]])
+      iex> Scholar.Metrics.Distance.euclidean(x, y, axes: [0])
+      #Nx.Tensor<
+        f32[2]
+        [7.071067810058594, 1.4142135381698608]
+      >
   """
   @spec euclidean(Nx.t(), Nx.t()) :: Nx.t()
```
polvalente, line 52:
```suggestion
  @spec euclidean(Nx.t(), Nx.t(), keyword()) :: Nx.t()
```

lib/scholar/metrics/distance.ex
```diff
@@ -198,17 +284,54 @@ defmodule Scholar.Metrics.Distance do
...
+        |> Nx.subtract(y)
+        |> Nx.abs()
+        |> Nx.power(p)
+        |> Nx.sum(axes: opts[:axes])
+        |> Nx.power(1.0 / p)
+    end
+  end
+
+  deftransformp check_p(p) do
+    if p < 0 do
```
polvalente, line 331: I don't think this check is working as intended given that `p` is passed as a tensor to the check

lib/scholar/metrics/distance.ex
```diff
@@ -240,25 +368,53 @@ defmodule Scholar.Metrics.Distance do
...
+      #Nx.Tensor<
+        f32[3][3]
+        [
+          [0.1704850196838379, 0.2418246865272522, 1.0],
+          [1.0, 1.0, 0.0],
+          [0.3740193247795105, 0.2843400239944458, 1.0]
+        ]
+      >
   """
   @spec cosine(Nx.t(), Nx.t()) :: Nx.t()
```
polvalente, line 384:
```suggestion
  @spec cosine(Nx.t(), Nx.t(), keyword()) :: Nx.t()
```

lib/scholar/metrics/distance.ex
```diff
@@ -240,25 +368,53 @@ defmodule Scholar.Metrics.Distance do
...
-  defn cosine(x, y) do
+  defn cosine(x, y, opts \\ []) do
+    cutoff = 10 * 2.220446049250313e-16
```
polvalente, line 386: What does this cutoff mean? Perhaps a comment could be useful here

lib/scholar/metrics/distance.ex
```diff
@@ -240,25 +368,53 @@ defmodule Scholar.Metrics.Distance do
...
+    norm_x = Nx.select(Nx.greater(norm_x, cutoff), norm_x, 1.0)
+    norm_y = Nx.select(Nx.greater(norm_y, cutoff), norm_y, 1.0)
+
+    norm_x = x / norm_x
+    norm_y = y / norm_y
+
+    res = Nx.dot(norm_x, Nx.transpose(norm_y))
+    res = Nx.select(zero_xor_one_mask, 0.0, res)
+    res = 1.0 - Nx.select(zero_mask, 1.0, res)
+    if m != 1, do: res, else: res[0][0]
```
polvalente, line 417: Why `res[0][0]`? What happens if the input has rank > 2

lib/scholar/metrics/distance.ex
```diff
@@ -240,25 +368,53 @@ defmodule Scholar.Metrics.Distance do
...
+      iex> y = Nx.tensor([[1, 5, 2], [2, 4, 1], [0, 0, 0]])
+      iex> Scholar.Metrics.Distance.cosine(x, y, axes: [1])
+      #Nx.Tensor<
+        f32[3][3]
+        [
+          [0.1704850196838379, 0.2418246865272522, 1.0],
+          [1.0, 1.0, 0.0],
+          [0.3740193247795105, 0.2843400239944458, 1.0]
+        ]
+      >
```
jonatanklosko, line 382: In other cases we just batch multiple distance computations, but in this case we calculate the distance between each pair of sub-tensors, is there a good reason for this?

lib/scholar/metrics/distance.ex
jonatanklosko: @polvalente I think on latest Nx this shouldn't matter because if works at compile time if the condition value is a constant. That said, I don't think there's a reason to explicitly convert to a tensor.

lib/scholar/metrics/distance.ex
```diff
@@ -198,17 +284,54 @@ defmodule Scholar.Metrics.Distance do
...
+        |> Nx.power(p)
+        |> Nx.sum(axes: opts[:axes])
+        |> Nx.power(1.0 / p)
+    end
+  end
+
+  deftransformp check_p(p) do
+    if p < 0 do
+      raise ArgumentError,
+            "The value of p must be non-negative"
```
jonatanklosko, line 333:
```suggestion
      raise ArgumentError, "expected the value of :p to be a non-negative integer, got: #{inspect(p)}"
```

lib/scholar/metrics/distance.ex
```diff
@@ -198,17 +284,54 @@ defmodule Scholar.Metrics.Distance do
...
+        x
+        |> Nx.subtract(y)
+        |> Nx.abs()
+        |> Nx.power(p)
+        |> Nx.sum(axes: opts[:axes])
+        |> Nx.power(1.0 / p)
+    end
+  end
+
+  deftransformp check_p(p) do
```
jonatanklosko, line 330: For simple checks I think it's fine to just inline them in the actual function :)

lib/scholar/metrics/distance.ex
```diff
@@ -13,6 +13,11 @@ defmodule Scholar.Metrics.Distance do
   D(x, y) = \\sqrt{\\sum_i (x_i - y_i)^2}
   $$
 
+  ## Options
+
+    * `:axes` - Axes to aggregate distance over. If `:axes` set to `nil` then function does not aggregate distances.
+      Defaults to `nil`.
```
jonatanklosko, line 19: Perhaps something like this?

```suggestion
    * `:axes` - Axes to calculate the distance over. By default the distance
      is calculated between the whole tensors.
```

lib/scholar/metrics/distance.ex
```diff
@@ -35,16 +40,31 @@ defmodule Scholar.Metrics.Distance do
...
   @spec euclidean(Nx.t(), Nx.t()) :: Nx.t()
-  defn euclidean(x, y) do
+  defn euclidean(x, y, opts \\ []) do
     assert_same_shape!(x, y)
+
+    opts =
+      keyword!(
+        opts,
+        axes: nil
+      )
```
jonatanklosko, line 60:
```suggestion
    opts = keyword!(opts, [:axes])
```

lib/scholar/metrics/distance.ex
msluszniak: Parameter `:p` might be a float, so I adapt this but with a non-negative number.

lib/scholar/metrics/distance.ex
msluszniak: It prevents me from dividing by close to zero values in the normalization process. Similar to [link](https://github.com/scikit-learn/scikit-learn/blob/36958fb240fbe435673a9e3c52e769f01f36bec0/sklearn/preprocessing/_data.py#L88). I'll leave a comment.

lib/scholar/metrics/distance.ex
msluszniak: At this moment the algorithm deals only with 2d inputs because it calculate pair-wise distance, but I'll change that to be consistent with the rest of API

lib/scholar/metrics/distance.ex
msluszniak: Probably not, I will change that.

lib/scholar/metrics/distance.ex
```diff
@@ -240,24 +349,62 @@ defmodule Scholar.Metrics.Distance do
...
+      0.0
+    else
+      x_squared = x |> then(&Nx.multiply(&1, &1))
+      y_squared = y |> then(&Nx.multiply(&1, &1))
+
+      norm_x =
+        x_squared
+        |> Nx.sum(axes: opts[:axes], keep_axes: true)
+        |> Nx.sqrt()
+        |> Nx.broadcast(Nx.shape(x))
```
jonatanklosko, line 381:
```suggestion
        |> Nx.broadcast(x)
```

lib/scholar/metrics/distance.ex
```diff
@@ -240,24 +349,62 @@ defmodule Scholar.Metrics.Distance do
...
+
+      norm_y =
+        y_squared
+        |> Nx.sum(axes: opts[:axes], keep_axes: true)
+        |> Nx.sqrt()
+        |> Nx.broadcast(Nx.shape(y))
```
jonatanklosko, line 387:
```suggestion
        |> Nx.broadcast(y)
```

lib/scholar/metrics/distance.ex
```diff
@@ -240,24 +349,62 @@ defmodule Scholar.Metrics.Distance do
...
-      norm_x == 0.0 or norm_y == 0.0 ->
-        1.0
+    opts = keyword!(opts, [:axes])
 
-      true ->
-        numerator = Nx.dot(x, y)
-        denominator = norm_x * norm_y
-        1.0 - numerator / denominator
+    if Nx.all(x == y) do
+      0.0
```
jonatanklosko, line 372: Why do we need this special case? In particular note that this currently ignores `:axes`.

lib/scholar/metrics/distance.ex
```diff
@@ -240,24 +349,62 @@ defmodule Scholar.Metrics.Distance do
...
+        x_squared
+        |> Nx.sum(axes: opts[:axes], keep_axes: true)
+        |> Nx.sqrt()
+        |> Nx.broadcast(x)
+
+      norm_y =
+        y_squared
+        |> Nx.sum(axes: opts[:axes], keep_axes: true)
+        |> Nx.sqrt()
+        |> Nx.broadcast(y)
```
jonatanklosko, line 387: We shouldn't duplicate the computation of `norm_x` and `norm_x_mask`. I think we don't even need the explicit broadcast. So I would just compute `norm_x` and keep axis, then squeeze `opts[:axes]` for calculating the masks.

lib/scholar/metrics/distance.ex
```diff
@@ -240,24 +349,62 @@ defmodule Scholar.Metrics.Distance do
...
+        |> Nx.sum(axes: opts[:axes])
+        |> Nx.sqrt()
+
+      norm_y_mask =
+        y_squared
+        |> Nx.sum(axes: opts[:axes])
+        |> Nx.sqrt()
+
+      zero_mask = norm_x_mask == 0.0 and norm_y_mask == 0.0
+      zero_xor_one_mask = Nx.logical_xor(norm_x_mask == 0.0, norm_y_mask == 0.0)
```
jonatanklosko, line 400: Maybe a more descriptive names would make it more clear, like `both_zero?` and `one_zero?`.

lib/scholar/metrics/distance.ex
```diff
@@ -218,47 +317,73 @@ defmodule Scholar.Metrics.Distance do
...
+    norm_y_mask = norm_y |> Nx.squeeze()
+
+    both_zero? = norm_x_mask == 0.0 and norm_y_mask == 0.0
+    one_zero? = Nx.logical_xor(norm_x_mask == 0.0, norm_y_mask == 0.0)
+
+    norm_x = Nx.select(norm_x > cutoff, norm_x, 1.0)
+    normalized_x = x / norm_x
+
+    norm_y = Nx.select(norm_y > cutoff, norm_y, 1.0)
+    normalized_y = y / norm_y
```
jonatanklosko, line 382: We can move this before we compute masks, so then we can just do `norm_x = Nx.squeeze(norm_x, axes: opts[:axes])` and use this.

lib/scholar/metrics/distance.ex
```diff
@@ -218,47 +312,72 @@ defmodule Scholar.Metrics.Distance do
...
+
+    norm_x = Nx.squeeze(norm_x, axes: opts[:axes])
+    norm_y = Nx.squeeze(norm_y, axes: opts[:axes])
+
+    both_zero? = norm_x == 0.0 and norm_y == 0.0
+    one_zero? = Nx.logical_xor(norm_x == 0.0, norm_y == 0.0)
```
jonatanklosko, line 376: Oh one last thing, we should extract `x_zero? = norm_x == 0.0`, so that we can reuse it in both of these expressions!

jonatanklosko (approved):
Looks great, just one last comment :)

@polvalente feel free to merge whenever you're happy!

jonatanklosko:
I'm gonna merge since further work depends on this :)

PR #29, opened 2022-09-13
Refactor metrics

msluszniak:


lib/scholar/metrics.ex
```diff
@@ -380,44 +367,11 @@ defmodule Scholar.Metrics do
...
-    assert_rank(y_true, 1)
-    assert_same_shape(y_pred, y_true)
-
-    num_classes =
-      transform(opts[:num_classes], fn num_classes ->
-        num_classes || raise ArgumentError, "missing option :num_classes"
-      end)
-
-    transform(opts[:average], fn average ->
-      if Elixir.Kernel.==(average, :micro) do
```
jonatanklosko, line 392: Actually, now we can use `case` directly in defn, so we can use it here :)

msluszniak: Oh, ok

jonatanklosko (approved):

PR #30, opened 2022-09-14
Handle zero division in metrics

### LINKED ISSUE #16: Handle zero division in metrics
Currently the formulas add `1.0e-16` to avoid zero division, but this unnecessarily impacts valid results. Instead, we should check for zero division and choose a reasonable behaviour.

msluszniak:
Closes #16.

### DIFF
--- a/lib/scholar/metrics.ex
+++ b/lib/scholar/metrics.ex
@@ -48,7 +48,8 @@ defmodule Scholar.Metrics do
   Computes the precision of the given predictions with respect to
   the given targets for binary classification problems.
 
-  If the sum of true positive and false positive examples equals to 0, then the result is set to 0.
+  If the sum of true positives and false positives is 0, then the
+  result is 0 to avoid zero division.
 
   ## Examples
 
@@ -73,7 +74,8 @@ defmodule Scholar.Metrics do
   Computes the precision of the given predictions with respect to
   the given targets for multi-class classification problems.
 
-  If the sum of true positive and false positive examples equals to 0, then the result is set to 0.
+  If the sum of true positives and false positives is 0, then the
+  result is 0 to avoid zero division.
 
   ## Options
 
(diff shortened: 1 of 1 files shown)


msluszniak:
I haven't added an option for values other than 0 since @jonatanklosko convinced me to wait for a use case.

lib/scholar/metrics.ex
```diff
@@ -118,13 +124,15 @@ defmodule Scholar.Metrics do
...
 
-    true_positives / (false_negatives + true_positives + 1.0e-16)
+    safe_division(true_positives, false_negatives + true_positives)
   end
 
   @doc ~S"""
   Computes the recall of the given predictions with respect to
   the given targets for multi-class classification problems.
 
+  If the sum of true positive and false positive examples equals to 0, then the result is set to 0.
```
jonatanklosko, line 134:
```suggestion
  If the sum of true positives and false negatives is 0, then the
  result is 0 to avoid zero division.
```

msluszniak: Ok, I'll change the other docs accordingly.

lib/scholar/metrics.ex
```diff
@@ -327,6 +345,13 @@ defmodule Scholar.Metrics do
   Calculates F1 score given rank-1 tensors which represent
   the expected (`y_true`) and predicted (`y_pred`) classes.
 
+  If the sum of true positives and false positives is 0, then the
+  precision is 0 to avoid zero division.
+  If the sum of true positives and false negatives is 0, then the
+  recall is 0 to avoid zero division.
+  If the sum of precision and recall is 0, then the
+  F1 score is 0 to avoid zero division.
```
jonatanklosko, line 353:
```suggestion
  If all examples are true negatives, then the result is 0 to
  avoid zero division.
```

In all other case we return the expected value, because with 0 true positives F1 should be 0.

jonatanklosko (approved):

PR #34, opened 2022-09-15
Nimble options

msluszniak:


lib/scholar/cluster/k_means.ex
```diff
@@ -47,20 +47,23 @@ defmodule Scholar.Cluster.KMeans do
...
+    num_clusters: [required: true, type: :pos_integer],
+    max_iterations: [
+      type: :pos_integer,
+      default: 300
+    ],
+    num_runs: [type: :pos_integer, default: 10],
+    tol: [type: :float, default: 1.0e-4],
+    weights: [type: {:or, [:atom, {:list, {:or, [:float, :non_neg_integer]}}]}],
+    init: [type: :atom, default: :k_means_plus_plus]
+  ]
```
josevalim, line 60:
```suggestion
  @schema NimbleOptions.new!(
    num_clusters: [required: true, type: :pos_integer],
    max_iterations: [
      type: :pos_integer,
      default: 300
    ],
    num_runs: [type: :pos_integer, default: 10],
    tol: [type: :float, default: 1.0e-4],
    weights: [type: {:or, [:atom, {:list, {:or, [:float, :non_neg_integer]}}]}],
    init: [type: :atom, default: :k_means_plus_plus]
  )
```

josevalim:
@msluszniak looks good, we need to do two other changes:

1. We should use `NimbleOptions.docs/1` to generate the docs of the options. Please make sure it looks similar to what we have today. :)

2. We should remove the validation we have in the transformer action and do it with NimbleOptions. Let me know if something is missing!

msluszniak:
There are a few problems that I've met:
1.  Every time I've tried to break a doc for k_means_plus_plus, something does not work
2.  For the `:init` option, information about the default value is placed with the wrong indentation
3.  There is no type for positive_float, and I cannot check it correctly
4.  Some options might be validated only with input (Nx tensor), and we cannot use NimbleOptions

josevalim:
For 3 and 4, you can add custom types to NimbleOptions by passing a module as the type. Please check the docs :) I will take a look at the other ones. :)

lib/scholar/cluster/k_means.ex
```diff
@@ -8,31 +8,70 @@ defmodule Scholar.Cluster.KMeans do
...
+                   all observations are assigned equal weight.
+                   """
+                 ],
+                 init: [
+                   type: {:in, [:k_means_plus_plus, :random]},
+                   default: :k_means_plus_plus,
+                   doc: """
+                   Method for centroid initialization, either of:
+
+                   #{NimbleOptions.docs(@initialization_schema, nest_level: 1)}
```
josevalim, line 66: Instead of cascading NimbleOptions, you should just inline those docs here:

```suggestion
                   * `:k_means_plus_plus` - does this
                   * `:random` - does that
```
```

File: lib/scholar/cluster/k_means.ex [hunk 3]
```diff
@@ -8,31 +8,70 @@ defmodule Scholar.Cluster.KMeans do
...
+                   type: :float,
+                   default: 1.0e-4,
+                   doc: """
+                    Relative tolerance with regards to Frobenius norm of the difference in
+                    the cluster centers of two consecutive iterations to declare convergence.
+                   """
+                 ],
+                 weights: [
+                   default: nil,
+                   type: {:or, [:atom, {:list, {:or, [:float, :non_neg_integer]}}]},
```
josevalim, line 54: I am wondering if it is best to not specify default of `nil`? That's the implicit default value:

```suggestion
                   type: {:list, {:or, [:float, :non_neg_integer]}},
```

lib/scholar/cluster/k_means.ex
```diff
@@ -8,31 +8,70 @@ defmodule Scholar.Cluster.KMeans do
...
+                   type: :pos_integer,
+                   default: 300,
+                   doc: "Maximum number of iterations of the k-means algorithm for a single run."
+                 ],
+                 num_runs: [
+                   type: :pos_integer,
+                   default: 10,
+                   doc: """
+                    Number of time the k-means algorithm will be run with different centroid seeds.
+                     The final results will be the best output of num_runs runs in terms of inertia.
```
josevalim, line 41: In Elixir multiline strings, we should never indent in relation to the closing `"""`: 

```suggestion
                   Number of time the k-means algorithm will be run with different centroid seeds.
                   The final results will be the best output of num_runs runs in terms of inertia.
```

:)

lib/scholar/cluster/k_means.ex
```diff
@@ -8,31 +8,70 @@ defmodule Scholar.Cluster.KMeans do
...
+                   """
+                 ],
+                 tol: [
+                   type: :float,
+                   default: 1.0e-4,
+                   doc: """
+                    Relative tolerance with regards to Frobenius norm of the difference in
+                    the cluster centers of two consecutive iterations to declare convergence.
```
josevalim, line 49:
```suggestion
                   Relative tolerance with regards to Frobenius norm of the difference in
                   the cluster centers of two consecutive iterations to declare convergence.
```

lib/scholar/cluster/k_means.ex
msluszniak: When I try this one, the `:k_means_plus_plus` and `:random` are indented wrongly.

josevalim: I will fix it in nimble options.  :)

lib/scholar/cluster/k_means.ex
```diff
@@ -8,31 +8,66 @@ defmodule Scholar.Cluster.KMeans do
...
+                   type: {:in, [:k_means_plus_plus, :random]},
+                   default: :k_means_plus_plus,
+                   doc: """
+                   Method for centroid initialization, either of:
+
+                   * `:k_means_plus_plus` - selects initial cluster centroids using sampling based
+                   on an empirical probability distribution of the points’ contribution to
+                   the overall inertia. This technique speeds up convergence, and is
+                   theoretically proven to be O(log(k))-optimal.
+                   * `:random` - choose `:num_clusters` observations (rows) at random from data for the initial centroids.
```
josevalim, line 62:
```suggestion
                   * `:k_means_plus_plus` - selects initial cluster centroids using sampling based
                   on an empirical probability distribution of the points’ contribution to
                   the overall inertia. This technique speeds up convergence, and is
                   theoretically proven to be O(log(k))-optimal.

                   * `:random` - choose `:num_clusters` observations (rows) at random from data for the initial centroids.
```

lib/scholar/cluster/k_means.ex
```diff
@@ -8,31 +8,66 @@ defmodule Scholar.Cluster.KMeans do
...
-      are assigned equal weight.
-
-    * `:init` - Method for centroid initialization, either of:
-
-        * `:k_means_plus_plus` (default) -  selects initial cluster centroids using sampling based on
-          an empirical probability distribution of the points’ contribution to the overall inertia.
-         This technique speeds up convergence, and is theoretically proven to be O(log(k))-optimal.
-
-        * `:random` - choose `:num_clusters` observations (rows) at random from data for the initial centroids.
+  ## Options \n#{NimbleOptions.docs(@opts_schema)}
```
josevalim, line 70:
```suggestion
  ## Options

  #{NimbleOptions.docs(@opts_schema)}
```

josevalim:

PR #35, opened 2022-09-16
Add NimbleOptions to the other functions

msluszniak:


lib/scholar/metrics.ex
```diff
@@ -14,6 +14,43 @@ defmodule Scholar.Metrics do
...
+      * `:macro`. Calculate metrics for each label, and find their unweighted mean.
+      This does not take label imbalance into account.
+
+      * `:weighted`. Calculate metrics for each label, and find their average weighted by
+      support (the number of true instances for each label).
+
+      * `:micro`. Calculate metrics globally by counting the total true positives,
+      false negatives and false positives.
+
+      * `nil`. The f1 scores for each class are returned.
```
josevalim, line 46:
```suggestion
      * `:macro` - Calculate metrics for each label, and find their unweighted mean.
      This does not take label imbalance into account.

      * `:weighted` - Calculate metrics for each label, and find their average weighted by
      support (the number of true instances for each label).

      * `:micro` - Calculate metrics globally by counting the total true positives,
      false negatives and false positives.

      * `nil` - The f1 scores for each class are returned.
```

josevalim:

PR #37, opened 2022-09-21
Add PCA and minor fixes

msluszniak:


lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,216 @@
...
+  @doc """
+  Fits a PCA for sample inputs `x`.
+
+  ## Options
+
+  #{NimbleOptions.docs(@fit_opts_schema)}
+
+  ## Returns
+
+    The function returns a struct with the following parameters:
```
josevalim, line 81:
```suggestion
  The function returns a struct with the following parameters:
```

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,216 @@
...
+      type:
+        {:or, [{:custom, Scholar.Options, :positive_number, []}, {:in, [:none, :pos_integer]}]},
+      default: :none,
+      doc: ~S"""
+      Number of components to keep. if `:num_components` is not set all components are kept:
+
+      $num\\_components = min(num\\_samples, num\\_features)$
+
+      If $0 < num\\_components <= 1$, select the number of components such
+      that the amount of variance that needs to be explained is greater than the percentage specified by `:num_components`.
```
josevalim, line 42: What does the `:pos_integer` value does?

josevalim (approved):
LGTM, just one tiny comment!

lib/scholar/decomposition/pca.ex
msluszniak: If `:num_components` is an integer then it simply indicates the number of components to keep.

msluszniak:
There is one problem. After I added the $ ... $ notation in `mix.exs` one of the docs in `similarity.ex` does not look as it is supposed to @josevalim.

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,216 @@
...
+    explained_variance = singular_values * singular_values / (num_samples - 1)
+    explained_variance_ratio = explained_variance / Nx.sum(explained_variance)
+
+    num_components =
+      calculate_num_components(
+        num_components,
+        num_features,
+        num_samples,
+        explained_variance
+      )
```
jonatanklosko, line 133: I think it's confusing if the returned values include all components, but `num_components` is smaller. Even more so, that after `fit` the user needs to do `PCA.transform(num_components: Nx.to_number(model.num_components))`.

I think we can either:

1. Only accept `:num_components` as an integer and slice all the values accordingly (and revisit once/if we have dynamic shapes?)
2. Have a separate function that takes the desired explained variance and reduces the number of components accordingly. This would need to be non-defn because the shape is dynamic

Opinions? cc @polvalente @seanmor5

lib/scholar/metrics/distance.ex
```diff
@@ -194,16 +194,16 @@ defmodule Scholar.Metrics.Distance do
   """
   @spec manhattan(Nx.t(), Nx.t(), keyword()) :: Nx.t()
   deftransform manhattan(x, y, opts \\ []) do
-    nmanhattan(x, y, NimbleOptions.validate!(opts, @general_schema))
+    manhattan_n(x, y, NimbleOptions.validate!(opts, @general_schema))
   end
 
-  defnp nmanhattan(x, y, opts \\ []) do
+  defnp manhattan_n(x, y, opts \\ []) do
```
jonatanklosko, line 200: For regular elixir function it's common to use the naming `def manhattan` and `defp do_manhattan` when "wrapping" a function into a more friendly API.

Note sure if we want to push a different convention for `defn`, but I think `_n` can be more surprising for an external reader.

If anyone has a strong opinion, please speak :)

josevalim: I started the `_n` exactly to signal that we are calling the actual numeric implementation. So I am in favor :D

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,216 @@
...
+  @doc """
+  For a fitted model performs a decomposition.
+
+  ## Options
+
+  #{NimbleOptions.docs(@transform_opts_schema)}
+
+  ## Returns
+
+    The function returns a decomposed data.
```
josevalim, line 157:
```suggestion
  The function returns a decomposed data.
```

lib/scholar/decomposition/pca.ex
josevalim: Should we then change transform to get the num_components from the model instead?

jonatanklosko: @josevalim we would need to call `Nx.to_number` inside that transform, so it would no longer make sense as a transform.

josevalim: So we need to compute `num_components` as a `defn`?

msluszniak: If we compute num_components as defn, then num_components will be a tensor, and the problem still occurs.

jonatanklosko: @josevalim the thing is that in sklearn `num_components` can be either a static number or a 0.0-1.0 float, in which case they compute `num_components` dynamically to preserve the given amount of variance. So we need `num_components` to be static, but it is dynamic in the float case.

josevalim: I pushed a commit that makes `num_components` a dynamic value on `fit`. The idea was to force dynamic values so we force thinking about the dynamic nature of the problem.

For `transform`, my suggestion is to rename `transform` to `mask` and convert all non component values to zero. By masking the decomposed values, we leave the dim reduction to the caller who will hopefully have more knowledge about the shapes. Would that makes sense?

josevalim: Btw, if you want to keep my commit, then please make sure to add tests that both integer and float values for num_components are clipped.

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,220 @@
...
+
+  ## Options
+
+  #{NimbleOptions.docs(@transform_opts_schema)}
+
+  ## Returns
+
+  The function returns a decomposed data.
+  """
+  deftransform transform(model, opts \\ []) do
```
jonatanklosko, line 198: Shouldn't we accept arbitrary dataset `x` and express it in terms of the trained principal components?

lib/scholar/decomposition/pca.ex
jonatanklosko: > By masking the decomposed values, we leave the dim reduction to the caller who will hopefully have more knowledge about the shapes. Would that makes sense?

I'm not sure which approach is best here. If a user has a dataset with 100 features and wants to reduce it to 10 features with PCA, then by masking we don't really reduce the dimensionality, so they may end up with unnecessarily heavy computation later on. They could slice the transformed dataset themselves, but it's kinda weird to say `num_components: 10` just to manually slice 10 afterwards.

I can't say much about the typical use case, but I would assume that the dynamic case is more useful during manual exploration, rather than a part of a bigger algorithm. So I would stick to static `num_components` and for exploration they can do this:

```elixir
model = Scholar.Decomposition.PCA.fit(data)
Nx.cumulative_sum(model.explained_variance)
# See which point is best to drop components
model = Scholar.Decomposition.PCA.fit(data, num_components: 4)
```

We can also add a separate function for slicing the model if somebody requests it.

> For transform, my suggestion is to rename transform to mask and convert all non component values to zero.

I'd stick with `transform`, because even if we mask some features, the other ones change :)

josevalim: So we could do this:

* if `num_components` is integer or :none, we store it in the model as a static value
* if `float`, we don't store `num_components`
* On transform, you can pass your own or we fallback to `num_components`. If `num_components` is nil, we raise

Thoughts?

Alternatively we don't support `float` `num_components`at all and revert my changes.

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,216 @@
...
+
+  defnp transform_n(
+          %__MODULE__{
+            singular_values: singular_values,
+            num_samples: num_samples,
+            decomposer: decomposer
+          } = _model,
+          opts \\ []
+        ) do
+    opts = keyword!(opts, [:num_components, whiten: false])
```
josevalim, line 171: We don't need this `keyword!` due to NimbleOptions.

lib/scholar/decomposition/pca.ex
jonatanklosko: > if `float`, we don't store `num_components`

In that case I don't think we end up using the float at all.

I would go with static `num_components` (dropping support for floats) :)

lib/scholar/decomposition/pca.ex
msluszniak: Yes, I'll correct that.

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,216 @@
...
+             :components,
+             :explained_variance,
+             :explained_variance_ratio,
+             :singular_values,
+             :mean,
+             :num_components,
+             :num_features,
+             :num_samples,
+             :decomposer
+           ]}
```
josevalim, line 18: If components is not meant to be a tensor, you'll need this:

```suggestion
           keep: [:num_components],
           containers: [
             :components,
             :explained_variance,
             :explained_variance_ratio,
             :singular_values,
             :mean,
             :num_features,
             :num_samples,
             :decomposer
           ]}
```

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,200 @@
...
+    num_components: [
+      type: {:or, [:pos_integer, {:in, [:none]}]},
+      default: :none,
+      doc: ~S"""
+      Number of components to keep. if `:num_components` is not set all components are kept:
+
+      $num\\_components = min(num\\_samples, num\\_features)$
+
+      If $0 < num\\_components <= 1$, select the number of components such
+      that the amount of variance that needs to be explained is greater than the percentage specified by `:num_components`.
```
jonatanklosko, line 41:
```suggestion
      Number of components to keep. If `:num_components` is not set, all components are kept:

      $num\\_components = min(num\\_samples, num\\_features)$
```

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,200 @@
...
+    :num_components,
+    :num_features,
+    :num_samples,
+    :decomposer
+  ]
+
+  fit_opts_schema = [
+    num_components: [
+      type: {:or, [:pos_integer, {:in, [:none]}]},
+      default: :none,
```
jonatanklosko, line 34: Let's just use `nil` instead :)

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,200 @@
...
+  ]
+
+  transform_opts_schema = [
+    num_components: [
+      required: true,
+      type: :pos_integer,
+      doc: """
+      Number of components to keep. It should equal to the value calculated in the `fit/2` method.
+      """
+    ],
```
jonatanklosko, line 53: We no longer need this, as we can infer it from the model, no?

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,215 @@
...
+        ) do
+    whiten? = opts[:whiten]
+
+    x = x - mean
+
+    x_transformed =
+      Nx.dot(
+        x,
+        Nx.transpose(components)
+      )
```
jonatanklosko, line 169:
```suggestion
    x_transformed =
      Nx.dot(x, [1], components, [1])
```

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,215 @@
...
+
+    x_transformed =
+      if not whiten? do
+        x_transformed
+      else
+        x_transformed / Nx.sqrt(explained_variance)
+      end
```
jonatanklosko, line 176:
```suggestion
      if whiten? do
        x_transformed / Nx.sqrt(explained_variance)
      else
        x_transformed
      end
```

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,215 @@
...
+
+    x_transformed[[0..-1//1, 0..(num_components - 1)]]
```
jonatanklosko, line 178: We should slice the model in fit, such that we only have `:num_components` components (and the other fields accordingly). This way we don't need to unnecessarily compute and slice on transform.

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,215 @@
...
+  end
+
+  deftransformp calculate_num_components(
+                  num_components,
+                  num_features,
+                  num_samples
+                ) do
+    cond do
+      num_components == nil ->
+        min(num_features, num_samples)
```
jonatanklosko, line 197: You can assign it to a variable like `default_num_components, since we use it in multiple places here.

msluszniak:
I'll implement a fit_transform functionality which will be calculated differently than fit and then transform.

josevalim (approved):

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,252 @@
...
+    %__MODULE__{
+      components: components,
+      explained_variance: explained_variance,
+      explained_variance_ratio: explained_variance_ratio,
+      singular_values: singular_values,
+      mean: mean,
+      num_components: num_components,
+      num_features: num_features,
+      num_samples: num_samples
+    }
```
jonatanklosko, line 135: I think we should slice all the values to only include `:num_component`. That's what sklearn does as well.

lib/scholar/decomposition/pca.ex
```diff
@@ -0,0 +1,252 @@
...
+    * `:mean` - Per-feature empirical mean, estimated from the training set.
+
+    * `:num_components` - It equals the parameter `:num_components`, or the lesser
+      value of `:num_features` and `:num_samples` if the parameter `:num_components` is `nil`.
+
+    * `:num_features` - Number of features in the training data.
+
+    * `:num_samples` - Number of samples in the training data.
+  """
+  deftransform fit(x, opts \\ []) do
```
jonatanklosko, line 96: Did we settle on `train`?

msluszniak: For this particular one, the `fit` fits better than the `train` in my opinion. However, if we want to stick to one name then I'm ok with the `train`.

jonatanklosko: I agree, though I think unified naming is best for the end user. I would also be fine with `fit` everywhere. @josevalim I believe you advocated for `train`, thoughts on this case?

jonatanklosko: To me the semantic difference is that `train` implies a more iterative process, while `fit` is more like a fixed calculation, so this differs on algorithm basis. However, for the end user the result is the same.

josevalim: I am fine either way, as long as we keep one. Perhaps fit is more general, so we should stabilize on that?

jonatanklosko: Sounds good to me!

josevalim: So let's merge this as `fit` whenever ready and then rename the rest to `fit` in a separate PR?

jonatanklosko (approved):

PR #40, opened 2022-09-23
feat: add cubic spline interpolation

polvalente:


lib/scholar/interpolation/cubic_spline.ex
```diff
@@ -0,0 +1,147 @@
...
+
+    x_poly = target_x - Nx.take_along_axis(x, idx)
+
+    result =
+      x_poly
+      |> Nx.power(Nx.tensor([[3], [2], [1], [0]], type: :f64))
+      |> Nx.multiply(c_poly)
+      |> Nx.sum(axes: [0])
+
+    Nx.select(nan_selector, nan, result)
```
jonatanklosko, line 145: We should do

```elixir
if opts[:extrapolate] do
  result
else
  nan_selector = ...
  Nx.select(...)
end
```

This way when `extrapolate: true` we will optimize-out that branch entirely (especialy that it's the default :)).

lib/scholar/interpolation/cubic_spline.ex
```diff
@@ -0,0 +1,147 @@
...
+      doc: "if false, out-of-bounds x values raise."
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(opts)
+
+  @doc """
+  Fits a cubic spline interpolation of the given `(x, y)` points
+  """
+  defn train(x, y) do
```
jonatanklosko, line 24: We should document that x must be in increasing order.

lib/scholar/interpolation/cubic_spline.ex
```diff
@@ -0,0 +1,147 @@
...
+
+  defnp predict_n(%__MODULE__{x: x, coefficients: coefficients}, target_x, opts \\ []) do
+    nan_selector =
+      Nx.logical_and(opts[:extrapolate] == false, target_x < x[0] or target_x > x[-1])
+
+    nan = Nx.tensor(:nan, type: :f64)
+    target_x = Nx.as_type(target_x, :f64)
+
+    idx =
+      Nx.sum(x < Nx.new_axis(target_x, 1), axes: [1])
```
jonatanklosko, line 132: It may be just me, but I think broadcasting tensor rank is confusing and adds overhead when reading the code, so I favor reshaping both operands.

```suggestion
      (Nx.new_axis(target_x, 1) > Nx.new_axis(x, 0))
      |> Nx.sum(axes: [1])
```

Sidenote: instead of `Nx.sum` we could use `Nx.argmax`, but I don't see a clear reason why either should be more efficient :)

lib/scholar/interpolation/cubic_spline.ex
```diff
@@ -0,0 +1,147 @@
...
+  end
+
+  defnp predict_n(%__MODULE__{x: x, coefficients: coefficients}, target_x, opts \\ []) do
+    nan_selector =
+      Nx.logical_and(opts[:extrapolate] == false, target_x < x[0] or target_x > x[-1])
+
+    nan = Nx.tensor(:nan, type: :f64)
+    target_x = Nx.as_type(target_x, :f64)
+
+    idx =
```
jonatanklosko, line 131: `spline_idx` perhaps (or however we should call the indexed thing)?

lib/scholar/interpolation/cubic_spline.ex
```diff
@@ -0,0 +1,147 @@
...
+      Nx.sum(x < Nx.new_axis(target_x, 1), axes: [1])
+      |> Nx.min(Nx.size(x) - 2)
+
+    c_poly = Nx.take_along_axis(coefficients, Nx.tile(idx, [4, 1]), axis: 1)
```
jonatanklosko, line 135: Pretty sure that's

```suggestion
    c_poly = Nx.take(coefficients, idx, axis: 1)
```

lib/scholar/interpolation/cubic_spline.ex
```diff
@@ -0,0 +1,147 @@
...
+
+    x_poly = target_x - Nx.take_along_axis(x, idx)
+
+    result =
+      x_poly
```
jonatanklosko, line 140: For clarity

```suggestion
      x_poly
      |> Nx.new_axis(1)
```

lib/scholar/interpolation/cubic_spline.ex
```diff
@@ -0,0 +1,147 @@
...
+      |> Nx.power(Nx.tensor([[3], [2], [1], [0]], type: :f64))
+      |> Nx.multiply(c_poly)
+      |> Nx.sum(axes: [0])
```
jonatanklosko, line 143:
```suggestion
      |> Nx.dot([0], c_poly, [0])
```

lib/scholar/interpolation/cubic_spline.ex
polvalente: We can apply sorting with argsort to deal with this, I think

jonatanklosko: Sounds good! In most cases `train` can be expected to be somewhat expensive, so an additional sort doesn't hurt and will prevent from bugs.

lib/scholar/interpolation/cubic_spline.ex
polvalente: this worked :)

lib/scholar/interpolation/cubic_spline.ex
polvalente: I couldn't get this to yield the correct shape/values :(

polvalente: nvm, I managed to find the correct values. We need to interpret the coeff-x pairs as batches :)

lib/scholar/interpolation/cubic_spline.ex
polvalente: argmax ended up being the fix I needed for an off-by-one

lib/scholar/interpolation/cubic_spline.ex
polvalente: I used *_poly as the common suffix for the related variables

lib/scholar/interpolation/cubic_spline.ex
```diff
@@ -0,0 +1,164 @@
...
+    result =
+      x_poly
+      |> Nx.new_axis(1)
+      |> Nx.power(Nx.tensor([3, 2, 1, 0]))
+      |> Nx.dot([1], [0], coef_poly, [1], [0])
+
+    if opts[:extrapolate] do
+      result
+    else
+      Nx.select(nan_selector, nan, result)
```
jonatanklosko, line 161: I would move the `nan_selector = ...` here, since we don't use it anywhere else.

lib/scholar/interpolation/cubic_spline.ex
jonatanklosko: Ah yeah, that makes sense!

lib/scholar/interpolation/cubic_spline.ex
```diff
@@ -0,0 +1,164 @@
...
+
+  @doc """
+  Fits a cubic spline interpolation of the given `(x, y)` points
+  """
+  defn train(x, y) do
+    # https://en.wikiversity.org/wiki/Cubic_Spline_Interpolation
+    # Reference implementation in Scipy
+
+    x = Nx.as_type(x, :f64)
+    y = Nx.as_type(y, :f64)
```
josevalim, line 29: What do we force such type conversions? If the user wants f64 precision, then they must pass f64 in?

jonatanklosko (commented):
@polvalente there are a couple comments left, but other than that everything looks good to me!

lib/scholar/interpolation/cubic_spline.ex
```diff
@@ -0,0 +1,189 @@
...
+        :ok
+
+      {y_shape, x_shape} ->
+        raise ArgumentError,
+              "expected y to have shape #{inspect(x_shape)}, got: #{inspect(y_shape)}"
+    end
+
+    sort_idx = Nx.argsort(x)
+    x = Nx.take_along_axis(x, sort_idx)
+    y = Nx.take_along_axis(y, sort_idx)
```
jonatanklosko, line 53:
```suggestion
    x = Nx.take(x, sort_idx)
    y = Nx.take(y, sort_idx)
```

For 1D tensor it should be the same and I think take is less involved as it always uses the indices as is (at least in exla).

lib/scholar/interpolation/cubic_spline.ex
```diff
@@ -0,0 +1,189 @@
...
+  and at least 3 entries.
+  """
+  defn train(x, y) do
+    # https://en.wikiversity.org/wiki/Cubic_Spline_Interpolation
+    # Reference implementation in Scipy
+
+    {n, dx} =
+      case Nx.shape(x) do
+        {n} when n > 2 ->
+          dx = x[1..-1//1] - x[0..-2//1]
```
jonatanklosko, line 34: This doesn't need to be conditional, no?

lib/scholar/interpolation/cubic_spline.ex
```diff
@@ -0,0 +1,189 @@
+defmodule Scholar.Interpolation.CubicSpline do
+  @moduledoc """
+  Cubic spline interpolation
```
jonatanklosko, line 3:
```suggestion
  Cubic spline interpolation.
```

jonatanklosko (approved):
Final comments then ship it!

lib/scholar/interpolation/cubic_spline.ex
polvalente: What do you mean? the guard?
It's because I haven't dealt with the case where only 2 points are given, so the code would also break

lib/scholar/interpolation/cubic_spline.ex
polvalente: For torchx it's the opposite haha! They provide a gather function which is exactly our take_along_axis

lib/scholar/interpolation/cubic_spline.ex
jonatanklosko: I mean that we can compute `dx` separately, since it's not relevant for the check itself :)

lib/scholar/interpolation/cubic_spline.ex
jonatanklosko: Oh well :see_no_evil: Good to know!

PR #52, opened 2022-12-06
Add Gaussian Naive Bayes

msluszniak:


lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,425 @@
+defmodule Scholar.NaiveBayes.GaussianNB do
```
josevalim, line 1: Do we need NB in the name given that we are already in the NaiveBayes namespace?

```suggestion
defmodule Scholar.NaiveBayes.Gaussian do
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,425 @@
...
+      Prior probabilities of the classes. If specified, the priors are not
+      adjusted according to the data. We assume that priors are correct and
+      sum(priors) == 1.
+      """
+    ],
+    sample_weights: [
+      type: {:list, {:or, [:float, :integer]}},
+      doc: ~S"""
+      Array-like of shape (n_samples,), if not provided
+      then set to ones of shape (n_samples,).
```
josevalim, line 32:
```suggestion
      List of `n_samples` elements.
      
      A list of 1.0 values is used if none is given.
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,425 @@
...
+    end
+
+    opts =
+      opts ++
+        [
+          sample_weights_flag: if(opts[:sample_weights] != nil, do: true, else: false),
+          priors_flag: if(opts[:priors] != nil, do: true, else: false)
+        ]
+
+    fit_n(x, y, opts)
```
josevalim, line 150: We want to build the `sample_weights` as tensor here and pass an argument. Otherwise we will emit a different `defn` cache for each value of `sample_weights`. :)

josevalim: Same for `priors`!

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
polvalente: Don't we have an imputer namespace, given the moduledoc

polvalente: Perhaps the moduledoc is leftover from copy-paste

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,425 @@
...
+      iex> Scholar.NaiveBayes.GaussianNB.predict_log_proba(model, Nx.tensor([[6, 2, 4], [8, 5, 9]]))
+      #Nx.Tensor<
+        f32[2][3]
+        [
+          [-1822222336.0, -1822222208.0, 0.0],
+          [-399999968.0, -5733332992.0, 0.0]
+        ]
+      >
+  """
+  defn predict_log_proba(%__MODULE__{} = model, x) do
```
jonatanklosko, line 190: Usually we prefer explicit names, so let's rename to `_probabilities` :)

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,422 @@
...
+
+          if Nx.size(priors) != Nx.size(classes) do
+            raise ArgumentError,
+                  "Number of priors must match number of classes. Number of priors: #{inspect(Nx.size(priors))} does not match number of classes: #{inspect(Nx.size(classes))}"
+          else
+            priors
+          end
+      end
+
+    fit_n(x, y, sample_weights, class_priors, classes, opts)
```
josevalim, line 177: Sorry, this is a little bit annoying to handle, but you want to call as few tensor operations as possible outside of `defn`. Usually, you want to leave only tensor allocations outside:

```
    # Remove the value from opts so it is not part of the cache
    {sample_weights, opts} = Keyword.pop(opts, :sample_weights, 1.0)
    sample_weights = Nx.tensor(sample_weights)

    {priors, opts} = Keyword.pop(opts, :priors, 0.0)
    priors = Nx.tensor(priors)

    fit_n(x, y, sample_weights, class_priors, opts)
```

And move all broadcasting, iota, etc in:

```
    num_classes = opts[:num_classes]
    classes = Nx.iota({num_classes})

    priors =
      case Nx.shape(priors) do
        {} -> Nx.broadcast(priors, {num_classes})
        {^num_classes} -> priors
        _ -> raise ArgumentError, "..."
      end

    {num_samples, _} = Nx.shape(x)

    sample_weights =
      case Nx.shape(sample_weights) do
        {} -> Nx.broadcast(sample_weights, {num_samples})
        {^num_samples} -> sample_weights
        _ -> raise ArgumentError, "..."
      end
```

Does it make sense?

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,422 @@
...
+  end
+
+  defnp update_mean_variance_init(mu, var, x, sample_weights, mask, sample_weights_flag) do
+    {num_samples, _num_features} = Nx.shape(x)
+
+    if num_samples == 0 do
+      {mu, var, 0}
+    else
+      case sample_weights_flag do
+        false ->
```
josevalim, line 333: Maybe worth using `cond`?

cond do
  num_samples == 0 ->
    ...
  sample_weights_flag == false ->
    ...
  true ->
    ...
end
```

File: lib/scholar/naive_bayes/gaussian_naive_bayes.ex (hunk 5 above)
Comment (msluszniak) [reply lib/scholar/naive_bayes/gaussian_naive_bayes.ex:177]: Ok, sure

File: lib/scholar/naive_bayes/gaussian_naive_bayes.ex [hunk 7]
```diff
@@ -0,0 +1,420 @@
...
+      raise ArgumentError,
+            "Wrong input shape. Expect x to have the same first dimension as y, got: #{inspect(num_samples)} for x and #{inspect(num_targets)} for y"
+    end
+
+    opts =
+      opts ++
+        [
+          sample_weights_flag: if(opts[:sample_weights] != nil, do: true, else: false),
+          priors_flag: if(opts[:priors] != nil, do: true, else: false)
+        ]
```
josevalim, line 149:
```suggestion
    opts =
      [
        sample_weights_flag: if(opts[:sample_weights] != nil, do: true, else: false),
        priors_flag: if(opts[:priors] != nil, do: true, else: false)
      ] ++
        opts
```

It is cheaper to prepend than append in linked lists. Now that it will matter here... but good to follow best practices. :)

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,420 @@
...
+      raise ArgumentError,
+            "Wrong input rank. Expected: 2 for x and 1 for y, got: #{inspect(input_rank)} for x and #{inspect(targets_rank)} for y"
+    end
+
+    {num_samples, _} = Nx.shape(x)
+    {num_targets} = Nx.shape(y)
+
+    if num_samples != num_targets do
+      raise ArgumentError,
+            "Wrong input shape. Expect x to have the same first dimension as y, got: #{inspect(num_samples)} for x and #{inspect(num_targets)} for y"
```
josevalim, line 141: No need for inspect for numbers and small grammar fix:

```suggestion
            "wrong input shape. Expected x to have the same first dimension as y, got: #{num_samples} for x and #{num_targets} for y"
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,420 @@
...
+  """
+  deftransform fit(x, y, opts \\ []) do
+    opts = NimbleOptions.validate!(opts, @opts_schema)
+
+    input_rank = Nx.rank(x)
+    targets_rank = Nx.rank(y)
+
+    if input_rank != 2 or targets_rank != 1 do
+      raise ArgumentError,
+            "Wrong input rank. Expected: 2 for x and 1 for y, got: #{inspect(input_rank)} for x and #{inspect(targets_rank)} for y"
```
josevalim, line 133:
```suggestion
            "wrong input rank. Expected x to be rank 2 and y to be rank 1, got: #{input_rank} for x and #{targets_rank} for y"
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,420 @@
...
+      case Nx.shape(class_priors) do
+        {} ->
+          Nx.broadcast(class_priors, {num_classes})
+
+        {^num_classes} ->
+          class_priors
+
+        _ ->
+          raise ArgumentError,
+                "Number of priors must match number of classes. Number of priors: #{inspect(Nx.size(class_priors))} does not match number of classes: #{inspect(Nx.size(classes))}"
```
josevalim, line 252:
```suggestion
                "number of priors must match number of classes. Number of priors: #{Nx.size(class_priors)} does not match number of classes: #{Nx.size(classes)}"
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,420 @@
...
+      case Nx.shape(sample_weights) do
+        {} ->
+          Nx.broadcast(sample_weights, {num_samples})
+
+        {^num_samples} ->
+          sample_weights
+
+        _ ->
+          raise ArgumentError,
+                "Number of weights must match number of samples. Number of weights: #{inspect(Nx.size(sample_weights))} does not match number of samples: #{inspect(num_samples)}"
```
josevalim, line 265:
```suggestion
                "number of weights must match number of samples. Number of weights: #{Nx.size(sample_weights)} does not match number of samples: #{num_samples}"
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,420 @@
...
+      Nx.sum(broadcast_mask * broadcast_weights, axes: [0])
+  end
+
+  defnp check_input(%__MODULE__{theta: theta}, x) do
+    {_, num_features} = Nx.shape(theta)
+    {_, x_num_features} = Nx.shape(x)
+
+    if num_features != x_num_features do
+      raise ArgumentError,
+            "Wrong input shape. Expect x to have the same second dimension as the data for fitting process, got: #{inspect(x_num_features)} for x and #{inspect(num_features)} for training data"
```
josevalim, line 417:
```suggestion
            "wrong input shape. Expected x to have the same second dimension as the data for fitting process, got: #{x_num_features} for x and #{num_features} for training data"
```

test/scholar/naive_bayes/gaussian_naive_bayes.exs
```diff
@@ -0,0 +1,267 @@
+defmodule Scholar.NaiveBayes.GaussianTest do
```
josevalim, line 1: You are missing the _test.exs extension on this file, so these tests are not running. :)

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
+defmodule Scholar.NaiveBayes.Gaussian do
```
polvalente, line 1: Given that Scholar also houses non-classifier stuff like interpolation algorithms, how about using `Scholar.Classifiers.NaiveBayes` as a namespace?

There might be other things that could move into this `Classifiers` namespace, though, in a separate PR

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
...
+  @moduledoc """
+  Gaussian Naive Bayes.
```
polvalente, line 3: Perhaps that Stanford reference could be added as a link in the extended doc, as well as a short explanation on what this does

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
...
+
+    if num_samples != num_targets do
+      raise ArgumentError,
+            "wrong input shape. Expected x to have the same first dimension as y, got: #{num_samples} for x and #{num_targets} for y"
+    end
+
+    opts =
+      [
+        sample_weights_flag: if(opts[:sample_weights] != nil, do: true, else: false),
+        priors_flag: if(opts[:priors] != nil, do: true, else: false)
```
polvalente, line 147:
```suggestion
        priors_flag: opts[:priors] != nil)
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
...
+    {num_targets} = Nx.shape(y)
+
+    if num_samples != num_targets do
+      raise ArgumentError,
+            "wrong input shape. Expected x to have the same first dimension as y, got: #{num_samples} for x and #{num_targets} for y"
+    end
+
+    opts =
+      [
+        sample_weights_flag: if(opts[:sample_weights] != nil, do: true, else: false),
```
polvalente, line 146:
```suggestion
        sample_weights_flag: opts[:sample_weights] != nil,
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
...
+        epsilon: Nx.tensor(1.1250000042650754e-8)
+      }
+  """
+  deftransform fit(x, y, opts \\ []) do
+    opts = NimbleOptions.validate!(opts, @opts_schema)
+
+    input_rank = Nx.rank(x)
+    targets_rank = Nx.rank(y)
+
+    if input_rank != 2 or targets_rank != 1 do
```
polvalente, line 131: I'd break this into 2 separate rank validations because they aren't interdepent

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
...
+  defn predict_log_probability(%__MODULE__{} = model, x) do
+    check_input(model, x)
+    jll = joint_log_likelihood(model, x)
+
+    log_proba_x =
+      jll
+      |> Nx.exp()
+      |> Nx.sum(axes: [1])
+      |> Nx.log()
+      |> Nx.reshape({:auto, 1})
```
polvalente, line 228:
```suggestion
      |> Nx.new_axis(1)
```
assuming this is already a flat tensor

polvalente: Same result and efficiency, I just think it's clearer on what's happening

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
...
+    total_var = total_ssd / total_n
+
+    {total_mu, total_var}
+  end
+
+  defnp joint_log_likelihood(
+          %__MODULE__{class_priors: class_priors, var: var, theta: theta},
+          x
+        ) do
+    pi = 3.14159265359
```
polvalente, line 384:
```suggestion
    @pi :math.pi()
```

I prefer extracting this to an attribute instead

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
...
+    {num_classes, num_features} = Nx.shape(theta)
+    {samples_x, _features_x} = Nx.shape(x)
+
+    n1 =
+      (-0.5 * Nx.sum(Nx.log(2.0 * pi * var), axes: [1]))
+      |> Nx.reshape({:auto, 1})
+      |> Nx.broadcast({num_classes, samples_x})
+
+    x_broadcast =
+      Nx.reshape(x, {1, samples_x, num_features})
```
polvalente, line 395: `num_features` and `_features_x` are the same, right?
```suggestion
      Nx.new_axis(x, 0)
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
...
+      |> Nx.broadcast({num_classes, samples_x, num_features})
+
+    theta_broadcast =
+      Nx.reshape(theta, {num_classes, 1, num_features})
```
polvalente, line 399:
```suggestion
      Nx.new_axis(theta, 1)
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
...
+      |> Nx.broadcast({num_classes, samples_x, num_features})
+
+    var_broadcast =
+      Nx.reshape(var, {num_classes, 1, num_features})
```
polvalente, line 403:
```suggestion
      Nx.new_axis(var, 1)
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
...
+      |> Nx.broadcast({num_classes, samples_x, num_features})
+
+    n2 = -0.5 * Nx.sum((x_broadcast - theta_broadcast) ** 2 / var_broadcast, axes: [2])
+
+    (n1 + n2) |> Nx.transpose() |> Nx.add(joint)
```
polvalente, line 408:
```suggestion
    Nx.transpose(n1 + n2) + joint
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
...
+  end
+
+  defnp mean_masked(t, mask) do
+    {num_samples, num_features} = Nx.shape(t)
+
+    broadcast_mask =
+      mask |> Nx.reshape({num_samples, 1}) |> Nx.broadcast({num_samples, num_features})
```
polvalente, line 415:
```suggestion
      mask |> Nx.new_axis(1) |> Nx.broadcast({num_samples, num_features})
```

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
```diff
@@ -0,0 +1,442 @@
...
+
+    broadcast_weights =
+      weights |> Nx.reshape({num_samples, 1}) |> Nx.broadcast({num_samples, num_features})
+
+    Nx.sum(t * broadcast_mask * broadcast_weights, axes: [0]) /
+      Nx.sum(broadcast_mask * broadcast_weights, axes: [0])
+  end
+
+  defnp check_input(%__MODULE__{theta: theta}, x) do
```
polvalente, line 433: given this is basically a transform, perhaps deftransformp would be more fitting (no pun intended)
```suggestion
  deftransformp check_input(%__MODULE__{theta: theta}, x) do
```

josevalim: If it works as defnp, i think we should stick with it, because it guarantees certain properties.

lib/scholar/naive_bayes/gaussian_naive_bayes.ex
josevalim: I believe our goal is to mirror Scipy as much as possible, no? Or do you mean we could have both `Scholar.Classifiers.NaiveBayes.Gaussian` and `Scholar.SomethingElse.NaiveBayes.Gaussian`?

polvalente: I didn't find Naive Bayes in SciPy, but there is sklearn.naive_bayes. There, they also suffix the submodules like sklearn.naive_bayes.GaussianNB.

Not sure if I like it, but it's your call

josevalim: @polvalente it is up to us (you?) to decide because honestly I am not fully qualified. Can you look at the sklearn.naive_bayes namespace and say if those are all classifiers or if each of them belong to a different category and therefore effectively deserve their own namespace?

polvalente: All of the algorithms in sklearn.naive_bayes are classification algorithms based on the naïve Bayes concept/methodology (i.e. assumption that features are all independent random variables).

I'd put them in Scholar.Classification.NaiveBayes and also move Scholar.Linear.LogisticRegression to Scholar.Classification.Linear.LogisticRegression.

LogisticRegression by itself isn't necessarily a classification algorithm, but as it is implemented here, it is a classifier based on logistic regression. Same for NaiveBayes, they are classifiers based on the naïve interpretation of Bayes' theorem.

josevalim: Hrm... Honestly, I would like to avoid four levels of nesting. It is not very common in Elixir, especially in public APIs. 

Also, is it possible that we would augment LogisticRegression APIs in the future so it does more than classification?

josevalim:

PR #55, opened 2022-12-17
Modify Scholar to work with newer packages

msluszniak:


mix.exs
```diff
@@ -21,9 +21,9 @@ defmodule Scholar.MixProject do
   defp deps do
     [
       {:ex_doc, "~> 0.27", only: :dev, runtime: false},
-      {:nx, github: "elixir-nx/nx", sparse: "nx"},
-      {:explorer, "~> 0.3.0", only: [:test, :dev]},
-      {:nimble_options, "~> 0.5"}
+      {:nx, "~> 0.4.1", override: true},
```
polvalente, line 24: We shouldn't use `override` here. Why is jt necessary?

msluszniak: artefact

josevalim:

PR #56, opened 2022-12-20
Add Linear Regression

msluszniak:
Currently, we need [#1032](https://github.com/elixir-nx/nx/pull/1032) and [#1027](https://github.com/elixir-nx/nx/issues/1027) to be implemented before merging this PR.


lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,123 @@
...
+          Nx.broadcast(sample_weights, {n})
+
+        _ ->
+          sample_weights
+      end
+
+    scale = Nx.sqrt(sample_weights)
+    scale = Nx.broadcast(0.0, {n, n}) |> Nx.put_diagonal(scale)
+
+    {Nx.dot(scale, x), Nx.dot(scale, y)}
```
polvalente, line 77:
```suggestion
    scale = Nx.new_axis(scale, 1)
    {Nx.multiply(scale, x), Nx.multiply(scale, y)}
```

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,123 @@
...
+    y = if Nx.rank(y) == 1, do: Nx.new_axis(y, -1), else: y
+
+    {u_rows, _u_cols} = Nx.shape(u)
+    {vt_rows, _vt_cols} = Nx.shape(vt)
+    min = Kernel.min(u_rows, vt_rows)
+    vt = vt[[0..(min - 1), 0..-1//1]]
+    u = u[[0..-1//1, 0..(min - 1)]]
+    mask = s > rcond * Nx.reduce_max(s)
+    safe_s = Nx.select(mask, s, 1)
+    print_value(safe_s)
```
polvalente, line 98:
```suggestion
```

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,123 @@
...
+    y_rank = Nx.rank(y)
+    y = if Nx.rank(y) == 1, do: Nx.new_axis(y, -1), else: y
+
+    {u_rows, _u_cols} = Nx.shape(u)
+    {vt_rows, _vt_cols} = Nx.shape(vt)
+    min = Kernel.min(u_rows, vt_rows)
+    vt = vt[[0..(min - 1), 0..-1//1]]
+    u = u[[0..-1//1, 0..(min - 1)]]
+    mask = s > rcond * Nx.reduce_max(s)
+    safe_s = Nx.select(mask, s, 1)
```
polvalente, line 97: I don't think you need to build `safe_s` because `s` is always guaranteed to be floating, so 1/0 becomes nan, which is then replaced on `s_inv = ...`

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,123 @@
...
+      end
+
+    scale = Nx.sqrt(sample_weights)
+    scale = Nx.broadcast(0.0, {n, n}) |> Nx.put_diagonal(scale)
+
+    {Nx.dot(scale, x), Nx.dot(scale, y)}
+  end
+
+  # Implements ordinary least-squares by estimating the
+  # solution A to the equation A.X = b.
```
polvalente, line 81:
```suggestion
  # solution A to the equation A.X = y.
```

lib/scholar/linear/linear_regression.ex
msluszniak: Valid point

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,123 @@
...
+    vt = vt[[0..(min - 1), 0..-1//1]]
+    u = u[[0..-1//1, 0..(min - 1)]]
+    mask = s > rcond * Nx.reduce_max(s)
+    safe_s = Nx.select(mask, s, 1)
+    print_value(safe_s)
+
+    s_inv = Nx.new_axis(Nx.select(mask, 1 / safe_s, 0), -1)
+
+    utb = Nx.dot(Nx.transpose(u), y)
+    coeffs = Nx.dot(Nx.transpose(vt), s_inv * utb)
```
polvalente, line 103:
```suggestion
    utb = Nx.dot(Nx.LinAlg.adjoint(u), y)
    coeffs = Nx.dot(Nx.LinAlg.adjoint(vt), s_inv * utb)
```

This way you'll already support complex numbers when SVD supports them

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,123 @@
...
+        else: Nx.mean(x, axes: [0])
+
+    x = Nx.subtract(x, x_offset)
+
+    y_offset =
+      if opts[:sample_weights_flag],
+        do: Nx.weighted_mean(y, sample_weights, axis: 0),
+        else: Nx.mean(y, axes: [0])
+
+    y = Nx.subtract(y, y_offset)
```
polvalente, line 120:
```suggestion
    y = y - y_offset
```

lib/scholar/linear/linear_regression.ex
polvalente: btw, theres elixir-nx/nx#995 which would simplify a lot of this code into: `x = Nx.LinAlg.pinv(a) |> Nx.dot(y)`

test/scholar/linear/linear_regression_test.exs
```diff
@@ -0,0 +1,567 @@
...
+          -0.02042902633547783,
+          0.2715783417224884,
+          0.16200493276119232
+        ])
+
+      %Scholar.Linear.LinearRegression{coefficients: actual} =
+        Scholar.Linear.LinearRegression.fit(a, b)
+
+      assert Nx.all_close(expected, actual, rtol: 1.0e-2, atol: 1.0e-2) ==
+               Nx.tensor(1, type: {:u, 8})
```
polvalente, line 67: I'd bring the `assert_all_close` helper from Nx into Scholar. Helps a lot with these assertions

polvalente:
Why is elixir-nx/nx#1027 needed for this PR?

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,123 @@
...
+    if y_rank == 1, do: Nx.flatten(coeffs), else: coeffs
+  end
+
+  defnp preprocess_data(x, y, sample_weights, opts \\ []) do
+    x_offset =
+      if opts[:sample_weights_flag],
+        do: Nx.weighted_mean(x, sample_weights, axis: 0),
+        else: Nx.mean(x, axes: [0])
+
+    x = Nx.subtract(x, x_offset)
```
msluszniak:
```suggestion
    x = x - x_offset
```

lib/scholar/linear/linear_regression.ex
polvalente: Right. Let's just simplify the original then:
```suggestion
    scale = Nx.make_diagonal(scale)

    {Nx.dot(scale, x), Nx.dot(scale, y)}
```

msluszniak:
> Why is [elixir-nx/nx#1027](https://github.com/elixir-nx/nx/issues/1027) needed for this PR?

Well, because SVD is based on eigh and now results of SVD are still slightly off. I guess that eigh might be the cause of the problem. Especially that

for this matrix
```
[[ 0.05477347, -0.33786303,  0.10690729, -0.02246768],
       [-0.33786303,  0.29224722,  0.26443713,  0.33520083],
       [ 0.10690729,  0.26443713,  0.07571281,  0.02051978],
       [-0.02246768,  0.33520083,  0.02051978,  0.09305623]]
```
we got in Nx
```
{#Nx.Tensor<
   f32[4]
   [0.7275585532188416, -0.40110692381858826, 0.17198389768600464, 0.017355527728796005]
 >,
 #Nx.Tensor<
   f32[4][4]
   [
     [0.35916027426719666, -0.5357038974761963, -0.6361281275749207, 0.42369478940963745],
     [-0.7836878299713135, -0.6060684323310852, -0.02153218351304531, -0.1344778835773468],
     [-0.2672005295753479, 0.4525536000728607, -0.7711756825447083, -0.35918542742729187],
     [-0.43062877655029297, 0.3753633499145508, -0.012862256728112698, 0.8205999135971069]
   ]
 >}
```
vs in jax
```
(DeviceArray([-0.40120745,  0.01702507,  0.17235827,  0.7276138 ], dtype=float32),
 DeviceArray([[ 0.5389725 ,  0.39385098, -0.6501477 ,  0.36290246],
              [ 0.6116504 , -0.1352644 , -0.00994076, -0.77941555],
              [-0.44395748, -0.39439   , -0.7578279 , -0.2702879 ],
              [-0.37187913,  0.8191714 , -0.05391173, -0.43331048]],            dtype=float32))
```

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,122 @@
...
+    min = Kernel.min(u_rows, vt_rows)
+    vt = vt[[0..(min - 1), 0..-1//1]]
+    u = u[[0..-1//1, 0..(min - 1)]]
+    mask = s > rcond * Nx.reduce_max(s)
+    safe_s = Nx.select(mask, s, 1)
+
+    s_inv = Nx.new_axis(Nx.select(mask, 1 / safe_s, 0), -1)
+
+    utb = Nx.dot(Nx.LinAlg.adjoint(u), y)
+    coeffs = Nx.dot(Nx.LinAlg.adjoint(vt), s_inv * utb)
```
polvalente, line 102: This can be replaced by something like:

```elixir
pinv = Nx.LinAlg.pinv(x)
Nx.dot(pinv, y)
```

Also might be a good idea to rename `x` and `y` to `a` and `b` for the `fit` function and this one

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,104 @@
...
+  end
+
+  # Implements ordinary least-squares by estimating the
+  # solution A to the equation A.X = y.
+  defnp lstsq(x, y) do
+    y_rank = Nx.rank(y)
+    y = if Nx.rank(y) == 1, do: Nx.new_axis(y, -1), else: y
+    pinv = Nx.LinAlg.pinv(x)
+    coeff = Nx.dot(pinv, y)
+    if y_rank == 1, do: Nx.flatten(coeff), else: coeff
```
polvalente, line 85: Is this rank manipulation with the new axis still needed?

polvalente (approved):

lib/scholar/linear/linear_regression.ex
msluszniak: No, it's not. I'll repair this one.

lib/scholar/linear/linear_regression.ex
msluszniak: @polvalente only in `fit`, `fit_n`, and `lstsq`?

polvalente: I guess so. Mostly because we usually refer to A.X = B instead of X.?? = Y

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,94 @@
...
+    opts = NimbleOptions.validate!(opts, @opts_schema)
+
+    opts =
+      [
+        sample_weights_flag: opts[:sample_weights] != nil
+      ] ++
+        opts
+
+    {sample_weights, opts} = Keyword.pop(opts, :sample_weights, 1.0)
+    sample_weights = Nx.tensor(sample_weights)
```
polvalente, line 32: `sample_weights` should be an optional argument instead.

#58 has a similar need (and solution!) for dealing with optional opts + weights

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,94 @@
...
+    scale = Nx.sqrt(sample_weights) |> Nx.make_diagonal()
+
+    {Nx.dot(scale, x), Nx.dot(scale, y)}
+  end
+
+  # Implements ordinary least-squares by estimating the
+  # solution A to the equation A.X = b.
+  defnp lstsq(a, b) do
+    pinv = Nx.LinAlg.pinv(a)
+    Nx.dot(pinv, b)
```
polvalente, line 81: I think you can avoid the transpose on L46 if you do:
```suggestion
    Nx.dot(b, [0], pinv, [1])
```

This doesn't work for a batched version (you'd need to pass batch axes as well), but I don't think you're supporting batches here anyway.

```elixir
t1 = Nx.random_uniform({5, 4})

t2 = Nx.random_uniform({4, 6})

t3 = Nx.dot(t1, t2)
t4 = Nx.dot(t2, [-2], t1, [-1])

Nx.transpose(t3) == t4 # this returns true (using binary backend)
```

lib/scholar/linear/linear_regression.ex
```diff
@@ -0,0 +1,94 @@
...
+        {} ->
+          Nx.broadcast(sample_weights, {n})
+
+        _ ->
+          sample_weights
+      end
+
+    scale = Nx.sqrt(sample_weights) |> Nx.make_diagonal()
+
+    {Nx.dot(scale, x), Nx.dot(scale, y)}
```
polvalente, line 74: For the scalar case wouldn't this all result in the same as the scalar * matrix case?
```suggestion
    sample_weights =
      case Nx.shape(sample_weights) do
        {} = scalar ->
          scalar = Nx.sqrt(scalar)
          {scalar * x, scalar * y}
        _ -> 
          scale = sample_weights |> Nx.sqrt() |> Nx.make_diagonal()
          {Nx.dot(scale, x), Nx.dot(scale, y)}
      end
```

lib/scholar/linear/linear_regression.ex
josevalim: @polvalente we agreed on making them tensors later on. So this approach is fine as is by me.

But I am also wondering if we should just allow a `Nx.Tensor` as a direct argument to `Nx.tensor`. WDYT?

lib/scholar/linear/linear_regression.ex
josevalim: It is ok to not support batching, let's just make sure to raise on the shapes.

polvalente: Yes, I didn't mean that batching should have to be supported here

lib/scholar/linear/linear_regression.ex
polvalente: So it would be a no-op? The only problem I see is that it could be less obvious to people where they have tensors vs where they have numbers. 

Didn't we soft deprecate tensors-as-arguments to Nx.iota for a similar reason?

josevalim: We deprecated because we were passing tensors in places where shapes were allowed and in many planes shapes are options, which led to whole tensors being cached. In this case, it is tensor in, tensor out, so we should be fine.

polvalente: Alright then

josevalim (approved):

polvalente (approved):

PR #57, opened 2022-12-20
Add assert_all_close to tests

msluszniak:
I added `Nx.all_close` in tests to avoid changing expected values in the future.


test/scholar/decomposition/pca_test.exs
```diff
@@ -4,25 +4,7 @@ defmodule Scholar.Decomposition.PCATest do
   @x Nx.tensor([[-1, -1], [-2, -1], [-3, -2], [1, 1], [2, 1], [3, 2]])
   @x2 Nx.tensor([[1, 4], [54, 6], [26, 7]])
   @x3 Nx.tensor([[-1, -1, 3], [-2, -1, 2], [-3, -2, 1], [3, 1, 1], [21, 2, 1], [5, 3, 2]])
-
-  def assert_all_close(left, right, opts \\ []) do
```
polvalente, line 8: You should keep `assert_all_close` and also move it to a new `ExUnit.CaseTemplate` which is to be used anywhere it's needed.

test/scholar_test.exs
```diff
@@ -1,2 +1,22 @@
...
+
+    if equals != Nx.tensor(1, type: {:u, 8}, backend: Nx.BinaryBackend) do
+      flunk("""
+      expected
+      #{inspect(left)}
+      to be within tolerance of
+      #{inspect(right)}
+      """)
+    end
+  end
```
josevalim, line 21: Cool! Now instead of `use ExUnit.Case`, you can do `use ScholarTest` and use `assert_all_close` throughout :)

jonatanklosko (approved):

PR #58, opened 2022-12-24
Add Hamming Distance

acalejos:
Added Hamming Distance. Compared implementation to scipy implementation found at https://github.com/scipy/scipy/blob/v1.9.3/scipy/spatial/distance.py#L671-L722

The main difference is mine does allow for more than 1-D arrays. Only takes in a single axis parameter (rather than `axes` list in other distance metrics).  There is likely a better way to handle `nil` options with `NimbleOptions` so suggestions on that are welcome.


lib/scholar/metrics/distance.ex
```diff
@@ -407,4 +426,61 @@ defmodule Scholar.Metrics.Distance do
     res = Nx.select(one_zero?, 0.0, res)
     1.0 - Nx.select(both_zero?, 1.0, res)
   end
+
+  @doc """
+  Hamming distance.
+
+  N_unequal(x, y) / N_tot
```
polvalente, line 433:
```suggestion
  $hamming(x ,y) = \frac{#unequal(x, y)}{#x}$
```

lib/scholar/metrics/distance.ex
```diff
@@ -407,4 +426,61 @@ defmodule Scholar.Metrics.Distance do
...
+
+  ## Examples
+
+      iex> x = Nx.tensor([1, 0, 0])
+      iex> y = Nx.tensor([0, 1, 0])
+      iex> Scholar.Metrics.Distance.hamming(x, y)
+      #Nx.Tensor<
+        f32
+        0.6666666865348816
+      >
```
seanmor5, line 447:
```suggestion
      >

```

lib/scholar/metrics/distance.ex
```diff
@@ -407,4 +426,61 @@ defmodule Scholar.Metrics.Distance do
...
+    if nil == options[:weights] do
+      hamming_unweighted(x, y, options)
+    else
+      hamming_weighted(x, y, options)
+    end
+  end
+
+  defnp hamming_unweighted(x, y, opts \\ []) do
+    assert_same_shape!(x, y)
+    (x != y) |> Nx.mean(axes: [opts[:axis]])
```
polvalente, line 479: Why can't we support :axes instead of just a single axis?

lib/scholar/metrics/distance.ex
```diff
@@ -407,4 +426,61 @@ defmodule Scholar.Metrics.Distance do
...
+      iex> Scholar.Metrics.Distance.hamming(x, y, axis: 1)
+      #Nx.Tensor<
+        f32[3]
+        [0.6666666865348816, 1.0, 1.0]
+      >
+  """
+  deftransform hamming(x, y, opts \\ []) do
+    options = NimbleOptions.validate!(opts, @hamming_schema)
+
+    if nil == options[:weights] do
```
seanmor5, line 470:
```suggestion
    if options[:weights] do
```

nil is falsey so I think you can flip this clause and remove

lib/scholar/metrics/distance.ex
acalejos: This was mainly due to `weighted_mean` only taking `axis` as a param. This brings up additional questions as to whether we'd want to support multi-dim tensors (scipy elects to only support 1D). The current `moduledoc` for `Metrics.Distance` reads `Distance metrics between 1-D tensors.`

lib/scholar/metrics/distance.ex
acalejos: Agreed, I will change that.

lib/scholar/metrics/distance.ex
polvalente: Perhaps weighted_mean could be extended for having N-dimensional weights?

lib/scholar/metrics/distance.ex
acalejos: The denominator is really the length of the axis you're summing over. If we wanted to allow no axis, `Nx.mean` would then use the total number of items so again the denominator would be $m \times n$. I'm not sure of a succinct way of expressing this

polvalente: I've edited the suggestion to mention the aggregation axes.

I think for non-weighted we can support many axes and extend the weighted version when weighted_mean supports them.

polvalente: Also, `#` here symbolizes the size of the set

lib/scholar/metrics/distance.ex
acalejos: @seanmor5 I think this is more of a question for you

lib/scholar/metrics/distance.ex
acalejos: Ah, thanks I didn't realize that about `#`. I'll add in support for many axes in non-weighted

lib/scholar/metrics/distance.ex
polvalente: It was kind of a self-question as well haha!

lib/scholar/metrics/distance.ex
acalejos: @polvalente Do you think it makes sense to change the `axes` validator to allow for `int`s for a single axis? Then within the validator just wrap the int in a list and pass that to the function for use as `:axes` parameter? This seems more intuitive than requiring a list of one item (but then again, the term `axes` suggests an enumerable type)

polvalente: I'd rather always have lists, mainly because it's the way that the :axes parameter works in all of Nx so far

acalejos: makes sense, thanks

lib/scholar/metrics/distance.ex
acalejos: `Nx.mean` accepts `axes` so it seems like it would be consistent to have it for `weighted_mean` as well

polvalente: @acalejos https://github.com/elixir-nx/nx/pull/1043 adds support for `:axes` instead of `:axis` in `Nx.weighted_mean`

lib/scholar/metrics/distance.ex
```diff
@@ -407,4 +426,62 @@ defmodule Scholar.Metrics.Distance do
...
+  end
+
+  defnp hamming_unweighted(x, y, opts \\ []) do
+    assert_same_shape!(x, y)
+    (x != y) |> Nx.mean(axes: [opts[:axis]])
+  end
+
+  defnp hamming_weighted(x, y, opts \\ []) do
+    assert_same_shape!(x, y)
+    (x != y) |> Nx.weighted_mean(opts[:weights], axis: opts[:axis])
```
polvalente, line 485: I just realized we can't have `weights` as an option here.
This is because options are enclosed in most of our compilation infrastructure, so tensors must be passed as proper arguments to be kept as tensor expressions.

The code to deal with this is unfortunately a bit unwieldly, but it's basically like this:

```elixir
defmodule T do
import Nx.Defn

deftransform hamming(x, y), do: non_weighted(x, y)
deftransform hamming(x, y, opts) when is_list(opts) do
  # call nimble opts here :)
  non_weighted(x, y, opts)
end

deftransform hamming(x, y, w), do: weighted(x, y, w)
deftransform hamming(x, y, w, opts) when is_list(opts) do
  # call nimble opts here :)
  weighted(x, y, w, opts)
end

defnp weighted(x, y, w, opts \\ []) do
   ...
end

defnp non_weighted(x, y, opts \\ []) do
...
end
end
```

the guard on the second clause is what enables us to have both hamming/2 (non-weighted) with optional opts and hamming/3 (weighted) with optional opts

lib/scholar/options.ex
```diff
@@ -11,6 +11,19 @@ defmodule Scholar.Options do
     end
   end
 
+  def weights(weights) do
```
polvalente, line 14: given my latest review, we don't need this function anymore

lib/scholar/metrics/distance.ex
acalejos: Oh nice. Should this wait on that getting merged into Nx then?

polvalente: You can point to my branch while it's not merged so you can keep working on it :)

lib/scholar/metrics/distance.ex
acalejos: I think i see what youre saying. I think weights make more sense as a tensor so that's probably the way to go rather than having them be a list and stay in options.

lib/scholar/metrics/distance.ex
acalejos: What do you mean exactly by point to your branch? Would I do that from the mix.exs file?

polvalente: Yes!

You can change the nx declaration in mix.exs to:

`{:nx, ">= 0.0.0", github: "elixir-nx/nx", sparse: "nx", branch: "pv-feat/multi-axes-weighted-mean"}`

This is basically saying: fetch from the elixir-nx/nx repo, only the nx directory, and checkout the given branch

lib/scholar/metrics/distance.ex
acalejos: Made the changes at 7f4c606fdd861c60237af92d416d48e34de21016

Using the new `weighted_mean` with `axes` parameter some other Scholar tests failed. Seems to be only the PCA tests

polvalente: The failed tests are actually due to SVD. I'll update Scholar's main branch with Nx main

polvalente: Done :)
You can update your fork with Scholar main and it'll pull latest Nx as a dependency

acalejos: Looks like I had to run mix format. I pushed again after pulling from Scholar/main

lib/scholar/metrics/distance.ex
```diff
@@ -407,4 +407,68 @@ defmodule Scholar.Metrics.Distance do
...
+  ## Examples
+
+      iex> x = Nx.tensor([1, 0, 0])
+      iex> y = Nx.tensor([0, 1, 0])
+      iex> Scholar.Metrics.Distance.hamming(x, y)
+      #Nx.Tensor<
+        f32
+        0.6666666865348816
+      >
+      iex> weights = Nx.tensor([1,0.5,0.5])
```
polvalente, line 431:
```suggestion
      iex> weights = Nx.tensor([1, 0.5, 0.5])
```

lib/scholar/metrics/distance.ex
```diff
@@ -407,4 +407,68 @@ defmodule Scholar.Metrics.Distance do
...
+
+      iex> x = Nx.tensor([[1, 2, 3], [0, 0, 0], [5, 2, 4]])
+      iex> y = Nx.tensor([[1, 5, 2], [2, 4, 1], [0, 0, 0]])
+      iex> Scholar.Metrics.Distance.hamming(x, y, axes: [1])
+      #Nx.Tensor<
+        f32[3]
+        [0.6666666865348816, 1.0, 1.0]
+      >
+  """
+  deftransform(hamming(x, y), do: hamming_unweighted(x, y))
```
polvalente, line 451:
```suggestion
  deftransform hamming(x, y), do: hamming_unweighted(x, y)
```

lib/scholar/metrics/distance.ex
```diff
@@ -407,4 +407,68 @@ defmodule Scholar.Metrics.Distance do
...
+
+  deftransform hamming(x, y, opts) when is_list(opts) do
+    NimbleOptions.validate!(opts, @general_schema)
+    hamming_unweighted(x, y, opts)
+  end
+
+  deftransform(hamming(x, y, w), do: hamming_weighted(x, y, w))
```
polvalente, line 458:
```suggestion
  deftransform hamming(x, y, w), do: hamming_weighted(x, y, w)
```

polvalente (approved):

polvalente:
Don't worry about the formatter! I'll fix this in a few minutes :)

acalejos:
> ut
It failed a check:
```
warning: this clause for __defn:hamming__/2 cannot match because a previous clause at line 451 always matches
Warning:   lib/scholar/metrics/distance.ex:451
```

I thought that the guard would make these function signatures distinct?

polvalente:
> It failed a check:
> 
> ```
> warning: this clause for __defn:hamming__/2 cannot match because a previous clause at line 451 always matches
> Warning:   lib/scholar/metrics/distance.ex:451
> ```
> 
> I thought that the guard would make these function signatures distinct?

This is probably a bug on how `deftransform` is defined. I'll try to fix this as well upstream.

acalejos:
> This is probably a bug on how deftransform is defined. I'll try to fix this as well upstream.

I'll be interested to see how you fix it. Surprised that hasn't come up before

polvalente:
@acalejos this seems to fix it https://github.com/elixir-nx/nx/pull/1045

Let's hold off until the fix is merged

lib/scholar/metrics/distance.ex
```diff
@@ -407,4 +407,68 @@ defmodule Scholar.Metrics.Distance do
...
+  end
+
+  deftransform hamming(x, y, w), do: hamming_weighted(x, y, w)
+
+  deftransform hamming(x, y, w, opts) when is_list(opts) do
+    NimbleOptions.validate!(opts, @general_schema)
+    hamming_weighted(x, y, w, opts)
+  end
+
+  defnp hamming_unweighted(x, y, opts \\ []) do
```
josevalim, line 465: Declaring `opts` as default argument is no longer necessary as of latest Nx.

```suggestion
  defnp hamming_unweighted(x, y, opts) do
```

lib/scholar/metrics/distance.ex
```diff
@@ -407,4 +407,68 @@ defmodule Scholar.Metrics.Distance do
...
+    assert_same_shape!(x, y)
+    (x != y) |> Nx.mean(axes: opts[:axes])
+  end
+
+  defnp hamming_weighted(x, y, w, opts \\ []) do
```
josevalim, line 470:
```suggestion
  defnp hamming_weighted(x, y, w, opts) do
```

lib/scholar/metrics/distance.ex
```diff
@@ -407,4 +407,68 @@ defmodule Scholar.Metrics.Distance do
...
+
+      iex> x = Nx.tensor([[1, 2, 3], [0, 0, 0], [5, 2, 4]])
+      iex> y = Nx.tensor([[1, 5, 2], [2, 4, 1], [0, 0, 0]])
+      iex> Scholar.Metrics.Distance.hamming(x, y, axes: [1])
+      #Nx.Tensor<
+        f32[3]
+        [0.6666666865348816, 1.0, 1.0]
+      >
+  """
+  deftransform hamming(x, y), do: hamming_unweighted(x, y)
```
josevalim, line 451: This should also work, no?

```suggestion
  deftransform hamming(x, y, opts \\ [])
```

lib/scholar/metrics/distance.ex
```diff
@@ -407,4 +407,68 @@ defmodule Scholar.Metrics.Distance do
...
+
+  deftransform hamming(x, y, opts) when is_list(opts) do
+    NimbleOptions.validate!(opts, @general_schema)
+    hamming_unweighted(x, y, opts)
+  end
+
+  deftransform hamming(x, y, w), do: hamming_weighted(x, y, w)
+
+  deftransform hamming(x, y, w, opts) when is_list(opts) do
```
josevalim, line 460:
```suggestion
  deftransform hamming(x, y, w, opts \\ []) when is_list(opts) do
```

polvalente: Will this work though? Because this third argument can be either a list or a tensor depending on the clause

polvalente: Nevermind, I misread this as being the arity-3 version

lib/scholar/metrics/distance.ex
polvalente: The problem here is that the 3rd argument can either be a list or a tensor

josevalim:

PR #60, opened 2023-01-02
Add Intercept to Linear Regression

msluszniak:
No we can fit Linear Regression with intercept and without intercept.


lib/scholar/linear/linear_regression.ex
```diff
@@ -4,12 +4,24 @@ defmodule Scholar.Linear.LinearRegression do
...
   opts = [
     sample_weights: [
-      type: {:list, {:custom, Scholar.Options, :positive_number, []}}
+      type: {:list, {:custom, Scholar.Options, :positive_number, []}},
+      doc: """
+      The weights for each observation in x. If not provided,
+      all observations are assigned equal weight.
+      """
+    ],
+    fit_intercept?: [
```
polvalente, line 18: Perhaps a reference on what the intercept is would be nice

polvalente (approved):

lib/scholar/linear/linear_regression.ex
msluszniak: Maybe these two sentences would be enough?

polvalente (approved):

PR #64, opened 2023-01-04
Introduce random seed & use Nx.Random module instead of depricated functions

msluszniak:


lib/scholar/cluster/k_means.ex
```diff
@@ -70,6 +70,12 @@ defmodule Scholar.Cluster.KMeans do
       * `:random` - choose `:num_clusters` observations (rows) at random from data for
         the initial centroids.
       """
+    ],
+    random_seed: [
```
josevalim, line 74: Let's call it `seed`. And let's document it defaults to `System.system_time`.

lib/scholar/cluster/k_means.ex
```diff
@@ -253,27 +259,37 @@ defmodule Scholar.Cluster.KMeans do
...
+  end
+
+  defnp k_means_plus_plus(x, num_clusters, num_runs, opts) do
     inf = Nx.Constants.infinity()
     {num_samples, num_features} = Nx.shape(x)
     centroids = Nx.broadcast(inf, {num_runs, num_clusters, num_features})
     inertia = Nx.broadcast(0.0, {num_runs, num_samples})
 
-    first_centroid_idx = Nx.random_uniform({num_runs}, 0, num_samples - 1)
+    seed = check_seed(opts[:random_seed])
```
josevalim, line 278: I would compute the seed on `deftransform fit`. Otherwise, once this function is compiled once for a given shape, the seed will never change.

lib/scholar/cluster/k_means.ex
```diff
@@ -71,10 +71,11 @@ defmodule Scholar.Cluster.KMeans do
         the initial centroids.
       """
     ],
-    random_seed: [
+    seed: [
       type: :integer,
       doc: """
       Determines random number generation for centroid initialization.
+      If the seed is not provided, it is set to `System.system_time().
```
josevalim, line 78:
```suggestion
      If the seed is not provided, it is set to `System.system_time()`.
```

lib/scholar/cluster/k_means.ex
```diff
@@ -141,10 +142,16 @@ defmodule Scholar.Cluster.KMeans do
             "invalid value for :num_clusters option: expected positive integer between 1 and #{inspect(num_samples)}, got: #{inspect(opts[:num_clusters])}"
     end
 
-    fit_n(x, opts)
+    seed =
+      case opts[:seed] do
+        nil -> System.system_time()
+        _ -> opts[:seed]
+      end
```
josevalim, line 149:
```suggestion
    seed = Keyword.get_lazy(opts, :seed, &System.system_time/0)
```

josevalim (approved):

josevalim:

PR #65, opened 2023-01-11
Add linear regression livebook

msluszniak:
I need to move some of the functionalities implemented in this livebook to Scholar like:
* correlation matrix (to Scholar.Covariance?)
* MSE and RMSE (to Scholar.Metrics)

Maybe it is worth implementing a stratified split shortly.
Also, I think that would be cool if there will be a button at the beginning of each notebook section such as 'Run all in this branch' and if there will be an option to hide the code of a cell. Some cells with Vega do not introduce any specific value, only the final plot is essential. It will be great to hide such cells.


notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+alias VegaLite, as: Vl
+require Explorer.DataFrame
+```
+
+<!-- livebook:{"branch_parent_index":0} -->
+
+## Linear Regression on Synthetic Data
+
+```elixir
+use Nx.Defn.Kernel
```
polvalente, line 36: I'm not sure if it's a good idea to actually `use Nx.Defn.Kernel` directly here. It has `@doc false`, which makes me think it's not actually a public API in itself.

@josevalim WDYT?

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+
+Firstly, we generate the data which simulates the function $f(x) = 3x + 4$ with added uniform, zero-mean noise. `Nx.Random.uniform` creates a tensor with a given shape and type.
+
+```elixir
+key = Nx.Random.key(42)
+size = 100
+{uniform_x, new_key} = Nx.Random.uniform(key, shape: {size, 1}, type: :f32)
+{uniform_y, _} = Nx.Random.uniform(new_key, shape: {size, 1}, type: :f32)
+x = 2 * uniform_x
+y = 3 * x + 4 + (uniform_y - Nx.mean(uniform_y))
```
polvalente, line 49:
```suggestion
{x, new_key} = Nx.Random.uniform(key, 0, 2shape: {size, 1}, type: :f32)
{noise, _} = Nx.Random.uniform(new_key, -0.5, 0.5, shape: {size, 1}, type: :f32)
y = 3 * x + 4 + noise
```

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+|> Vl.encode_field(:y, "y", type: :quantitative, scale: [domain: [2.5, 12]], axis: [grid: false])
+```
+
+For a regression task, we will use the `Scholar.Linear.LinearRegression` module.
+
+```elixir
+model = Scholar.Linear.LinearRegression.fit(x, y)
+```
+
+As we can see, the coefficient is almost 3.0, and the intercept is nearly 4.0. Those are decent estimations. They are not exactly equal to 3.0 and 4.0 because we introduce noise to our samples.
```
polvalente, line 81:
```suggestion
As we can see, the coefficient is almost 3.0, and the intercept is nearly 4.0. Those are decent estimations. They are not exactly equal to 3.0 and 4.0 because we introduced noise to our sample generation.
```

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+
+Now we will predict one value and draw it on the previous graph in a different color.
+
+```elixir
+x_prediction = Nx.tensor([[0.83]])
+
+y_prediction =
+  Scholar.Linear.LinearRegression.predict(model, x_prediction) |> Nx.reshape({}) |> Nx.to_number()
+
+x_prediction = x_prediction |> Nx.reshape({}) |> Nx.to_number()
```
polvalente, line 136:
```suggestion
x_prediction = Nx.tensor([[0.83]])  |> Nx.reshape({}) |> Nx.to_number()

y_prediction =
  Scholar.Linear.LinearRegression.predict(model, x_prediction)
```

polvalente: If this doesn't work, we should change LinearRegression.predict to accept as a valid input. I see no reason for it to force a 2D tensor as an input. Output should have the same shape as the input, though.

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+  |> Vl.data_from_values(df_prediction)
+  |> Vl.mark(:circle, color: :red, size: 80)
+  |> Vl.encode_field(:x, "x", type: :quantitative)
+  |> Vl.encode_field(:y, "y", type: :quantitative)
+])
+```
+
+As we expected, the red dot lies on the regression line.
+
+This implementation of linear regression is based on the so-called *Least Squares* method. In practice, the function computes $X^+y$ where $X^+$ is a pseudo-inverse matrix (more precisely, Moore-Penrose matrix). You can calculate the results using `Nx.LinAlg.pinv/2`.
```
polvalente, line 180:
```suggestion
This implementation of linear regression is based on the so-called *Least Squares* method. In practice, the function computes $X^+y$ where $X^+$ is the Moore-Penrose pseudo-inverse matrix, calculated using `Nx.LinAlg.pinv/2`.
```

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+  for name <- List.delete(df.names, "ocean_proximity"),
+      do:
+        Vl.new()
+        |> Vl.mark(:bar)
+        |> Vl.encode_field(:x, name, bin: [bin: true, maxbins: 50], axis: [ticks: false])
+        |> Vl.encode_field(:y, "value count", aggregate: :count)
+)
+```
+
+From histograms, we can spot that *median_income* and *median_house_values* have a similar distribution. Both of them are heavy-tailed with high skewness. We might speculate that those two features are strictly correlated. We will check that later on.
```
polvalente, line 246:
```suggestion
From the histograms, we can spot that *median_income* and *median_house_values* have a similar distribution. Both of them are heavy-tailed with high skewness. We might speculate that those two features are strictly correlated. We will check that later on.
```

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+    type: :quantitative,
+    axis: [grid: false],
+    scale: [domain: [32.45, 42.05]]
+  )
+  |> Vl.encode_field(:color, "median_house_value", type: :quantitative, scale: [scheme: :viridis])
+  |> Vl.encode_field(:size, "population", type: :quantitative)
+])
+```
+
+From This plot, we can read that prices are substantially dependent on **geolocalization** and **population**. For geolocalization, we see, those areas closer to the ocean are more expensive. But it's not a strict rule since houses on the northern bay of California are much more affordable than in in-land Mid California. For the population, there are two dense areas with expensive housing: Los Angeles Bay (In South California) and San Francisco Bay (in Mid Califonia). They are metropolises with a lot of different tech companies, and business and cultural institutions, so, logically, housing in those places will be expensive.
```
polvalente, line 368:
```suggestion
From this plot, we can read that prices are substantially dependent on **geolocalization** and **population**. For geolocalization, we see that those areas closer to the ocean are more expensive. But it's not a strict rule since houses on the northern bay of California are much more affordable than in-land Mid California. For the population, there are two dense areas with expensive housing: Los Angeles Bay (In South California) and San Francisco Bay (in Mid California). They are metropoles with a lot of different tech companies, and business and cultural institutions, so, logically, housing in those places will be expensive.
```

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+
+<ins>
+<i>
+Hint:
+</i>
+</ins>
+
+<br />
+
+You can try to add another feature by computing clustering on this data set. It might be a sum or power mean of distances to the clusters. We may predict that centroids will be located in San Francisco Bay and Los Angeles Bay. You can also pass population as weights to k-means.
```
polvalente, line 378:
```suggestion
You can try to add another feature by computing clustering on this data set. It might be a sum or power mean of distances to the clusters. We may predict that centroids will be located in San Francisco Bay and Los Angeles Bay. You can also pass population as weights to K-means.
```

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+```elixir
+df =
+  Explorer.DataFrame.mutate(df,
+    rooms_per_family: total_rooms / households,
+    bedrooms_per_rooms: total_bedrooms / total_rooms,
+    population_per_family: population / households
+  )
+```
+
+In the next step, we will find the correlation matrix. But to do this, we need to cast our data frame to Nx tensor and split data into train and test sets.
```
polvalente, line 399:
```suggestion
In the next step, we will find the correlation matrix. But to do this, we need to cast our data frame to an Nx tensor and split data into train and test sets.
```

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+x =
+  df
+  |> Explorer.DataFrame.discard(["median_house_value"])
+  |> ToTensor.df_to_matrix()
+  |> Nx.take(permutation_indices)
+
+{x, y}
+```
+
+Since we don't have a stratified spilt of data implemented (to learn more see [Stratified Sampling](https://en.wikipedia.org/wiki/Stratified_sampling)), we shuffle data set and take advantage of [Law of large numbers](https://en.wikipedia.org/wiki/Law_of_large_numbers). Now, we will split the data into training and test sets.
```
polvalente, line 462:
```suggestion
Since we don't have a stratified split of data implemented (to learn more see [Stratified Sampling](https://en.wikipedia.org/wiki/Stratified_sampling)), we shuffle data set and take advantage of [Law of large numbers](https://en.wikipedia.org/wiki/Law_of_large_numbers). Now, we will split the data into training and test sets.
```

polvalente: I would either explain how the law of large numbers comes in play here or just not mention it at all

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+
+Before we compute the correlation matrix, we will check if we have NaNs (Not a Number) in the data set.
+
+```elixir
+y_nan_count = Nx.sum(Nx.is_nan(y))
+x_nan_count = Nx.sum(Nx.is_nan(x))
+{x_nan_count, y_nan_count}
+```
+
+Ups, we have some. Fortunately, for y, we don't have any NaNs. If we dig a little bit more, it turns out that NaNs are in <pre> bedrooms_per_rooms (1st row) </pre>  <pre> total_bedrooms (10th row) </pre>
```
polvalente, line 480:
```suggestion
Oops, we have some! Fortunately, for y we don't have any NaNs. If we dig a little bit more, it turns out that NaNs are in <pre> bedrooms_per_rooms (1st row) </pre>  <pre> total_bedrooms (10th row) </pre>
```

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+
+```elixir
+{bedrooms_per_rooms_idx, total_bedrooms_idx} = {0, 9}
+bedrooms_per_rooms_nan_count = Nx.sum(Nx.is_nan(x[[0..-1//1, bedrooms_per_rooms_idx]]))
+total_bedrooms_nan_count = Nx.sum(Nx.is_nan(x[[0..-1//1, total_bedrooms_idx]]))
+Nx.equal(x_nan_count, Nx.add(bedrooms_per_rooms_nan_count, total_bedrooms_nan_count))
+```
+
+For these two, we use `Scholar.Impute.SimpleImputer` with startegy set to median of values. Function `fit` learn the median of features and `transform` for trained model replace all NaNs with a given startegy. It is important that we perform imputation after splitting data because otherwise we will have a leakage of information from test data.
```
polvalente, line 489:
```suggestion
For these two, we use `Scholar.Impute.SimpleImputer` with strategy set to median of values. The `fit` function learns the median of features and `transform` with the trained model replaces all NaNs with a given strategy. It is important that we perform imputation after splitting data because otherwise we will have a leakage of information from test data.
```

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+  defn correlation_matrix_n(x, opts \\ []) do
+    variances =
+      if opts[:biased] do
+        Nx.variance(x, axes: [0])
+      else
+        Nx.variance(x, axes: [0], ddof: 1)
+      end
+
+    Scholar.Covariance.covariance_matrix(x, opts) /
+      Nx.sqrt(Nx.new_axis(variances, 1) * Nx.new_axis(Nx.transpose(variances), 0))
```
polvalente, line 518: Why do you need the transpose here?

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+end
+```
+
+```elixir
+correlation =
+  Nx.concatenate([x_train, Nx.new_axis(y_train, 1)], axis: 1)
+  |> Correlation.correlation_matrix_n(biased: true)
+```
+
+Maybe visual representation would be nicer. 😅
```
polvalente, line 529:
```suggestion
Maybe a visual representation would be nicer. 😅
```

notebooks/linear_regression.livemd
polvalente: Won't `variances` be 1D and as such transpose be a noop?

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+{rmse, mae}
+```
+
+Ok, but is it a good or poor estimation? Huh, check the mean value of the target and then compare it to the value of errors.
+
+```elixir
+Nx.mean(y)
+```
+
+For such a simple model as linear regression, it seems to be a pretty good result. But there is space to improve this result. You can, for example, add some additional features like the logarithm of the population. In the future, you will be able to try more complicated models, such as random forests.
```
polvalente, line 610: Why mention the logarithm of the population here? You should either explain what that would accomplish, or remove this comment from the conclusion.

Also, I expected the skewness of the income and the house values to have come in play in the algorithm as well.

I though you'd have used log(1 + p) on both of those features and the applied the inverse function for the final prediction. I think either we do that, or just don't comment on the skewness at all.

polvalente:
@msluszniak regarding covariance matrix, RMSE and MSE, I agree that all of these could be extracted to Scholar :)

josevalim:
> @msluszniak regarding covariance matrix, RMSE and MSE, I agree that all of these could be extracted to Scholar :)

me too! So please extract those out @msluszniak as we prepare this notebook!

notebooks/linear_regression.livemd
msluszniak: In line 46 there is no comma between 0 and 2, I will add this change separately.

notebooks/linear_regression.livemd
msluszniak: I think that this might have a reason. For training, we pass 2D data which has `n_samples` and `n_features`. So it somehow forces input for prediction to have the same `n_features` which is the first axis. But features are considered usually as the second axis.  For scalar tensors, we can guess that input has one row and one feature (but we virtually add those axes). For one-dimensional input, it will be ambiguous if the input has one row with n features or n rows with 1 feature. There is no specific reason to assume one version. If we wrap it into a 2D tensor then it is clear what we mean.

msluszniak: I believe that's why sklearn also forces users to use only 2D input.

notebooks/linear_regression.livemd
msluszniak: Yeah, I will add more explanation about the law of large numbers. I find it important to explain the sampling process.

notebooks/linear_regression.livemd
msluszniak: Yes, you are totally right. It is redundant.

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,610 @@
...
+model = Scholar.Linear.LinearRegression.fit(x_train, y_train)
+```
+
+Now we can predict the values for the test set and measure the error of our prediction. We will calculate root mean square error (RMSE) and mean absolute error (MAE).
+
+```elixir
+defmodule RMSE do
+  import Nx.Defn
+
+  defn root_mean_square_error(y_true, y_pred) do
```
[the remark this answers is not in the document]
msluszniak: Yes, I will write this functionality.

notebooks/linear_regression.livemd
msluszniak: Ok, I will remove the logarithm of the population from the conclusion because it doesn't make sense.  I also tried with log(1+p) of mentioned features and it turns out to make prediction worse than without logs. However, I would like to keep the mention of skewness. I made it mostly because of the correlation matrix at the end of the notebook.

notebooks/linear_regression.livemd
polvalente: I don't think it would be ambiguous if we define/standardize on 1D tensors being a single example (i.e. 1 row, n features)

polvalente (approved):

notebooks/linear_regression.livemd
msluszniak: Ok, I will introduce such feature in a separate PR and then update the livebook.

polvalente (approved):

notebooks/linear_regression.livemd
polvalente: @msluszniak last thing here. I confirmed that this is a private API, so we need to remove this `use`. I think this only affects the next cell, right?

Perhaps just adding a module around it will suffice

msluszniak: Sure

notebooks/linear_regression.livemd
```diff
@@ -0,0 +1,582 @@
...
+require Explorer.DataFrame
+```
+
+<!-- livebook:{"branch_parent_index":0} -->
+
+## Linear Regression on Synthetic Data
+
+```elixir
+use Nx.Defn.Kernel
+```
```
msluszniak, line 37:
```suggestion
```

notebooks/linear_regression.livemd
polvalente: Resolving this in favor of the PR

polvalente (approved):

PR #67, opened 2023-01-16
Add MSE

msluszniak:
Mentioned in #65, but I added only MSE since RMSE is just a square root of MSE and popular packages usually implement only MSE.


josevalim (approved):

josevalim:

PR #68, opened 2023-01-16
Remove opts from defnp

msluszniak:
In Nx 0.4.2, opts in defnp are redundant. In the formatter for elixir 1.13, there is a bug that causes wrong formatting in a very specific case. I propose to use elixir 1.14 in CI.


josevalim (approved):

josevalim:

josevalim:
@msluszniak btw, Nx v0.4.2 also changed Nx.tensor to accept tensors. which means that all places that accept weights as lists can now accept tensors but we will most likely need to change the NimbleOptions validator to a custom one that also accepts tensors. do you want to look into that too?

msluszniak:
> @msluszniak btw, Nx v0.4.2 also changed Nx.tensor to accept tensors. which means that all places that accept weights as lists can now accept tensors but we will most likely need to change the NimbleOptions validator to a custom one that also accepts tensors. do you want to look into that too?

Sure

PR #69, opened 2023-01-16
Polynomial feature transform support

matiascr:
## Added new polynomial features matrix support in the polynomial.ex file. 

It creates the polynomial feature matrix of an input matrix, and has optional values `degree` set to `default=2` and `fit_intercept?` set to `true` by default.

Can be further optimized in the future using `defn`s. I don't know too how to go about it (I'm not an expert in `defn`'s capabilities), so I leave it here so that it can be further improved in the future (suggestions are welcome).

Tests are included. I apologize in advance for the 20+ line tensors in some of the expected results (in my defense, it's `mix format`'s fault), but I wanted to show that the algorithm was scalable by trying high degrees and high number of features.

I am aware that this is not a 1:1 equivalent to scipy. It's missing `interaction_only` and some other things, but I can get to work to add them if you feel that they'd be useful.


josevalim:
Thank you @matiascr for the PR. However it is very important for all functionality to be implements with defn, as that gives orders of magnitude improvements to performance.

For this reason we expect all functionality to be written using defn, otherwise we won’t be able to accept the PR unfortunately.

matiascr:
Understood. I'll keep working on it.
Is there any preferred way to iterate over tensors in defn? I saw that `Nx.map` is discouraged in the docs.

josevalim:
You most likely want to use while/for from Nx.Defn.Kernel. :)

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,189 @@
...
+  Fits a polynomial regression model for sample inputs `a` and
+  sample targets `b`.
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+    The function returns a struct with the following parameters:
```
josevalim, line 51:
```suggestion
  The function returns a struct with the following parameters:
```

matiascr:
Thanks for the feedback.
By the way, what do you think about having a `fit` (and later `predict`) function for Polynomial Regression. Instead of having to do the intermediate step of doing the transform and then input it for the `fit` and `predict` function (like in scikit), I think it would be more intuitive to have a PolynomialRegression module. This would do the intermediate `transform` step in the background, but it would actually map to LinearRegression but using polynomial features.

In Scikit:
```python3
poly = PolynomialFeatures(degree=2, include_bias=False)
X_poly = poly.fit_transform(X)
model = LinearRegression().fit(X_poly, y)
poly_model.predict(X_poly)
```
vs Scholar:
```elixir
Scholar.Linear.PolynomialRegression.fit(x, degree: 2, fit_intercept?: false)
|> Scholar.Linear.PolynomialRegression.predict(x)
```

If this is of no interest I can just leave it with the transform functionality only.

josevalim:
I think we have `fit_transform` in other Scholar modules, so it would be welcome here!

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,190 @@
...
+        >, intercept: #Nx.Tensor<
+          f32
+          5.896470069885254
+        >
+      }
+
+      iex> x = Nx.tensor([[1.0, 2.0], [3.0, 2.0], [4.0, 7.0]])
+      iex> y = Nx.tensor([4.0, 3.0, -1.0])
+      iex> model = Scholar.Linear.PolynomialRegression.fit(x, y, degree: 2)
+      %Scholar.Linear.LinearRegression{
```
msluszniak, line 75:
```suggestion
      %Scholar.Linear.PolynomialRegression{
```
And analogical

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,190 @@
+defmodule Scholar.Linear.PolynomialRegression do
+  @moduledoc """
+  Least squares polynomial regression.
+  """
+
+  import Nx.Defn
+
+  # @derive {Nx.Container, containers: [:coefficients, :intercept]}
+  # defstruct [:coefficients, :intercept]
```
msluszniak, line 9: We eventually would like to return the struct with coef and intercept. So this might be useful.

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,190 @@
...
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+  The function returns a struct with the following parameters:
+
+    * `:coefficients` - Estimated coefficients for the linear regression problem.
```
msluszniak, line 53:
```suggestion
    * `:coefficients` - Estimated coefficients for the polynomial regression problem.
```

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,190 @@
...
+
+    * `:intercept` - Independent term in the linear model.
```
msluszniak, line 55:
```suggestion
    * `:intercept` - Independent term in the polynomial model.
```

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,190 @@
+defmodule Scholar.Linear.PolynomialRegression do
```
msluszniak, line 1: I think this might not be the correct localization of this module.

matiascr: Where do you think it makes the most sense to have this module?

polvalente: Given we have Scholar.Preprocessing and that sklearn has this under preprocessing, perhaps Scholar.Preprocessing.PolynomialRegression would fit?

However, I'm ok with keeping this here. Why do you think it should be elsewhere, @msluszniak ?

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,190 @@
...
+      |> compute_column(x, nf)
+    end)
+  end
+
+  defnp compute_column(previous, x, n) do
+    {n_samples, _n_features} = Nx.shape(x)
+
+    x[[0..-1//1, n]]
+    |> Nx.reshape({n_samples, :auto})
+    |> Nx.multiply(previous)
```
polvalente, line 177: multiplying by some sort of previous value sounds like we could apply Nx.cumulative_product somehow for this scaling, but I'm not sure about that

polvalente: perhaps finding a reference implementation for this in scipy, jax or sklearn would be helpful

polvalente: by this I mean the whole calculation, not just this `compute_degree` function

lib/scholar/linear/polynomial_regression.ex
matiascr: I think that if this module has `fit` and `predict` functions and it forms a struct similar to that of LinearRegression, it makes sense to have the module next to LinearRegression

lib/scholar/linear/polynomial_regression.ex
matiascr: > multiplying by some sort of previous value sounds like we could apply Nx.cumulative_product somehow for this scaling, but I'm not sure about that

"previous" makes reference to the previous degree iteration of the operation, so the columns that are multiplied are not (usually) adjacent

lib/scholar/linear/polynomial_regression.ex
josevalim: :+1: for `Scholar.Preprocessing.PolynomialRegression`.

josevalim: Actually, @polvalente convinced me to keep it in Linear. :)

msluszniak: Yeah, I also though about it and now i'm also in favour to keep it in Linear

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,215 @@
...
+      doc: """
+      The weights for each observation. If not provided,
+      all observations are assigned equal weight.
+      """
+    ],
+    degree: [
+      type: :pos_integer,
+      default: 2,
+      doc: """
+      The degree of the feature matrix to return. Must be a >1 integer. 1
```
polvalente, line 23:
```suggestion
      The degree of the feature matrix to return. Must be an integer greater than 1. 1
```

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,215 @@
...
+      }
+  """
+  deftransform fit(a, b, opts \\ []) do
+    opts = NimbleOptions.validate!(opts, @opts_schema)
+    a_transform = transform(a, opts |> Keyword.put(:fit_intercept?, false))
+
+    %{
+      Scholar.Linear.LinearRegression.fit(a_transform, b, Keyword.delete(opts, :degree))
+      | __struct__: Scholar.Linear.PolynomialRegression
+    }
```
polvalente, line 95:
```suggestion
    %__MODULE__{
      Scholar.Linear.LinearRegression.fit(a_transform, b, Keyword.delete(opts, :degree))
    }
```

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,215 @@
...
+    |> Map.merge(%{degree: opts[:degree]})
```
polvalente, line 96:
```suggestion
    linear_reg = Scholar.Linear.LinearRegression.fit(a_transform, b, Keyword.delete(opts, :degree))
    %__MODULE__{
      coefficients: linear_reg.coefficients,
      intercept: linear_reg.intercept,
      degree: opts[:degree]
    }
```

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,215 @@
...
+      #Nx.Tensor<
+        f32[1]
+        [3.84888117]
+      >
+  """
+  deftransform predict(model, x) do
+    Scholar.Linear.LinearRegression.predict(
+      %{model | __struct__: Scholar.Linear.LinearRegression},
+      transform_n(x, degree: model.degree, fit_intercept?: false)
+    )
```
polvalente, line 117:
```suggestion
    Scholar.Linear.LinearRegression.predict(
      %Scholar.Linear.LinearRegression{coefficients: model.coefficients, intercept: model.intercept},
      transform_n(x, degree: model.degree, fit_intercept?: false)
    )
```

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,215 @@
...
+      previous_degree
+      |> Enum.slice(nf..-1)
+      |> Nx.concatenate(axis: 1)
+      |> compute_column(x, nf)
+    end)
+  end
+
+  defnp get_column(x, n) do
+    Nx.transpose(x)[n]
+    |> Nx.reshape({:auto, 1})
```
polvalente, line 198:
```suggestion
    Nx.slice_along_axis(x, n, 1, axis: 1)
```

polvalente (approved):
@josevalim @msluszniak this PR looks good to me.

Any blockers for merging?

test/scholar/linear/polynomial_regression_test.exs
```diff
@@ -0,0 +1,432 @@
...
+            45,
+            63,
+            75,
+            105,
+            147,
+            125,
+            175,
+            245,
+            343
+          ]
```
josevalim, line 385: Can you please write this using only few lines using `[...] ++ [...] ++ [...]`?

test/scholar/linear/polynomial_regression_test.exs
```diff
@@ -0,0 +1,432 @@
...
+            108,
+            162,
+            243,
+            64,
+            96,
+            144,
+            216,
+            324,
+            486,
+            729
```
josevalim, line 424: Same here.

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,221 @@
...
+          [2, 3, 4, 6, 9],
+          [4, 5, 16, 20, 25]
+        ]
+      >
+  """
+  deftransform transform(x, opts \\ []) do
+    transform_n(x, NimbleOptions.validate!(opts, @transform_opts_schema))
+  end
+
+  deftransform transform_n(x, opts) do
```
josevalim, line 170: This should be `deftransformp`, no?

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,221 @@
...
+        _, prev_degree ->
+          compute_degree(x, prev_degree)
+      end
+    )
+    |> List.flatten()
+    |> Nx.concatenate(axis: 1)
+    |> add_intercept(opts)
+  end
+
+  @spec compute_degree(Nx.Tensor, list(Nx.Tensor)) :: list(Nx.Tensor)
```
josevalim, line 191:
```suggestion
```

josevalim:
Sorry, this fell through the cracks. I have added additional comments. I assume it is ok that the API is mostly `deftransform` because they are all "meta" operations on the shape anyway?

test/scholar/linear/polynomial_regression_test.exs
```diff
@@ -0,0 +1,295 @@
...
+
+      actual_prediction =
+        Scholar.Linear.PolynomialRegression.fit(a, b,
+          degree: 2,
+          sample_weights: sample_weights,
+          fit_intercept?: true
+        )
+        |> Scholar.Linear.PolynomialRegression.predict(prediction_input)
+
+      assert_all_close(expected_prediction, actual_prediction, rtol: 1.0e-1, atol: 1.0e-1)
```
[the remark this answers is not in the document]
matiascr: It used to be more innaccurate. Fixed

notebooks/polynomial_regression.livemd
```diff
@@ -0,0 +1,157 @@
...
+  # config: [nx: [default_backend: EXLA.Backend]]
+)
+```
+
+## Section
+
+```elixir
+defmodule Polynomial do
+  import Nx.Defn
+
```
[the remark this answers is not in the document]
matiascr: This module is a work in progress and does not do everything yet. It's an implementation using only `defn`s, but only simple polynomial regression, not multiple.
Probably it should be removed from the PR.

notebooks/polynomial_regression.livemd
```diff
@@ -0,0 +1,157 @@
+# defn transform
```
josevalim, line 1: We should either convert this into a guide (see the others as an example) or remove this notebook. :)

matiascr: Could it be better to update the linear regression guide with polynomial regression also in it? Instead of having one with just polynomial?

lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,221 @@
...
+      #Nx.Tensor<
+        s64[3][5]
+        [
+          [0, 1, 0, 0, 1],
+          [2, 3, 4, 6, 9],
+          [4, 5, 16, 20, 25]
+        ]
+      >
+  """
+  deftransform transform(x, opts \\ []) do
```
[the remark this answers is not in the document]
josevalim: Ping. :)

test/scholar/linear/polynomial_regression_test.exs
```diff
@@ -0,0 +1,298 @@
+defmodule Scholar.Linear.PolynomialRegressionTest do
+  use Scholar.Case, async: true
+  # doctest Scholar.Linear.PolynomialRegression
```
josevalim, line 3: Should this be uncommented?

matiascr: Yes, currently there is a compilation error that I can't figure out with those doctests
```
  1) doctest Scholar.Linear.PolynomialRegression.fit/3 (2) (Scholar.Linear.PolynomialRegressionTest)
     test/scholar/linear/polynomial_regression_test.exs:3
     Doctest did not compile, got: (SyntaxError) lib/scholar/linear/polynomial_regression.ex:91:5: syntax error before: '['
         |
      91 |     [-0.021396497264504433, -0.004854594357311726, -0.0884987860918045, -0.062211357057094574, -0.04369127005338669]
         |     ^
     doctest:
       iex> x = Nx.tensor([[1.0, 2.0], [3.0, 2.0], [4.0, 7.0]])
       iex> y = Nx.tensor([4.0, 3.0, -1.0])
       iex> Scholar.Linear.PolynomialRegression.fit(x, y, degree: 2)
       %Scholar.Linear.PolynomialRegression{
         coefficients: #Nx.Tensor<
           f32[5]
           [-0.021396497264504433, -0.004854594357311726, -0.0884987860918045, -0.062211357057094574, -0.04369127005338669]
         >,
         intercept: #Nx.Tensor<
           f32
           4.418517112731934
         >,
         degree: 2
       }
     stacktrace:
       lib/scholar/linear/polynomial_regression.ex:88: Scholar.Linear.PolynomialRegression (module)
```
Maybe you're familiar with this error and know how what could be the cause

josevalim: Yeah, you can't inspect with tensors. You can see other examples, we typically do this:

```
       iex> Scholar.Linear.PolynomialRegression.fit(x, y, degree: 2)
       %Scholar.Linear.PolynomialRegression{
         coefficients: Nx.tensor([-0.021396497264504433, -0.004854594357311726, -0.0884987860918045, -0.062211357057094574, -0.04369127005338669]),
         intercept: Nx.tensor(4.418517112731934),
         degree: 2
       }
```

matiascr: I just saw that linear regression had the same problem, I'll do that change for both

josevalim: Yup, your approach is also great. :)

matiascr:
So... this is a mess. I don't know how I messed this up so badly. I'll create a new PR with only the changes regarding polynomial regression.

msluszniak:
If I understand correctly, we can close this PR, right?

PR #70, opened 2023-01-17
Weights as a Tensor

msluszniak:


lib/scholar/cluster/k_means.ex
```diff
@@ -364,8 +365,8 @@ defmodule Scholar.Cluster.KMeans do
   end
 
   deftransformp validate_weights(weights, num_samples, num_runs) do
-    if is_nil(weights) or
-         (is_list(weights) and length(weights) == num_samples) do
+    if is_nil(weights) or (is_list(weights) and length(weights) == num_samples) or
+         (Nx.is_tensor(weights) and Nx.shape(weights) == {num_samples}) do
```
polvalente, line 369: I'd assert on the shape after converting from a list, because otherwise your list can be nested and would become an N-dimensional tensor instead

lib/scholar/options.ex
```diff
@@ -30,4 +32,32 @@ defmodule Scholar.Options do
       {:error, "expected a non-negative number, got: #{inspect(num)}"}
     end
   end
+
+  def weights(weights) do
+    if Nx.is_tensor(weights) or is_list(weights) do
```
polvalente, line 37: You should ensure it's a flat list

lib/scholar/options.ex
```diff
@@ -30,4 +32,32 @@ defmodule Scholar.Options do
...
+      {:ok, weights}
+    else
+      {:error,
+       "expected weights to be a tensor or a list of positive numbers, got: #{inspect(weights)}"}
+    end
+  end
+
+  def non_negative_weights(weights) do
+    if (Nx.is_tensor(weights) and Nx.to_number(Nx.all(Nx.greater_equal(weights, 0.0))) == 1) or
+         (is_list(weights) and Enum.all?(weights, fn x -> x >= 0.0 end)) do
```
polvalente, line 56: I wonder if x being a list would compare as bigger than a float. I'm almost sure it would

lib/scholar/options.ex
```diff
@@ -30,4 +32,37 @@ defmodule Scholar.Options do
       {:error, "expected a non-negative number, got: #{inspect(num)}"}
     end
   end
+
+  def weights(weights) do
+    if (Nx.is_tensor(weights) and Nx.to_number(Nx.rank(weights)) == 1) or
+         (is_list(weights) and List.flatten(weights) == weights) do
```
polvalente, line 38:
```suggestion
         (is_list(weights) and Enum.all?(weights, &is_number/1)) do
```

lib/scholar/options.ex
```diff
@@ -30,4 +32,37 @@ defmodule Scholar.Options do
...
+      {:ok, weights}
+    else
+      {:error, "expected weights to be a flat tensor or a flat list, got: #{inspect(weights)}"}
+    end
+  end
+
+  def positive_weights(weights) do
+    if (Nx.is_tensor(weights) and Nx.to_number(Nx.rank(weights)) == 1 and
+          Nx.to_number(Nx.all(Nx.greater(weights, 0.0))) == 1) or
+         (is_list(weights) and List.flatten(weights) == weights and
```
polvalente, line 48:
```suggestion
         (is_list(weights) and Enum.all?(weights, &is_number/1) and
```

lib/scholar/options.ex
```diff
@@ -30,4 +32,37 @@ defmodule Scholar.Options do
...
+            Enum.all?(weights, fn x -> x > 0.0 end)) do
+      {:ok, weights}
```
polvalente, line 50:
```suggestion
         (is_list(weights) and Enum.all?(weights, fn x -> is_number(x) and x > 0.0 end)) do
      {:ok, weights}
```

lib/scholar/options.ex
```diff
@@ -30,4 +32,37 @@ defmodule Scholar.Options do
...
+      {:error,
+       "expected weights to be a flat tensor or a flat list of positive numbers, got: #{inspect(weights)}"}
+    end
+  end
+
+  def non_negative_weights(weights) do
+    if (Nx.is_tensor(weights) and Nx.to_number(Nx.rank(weights)) == 1 and
+          Nx.to_number(Nx.all(Nx.greater_equal(weights, 0.0))) == 1) or
+         (is_list(weights) and List.flatten(weights) == weights and
+            Enum.all?(weights, fn x -> x >= 0.0 end)) do
```
polvalente, line 61:
```suggestion
         (is_list(weights) and Enum.all?(weights, fn x -> is_number(x) and x >= 0.0 end)) do
```

polvalente (approved):

lib/scholar/options.ex
```diff
@@ -30,4 +32,35 @@ defmodule Scholar.Options do
...
+      {:ok, weights}
+    else
+      {:error,
+       "expected weights to be a flat tensor or a flat list of positive numbers, got: #{inspect(weights)}"}
+    end
+  end
+
+  def non_negative_weights(weights) do
+    if (Nx.is_tensor(weights) and Nx.to_number(Nx.rank(weights)) == 1 and
+          Nx.to_number(Nx.all(Nx.greater_equal(weights, 0.0))) == 1) or
```
josevalim, line 58:
```suggestion
    if (Nx.is_tensor(weights) and Nx.rank(weights) == 1) or
```

lib/scholar/cluster/k_means.ex
josevalim, line 369: I think you can break this into several blocks:

```
cond do
  is_nil(weights) ->
    1.0
  Nx.is_tensor(weights)  and Nx.shape(weights) == {num_samples} ->
    weights
  is_list(weights) and length(weights) == num_samples ->
    Nx.tensor(weights, type: :f32)
  true ->
    raise "..."
end
```

And then you do the `Nx.as_type(:f32) |> Nx.broadcast({num_runs, num_samples})` inside the defn.

lib/scholar/cluster/k_means.ex
```diff
@@ -143,17 +144,17 @@ defmodule Scholar.Cluster.KMeans do
...
+    weights = validate_weights(weights, num_samples, opts[:num_runs])
 
-    fit_n(x, seed, opts)
+    fit_n(x, weights, seed, opts)
   end
 
-  defnp fit_n(x, seed, opts) do
-    {num_samples, num_features} = Nx.shape(x)
+  defnp fit_n(%Nx.Tensor{shape: {num_samples, num_features}} = x, weights, seed, opts) do
+    # {num_samples, num_features} = Nx.shape(x)
```
josevalim, line 154:
```suggestion
```

lib/scholar/options.ex
```diff
@@ -30,4 +32,35 @@ defmodule Scholar.Options do
...
+         (is_list(weights) and Enum.all?(weights, &is_number/1)) do
+      {:ok, weights}
+    else
+      {:error, "expected weights to be a flat tensor or a flat list, got: #{inspect(weights)}"}
+    end
+  end
+
+  def positive_weights(weights) do
+    if (Nx.is_tensor(weights) and Nx.to_number(Nx.rank(weights)) == 1 and
+          Nx.to_number(Nx.all(Nx.greater(weights, 0.0))) == 1) or
```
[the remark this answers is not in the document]
msluszniak: These checks will run in `deftransform` in `NimbleOptions.validate`. If this is insufficient then I will use only `Options.weights` and remove the rest of the functions and test which check for positive weights.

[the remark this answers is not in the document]
josevalim: We can check for lists but not for tensors because we cannot raise from the GPU. So if a bad tensor is given, it will just give wrong results.

josevalim:

PR #71, opened 2023-01-17
Allow 1D and Scalar Input to LinearRegression.predict

msluszniak:
As mentioned in #65, I add 1d and scalar input in predict.

### DIFF
--- a/lib/scholar/linear/linear_regression.ex
+++ b/lib/scholar/linear/linear_regression.ex
@@ -103,6 +103,11 @@ defmodule Scholar.Linear.LinearRegression do
   @doc """
   Makes predictions with the given `model` on input `x`.
 
+  If the input is scalar or 1D tensor it is assumed that
+  it represents one sample with `n` features (columns).
+
+  Predictions have always the same size of the first axis as input data.
+
   ## Examples
 
       iex> x = Nx.tensor([[1.0, 2.0], [3.0, 2.0], [4.0, 7.0]])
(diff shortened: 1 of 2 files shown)


polvalente (commented):
@msluszniak Now that I see the PR, I'm not 100% sure that we should accept a 1D tensor here.

Thinking specifically about the case where we have 1 feature (and therefore the angular coefficient is shaped {1}), one could expect to be able to pass a tensor with n samples and the current implementation would accept it but raise either on the `reshape` or on the `dot`

Any thoughts?

msluszniak:
> @msluszniak Now that I see the PR, I'm not 100% sure that we should accept a 1D tensor here.
> 
> Thinking specifically about the case where we have 1 feature (and therefore the angular coefficient is shaped {1}), one could expect to be able to pass a tensor with n samples and the current implementation would accept it but raise either on the `reshape` or on the `dot`
> 
> Any thoughts?

Yeah, I agree that this might be confusing.

polvalente:
> > @msluszniak Now that I see the PR, I'm not 100% sure that we should accept a 1D tensor here.
> > Thinking specifically about the case where we have 1 feature (and therefore the angular coefficient is shaped {1}), one could expect to be able to pass a tensor with n samples and the current implementation would accept it but raise either on the `reshape` or on the `dot`
> > Any thoughts?
> 
> Yeah, I agree that this might be confusing.

Let's close this then? Thanks for exploring this anyway!

PR #72, opened 2023-01-22
Add r2_score metric

maartenvanvliet:
This PR adds the [R2^2 metric](https://en.wikipedia.org/wiki/Coefficient_of_determination#Relation_to_unexplained_variance) to the Metrics module.

I first named it r_squared but went with r2_score since that's what [sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.r2_score.html#sklearn.metrics.r2_score) uses.

### DIFF
--- a/lib/scholar/metrics.ex
+++ b/lib/scholar/metrics.ex
@@ -510,6 +510,32 @@ defmodule Scholar.Metrics do
     (diff * diff) |> Nx.mean()
   end
 
+  @doc ~S"""
+  Calculates the $R^2$ score of predictions with respect to targets.
+
+  $$R^2 = 1 - \frac{\sum (y_i - \hat{y}_i)^2}{\sum (y_i - \bar{y})^2}$$
+
+  ## Examples
+
+      iex> y_true = Nx.tensor([3, -0.5, 2, 7], type: {:f, 32})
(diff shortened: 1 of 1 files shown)


msluszniak (changes requested):
Thanks for the PR. I think we can add other parameters to this function like `:sample_weights` and `:force_finite`.

lib/scholar/metrics.ex
```diff
@@ -510,6 +510,32 @@ defmodule Scholar.Metrics do
...
+      iex> y_pred = Nx.tensor([2.5, 0.0, 2, 8], type: {:f, 32})
+      iex> Scholar.Metrics.r2_score(y_true, y_pred)
+      #Nx.Tensor<
+        f32
+        0.9486081600189209
+      >
+  """
+  defn r2_score(y_true, y_pred) do
+    diff = y_true - y_pred
+    ssr = (diff * diff) |> Nx.sum()
```
polvalente, line 530: it seems that `ssr = Scholar.Metrics.Distances.squared_euclidean(y_true, y_pred)` is a more expressive way to write this, and `sst = Scholar.Metrics.Distances.squared_euclidean(y_true, Nx.mean(y_true))` would also work

PR #80, opened 2023-02-18
Update KMeans notebook

msluszniak:
Except for KMeans, I added `Nx.Constants.pi()` and reformat some cells from the Linear Regression notebook.


josevalim:

PR #81, opened 2023-02-23
Add K-Nearest Neighbors

msluszniak:
I need to add tests

### DIFF
--- a/lib/scholar/neighbors/k_nearest_neighbors.ex
+++ b/lib/scholar/neighbors/k_nearest_neighbors.ex
@@ -38,7 +38,8 @@ defmodule Scholar.Neighbors.KNearestNeighbors do
       default: {:minkowski, 2},
       doc: ~S"""
       Name of the metric. Possible values:
-      * `{:minkowski, p}` - Minkowski metric. By changing value of `p` parameter we can set Euclidean, Chebyshev, Manhattan or any arbitrary $L_p$ metric.
+      * `{:minkowski, p}` - Minkowski metric. By changing value of `p` parameter (a positive number or :infinity)
+        we can set Manhattan (1), Euclidean (2), Chebyshev (:infinity), or any arbitrary $L_p$ metric.
 
       * `:cosine` - Cosine metric.
       """
--- a/lib/scholar/neighbors/k_nearest_neighbors.ex
+++ b/lib/scholar/neighbors/k_nearest_neighbors.ex
@@ -1,0 +1,401 @@
+defmodule Scholar.Neighbors.KNearestNeighbors do
+  @moduledoc """
+  The K-Nearest Neighbors.
+  """
+  import Nx.Defn
+  require Nx
+
+  @derive {Nx.Container,
+           keep: [:default_num_neighbors, :weights, :num_classes, :p, :task, :metric],
+           containers: [:data, :labels]}
+  defstruct [:data, :labels, :default_num_neighbors, :weights, :num_classes, :p, :task, :metric]
+
+  opts = [
+    num_neighbors: [
+      type: :pos_integer,
+      default: 5,
+      doc: "The number of neighbors to use by default for `k_neighbors` queries"
+    ],
+    num_classes: [
+      type: :pos_integer,
+      required: true,
+      doc: "Number of classes in provided labels"
+    ],
+    weights: [
+      type: {:in, [:uniform, :distance]},
+      default: :uniform,
+      doc: """
+      Weight function used in prediction. Possible values:
+
+        * `:uniform` - uniform weights. All points in each neighborhood are weighted equally.
+
+        * `:distance` - weight points by the inverse of their distance. in this case, closer neighbors of
+          a query point will have a greater influence than neighbors which are further away.
+      """
+    ],
+    p: [
+      type: {:custom, Scholar.Options, :positive_number, []},
+      default: 2,
+      doc: """
+      Power parameter for the Minkowski metric. When p = 1, this is equivalent
+      to using manhattan_distance ($L_1$), and euclidean_distance ($L_2$) for p = 2.
+      """
+    ],
+    metric: [
+      type: {:in, [:minkowski, :cosine]},
+      default: :minkowski,
+      doc: ~S"""
+      Name of the metric. Possible values:
+      * `:minkowski` - Minkowski metric. By changing value of `:p` parameter we can set Euclidean, Chebyshev, Manhattan or any arbitrary $L_p$ metric.
+
+      * `:cosine` - Cosine metric.
+      """
+    ],
+    task: [
+      type: {:in, [:classification, :regression]},
+      default: :classification,
+      doc: """
+      Task that will be performed using K Nearest Neighbors. Possible values:
+
+      * `:classification` - Classifier implementing the k-nearest neighbors vote.
+
+      * `:regression` - Regression based on K-Nearest Neighbors.
+        The target is predicted by local interpolation of the targets associated of the nearest neighbors in the training set.
+      """
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(opts)
+
+  @doc """
+  Fit the K-nearest neighbors classifier from the training data set.
+
+  For classification, provided labels needs to be consecutive natural numbers. If your labels does
+  not meet this condition please use `Scholar.Preprocessing.ordinal_encode`
+
+  Currently 2D labels only supported for regression task.
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+    The function returns a struct with the following parameters:
+
+    * `:data` - Training data.
+
+    * `:labels` - Labels of each point.
+
+    * `:default_num_neighbors` - The number of neighbors to use by default for `k_neighbors` queries.
+
+    * `:weights` - Weight function used in prediction.
+
+    * `:num_classes` - Number of classes in provided labels.
+
+  ## Examples
+
+      iex>  Scholar.Neighbors.KNearestNeighbors.fit(Nx.tensor([[1, 2], [2, 4], [1, 3], [2, 5]]), Nx.tensor([1, 0, 1, 1]),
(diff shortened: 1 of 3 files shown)


lib/scholar/neighbors/k_nearest_neighbors.ex
```diff
@@ -0,0 +1,401 @@
...
+    ],
+    metric: [
+      type: {:in, [:minkowski, :cosine]},
+      default: :minkowski,
+      doc: ~S"""
+      Name of the metric. Possible values:
+      * `:minkowski` - Minkowski metric. By changing value of `:p` parameter we can set Euclidean, Chebyshev, Manhattan or any arbitrary $L_p$ metric.
+
+      * `:cosine` - Cosine metric.
+      """
```
josevalim, line 52: What if we merge `metric` and `p`? We could do `metric: :cosine` and `metric: {:minkowski, 1}`?

lib/scholar/neighbors/k_nearest_neighbors.ex
```diff
@@ -0,0 +1,401 @@
...
+
+    * `:default_num_neighbors` - The number of neighbors to use by default for `k_neighbors` queries.
+
+    * `:weights` - Weight function used in prediction.
+
+    * `:num_classes` - Number of classes in provided labels.
+
+  ## Examples
+
+      iex>  Scholar.Neighbors.KNearestNeighbors.fit(Nx.tensor([[1, 2], [2, 4], [1, 3], [2, 5]]), Nx.tensor([1, 0, 1, 1]),
```
josevalim, line 98:
```suggestion
      iex> foo = Nx.tensor([[1, 2], [2, 4], [1, 3], [2, 5]])
      iex>  Scholar.Neighbors.KNearestNeighbors.fit(foo, Nx.tensor([1, 0, 1, 1]),
```

But please give it a better name to the tensor. :D

lib/scholar/neighbors/k_nearest_neighbors.ex
```diff
@@ -0,0 +1,401 @@
...
+
+  Currently 2D labels only supported for regression task.
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+    The function returns a struct with the following parameters:
```
josevalim, line 84:
```suggestion
  The function returns a struct with the following parameters:
```

lib/scholar/neighbors/k_nearest_neighbors.ex
```diff
@@ -0,0 +1,401 @@
...
+  Find the K-neighbors of a point.
+
+  ## Return Values
+
+    Returns indices of and distances to the neighbors of each point.
+
+  ## Examples
+
+      iex> model =
+      iex>  Scholar.Neighbors.KNearestNeighbors.fit(Nx.tensor([[1, 2], [2, 4], [1, 3], [2, 5]]), Nx.tensor([1, 0, 1, 1]),
```
josevalim, line 273: Same here, let's define the tensor separately so the code is less overloaded. :)

josevalim (approved):
LGTM! I have only added some minor notes.

Btw, is this a good candidate for a notebook? :)

lib/scholar/neighbors/k_nearest_neighbors.ex
```diff
@@ -0,0 +1,401 @@
...
+      end
+
+    indices =
+      Nx.stack(
+        [Nx.iota(Nx.shape(pred_labels), axis: 0), Nx.take(labels, pred_labels)],
+        axis: -1
+      )
+      |> Nx.flatten(axes: [0, 1])
+
+    proba = Nx.indexed_add(proba, indices, Nx.flatten(weights_vals))
```
polvalente, line 259:
```suggestion
    proba = Nx.indexed_put(proba, indices, Nx.flatten(weights_vals))
```

indexed_put seems a bit more semantic given it's all zeros anyway

polvalente: nevermind this comment :)

lib/scholar/neighbors/k_nearest_neighbors.ex
```diff
@@ -0,0 +1,401 @@
...
+    {tensor, weights, axis} =
+      if axis == 0,
+        do: {Nx.transpose(tensor), Nx.transpose(weights), 1},
+        else: {tensor, weights, axis}
+
+    {num_samples, num_features} = tensor_shape = Nx.shape(tensor)
+
+    indices = Nx.argsort(tensor, axis: axis)
+
+    sorted = Nx.sort(tensor, axis: axis)
```
polvalente, line 351:
```suggestion
    indices = Nx.argsort(tensor, axis: axis)

    sorted = Nx.take_along_axis(tensor, indices, axis: axis)
```

polvalente: This avoids sorting twice

lib/scholar/neighbors/k_nearest_neighbors.ex
```diff
@@ -0,0 +1,401 @@
...
+    counting_indices =
+      [
+        Nx.shape(group_indices)
+        |> Nx.iota(axis: 0)
+        |> Nx.reshape({num_elements, 1}),
+        Nx.reshape(group_indices, {num_elements, 1})
+      ]
+      |> Nx.concatenate(axis: 1)
+
+    to_add = Nx.broadcast(1, {num_elements}) * (weights |> Nx.reshape({:auto}))
```
polvalente, line 379: Isn't this the same as `Nx.flatten(weights)`?

I mean, if the shapes differ anyway, the multiplication won't work as there is no `1`-sized axis in neither tensor for broadcasting, so in the end this is just an element-wise multiplication by 1

msluszniak: Ofc you're right, I will fix that

msluszniak:
> LGTM! I have only added some minor notes.
> 
> Btw, is this a good candidate for a notebook? :)

I think I can try to create a notebook with KNN as a major theme. :)

lib/scholar/neighbors/k_nearest_neighbors.ex
msluszniak: Yes, we could do that. The only thing that convinces me to stay with `:metric` and `:p` separately is that the check with NimbleOptions is simpler for this setup.

josevalim: That's a good point but I think my suggestion has a better user experience and documentation, so please let's go with that!

lib/scholar/options.ex
```diff
@@ -41,4 +41,16 @@ defmodule Scholar.Options do
...
+    if metric == :cosine or
+         (is_tuple(metric) and tuple_size(metric) == 2 and elem(metric, 0) == :minkowski and
+            is_number(elem(metric, 1)) and
+            elem(metric, 1) >= 0) do
+      {:ok, metric}
+    else
+      {:error,
+       "expected metric to be a :cosine or tuple {:minkowski, p} where p is non-negative number, got: #{inspect(metric)}"}
+    end
+  end
```
josevalim, line 55:
```suggestion
  def metric(:cosine), do: :ok
  def metric({:minkowski, p}) when is_integer(p) and p >= 0, do: :ok

  def metric(metric) do
    {:error,
     "expected metric to be a :cosine or tuple {:minkowski, p} where p is non-negative number, got: #{inspect(metric)}"}
  end
```

lib/scholar/neighbors/k_nearest_neighbors.ex
```diff
@@ -0,0 +1,408 @@
...
+        * `:distance` - weight points by the inverse of their distance. in this case, closer neighbors of
+          a query point will have a greater influence than neighbors which are further away.
+      """
+    ],
+    metric: [
+      type: {:custom, Scholar.Options, :metric, []},
+      default: {:minkowski, 2},
+      doc: ~S"""
+      Name of the metric. Possible values:
+      * `{:minkowski, p}` - Minkowski metric. By changing value of `p` parameter we can set Euclidean, Chebyshev, Manhattan or any arbitrary $L_p$ metric.
```
josevalim, line 41:
```suggestion
      Name of the metric. Possible values:

      * `{:minkowski, p}` - Minkowski metric. By changing value of `p` parameter (an integer)
        we can set Manhattan (1), Euclidean (2), Chebyshev, or any arbitrary $L_p$ metric.
```

However, isn't  "Chebyshev" infinity? How would we set it in this case?

josevalim (commented):
Two minor nits and we can ship it!

lib/scholar/neighbors/k_nearest_neighbors.ex
msluszniak: Yeah, we should be able to pass `:inf` as `:p`. In the `Scholar.Metrics.Distance` there is currently an error that computes Chebyshev for p = 0 which is wrong. I will fix this and then change it here as well.

msluszniak: And p is actually not an integer in general, just a positive float. In sklearn it is indeed an integer, but I really don't know why.

josevalim: Ah, got it. So let’s make it a float in the validations and support `:infinity`.

lib/scholar/neighbors/k_nearest_neighbors.ex
```diff
@@ -0,0 +1,409 @@
...
+
+        * `:distance` - weight points by the inverse of their distance. in this case, closer neighbors of
+          a query point will have a greater influence than neighbors which are further away.
+      """
+    ],
+    metric: [
+      type: {:custom, Scholar.Options, :metric, []},
+      default: {:minkowski, 2},
+      doc: ~S"""
+      Name of the metric. Possible values:
```
josevalim, line 40:
```suggestion
      Name of the metric. Possible values:

```

josevalim:

PR #84, opened 2023-03-01
Update args

seanmor5:
I was surprised at the `a` and `b`, I think everything else is `x` and `y`

### DIFF
--- a/lib/scholar/linear/linear_regression.ex
+++ b/lib/scholar/linear/linear_regression.ex
@@ -73,7 +73,7 @@ defmodule Scholar.Linear.LinearRegression do
     {sample_weights, opts} = Keyword.pop(opts, :sample_weights, 1.0)
     sample_weights = Nx.tensor(sample_weights)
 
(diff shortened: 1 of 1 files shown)


lib/scholar/linear/linear_regression.ex
```diff
@@ -61,7 +61,7 @@ defmodule Scholar.Linear.LinearRegression do
         >
       }
   """
-  deftransform fit(a, b, opts \\ []) do
+  deftransform fit(x, y, opts \\ []) do
```
polvalente, line 64: I think that for this specifically a and b is the more semantic choice, given we're generally talking about AX = B

seanmor5: Perhaps mathematically, but from an API perspective I would say it's surprising and also not obvious to a beginner. Tbh I had to look up the order myself when it gave me hints in Livebook

polvalente (approved):

PR #85, opened 2023-03-04
Add moments, skewness, and kurtosis

msluszniak:


lib/scholar/stats.ex
```diff
@@ -0,0 +1,196 @@
...
+  @doc """
+  Calculate the nth moment about the mean for a sample.
+
+  ## Options
+
+  #{NimbleOptions.docs(@moment_schema)}
+
+  ## Return Values
+
+    The appropriate moment along the given axis or whole tensor.
```
josevalim, line 77:
```suggestion
  The appropriate moment along the given axis or whole tensor.
```

lib/scholar/stats.ex
```diff
@@ -0,0 +1,196 @@
...
+            "If :fisher then Fisher's definition is used, if :pearson then Pearson's definition is used."
+        ]
+      ]
+
+  @moment_schema NimbleOptions.new!(moment_schema)
+  @skew_schema NimbleOptions.new!(skew_schema)
+  @kurtosis_schema NimbleOptions.new!(kurtosis_schema)
+
+  @doc """
+  Calculate the nth moment about the mean for a sample.
```
josevalim, line 69: Use third person singular voice and we typically describe the return value at the top, instead of a separate section. Please replicate the changes in other functions! :)

```suggestion
  Calculates the nth moment about the mean for a sample.
  
  It returns the appropriate moment along the given axis or whole tensor.
```

test/scholar/stats_test.exs
```diff
@@ -0,0 +1,5 @@
+defmodule Scholar.StatsTest do
+  use ExUnit.Case, async: true
+
+  doctest Scholar.Stats
```
josevalim, line 4: Can you please add tests to this module for the options and so on?

lib/scholar/stats.ex
```diff
@@ -0,0 +1,184 @@
...
+    num_samples = if opts[:axis] == nil, do: Nx.size(tensor), else: Nx.axis_size(tensor, 0)
+    moment_n(tensor, num_samples, opts)
+  end
+
+  defnp moment_n(tensor, num_samples, opts) do
+    mean = Nx.mean(tensor, axes: opts[:axis], keep_axes: opts[:keep_axis])
+
+    Nx.sum((tensor - mean) ** opts[:moment], axes: opts[:axis], keep_axes: opts[:keep_axis]) /
+      num_samples
+  end
```
josevalim, line 102: If moment is reuqired, maybe it should be a value instead of an option? Also, could we support `axes` and `keep_axes` instead? Something like this:

```suggestion
  deftransform moment(tensor, moment, opts \\ []) do
    opts = NimbleOptions.validate!(opts, @moment_schema)
    num_samples = Enum.sum(Enum.map(opts[:axes] || Nx.axes(tensor), &Nx.axis_size(tensor, &1)))
    moment_n(tensor, moment, num_samples, opts)
  end

  defnp moment_n(tensor, moment, num_samples, opts) do
    mean = Nx.mean(tensor, opts)
    Nx.sum((tensor - mean) ** moment, opts) / num_samples
  end
```

lib/scholar/stats.ex
```diff
@@ -0,0 +1,184 @@
...
+    kurtosis_n(tensor, NimbleOptions.validate!(opts, @kurtosis_schema))
+  end
+
+  defnp kurtosis_n(tensor, opts) do
+    m2 = moment(tensor, moment: 2, axis: opts[:axis], keep_axis: opts[:keep_axis])
+    m4 = moment(tensor, moment: 4, axis: opts[:axis], keep_axis: opts[:keep_axis])
+    {_, num_bits} = Nx.type(tensor)
+
+    m2_mask =
+      Nx.select(m2 == 0, Nx.Constants.nan(if num_bits < 64, do: {:f, 32}, else: {:f, 64}), m2)
```
josevalim, line 164:
```suggestion
      Nx.select(m2 == 0, Nx.Constants.nan(Nx.Type.to_floating(Nx.type(tensor)), m2)
```

lib/scholar/stats.ex
```diff
@@ -0,0 +1,184 @@
...
+
+    vals = m4 / m2_mask ** 2
+    num_samples = Nx.axis_size(tensor, 0)
```
josevalim, line 167: Is `num_samples` always on axis 0? Even with axis option?

msluszniak: No, that's a mistake.

lib/scholar/stats.ex
msluszniak: Yeah, the number of moments as a value is a better idea. We can also add multiple axes, but then we will calculates moments of moments of moments and so on. I will add information about it in doc and it seems good to me.

lib/scholar/stats.ex
```diff
@@ -0,0 +1,174 @@
...
+        f32[3]
+        [0.9794093370437622, -0.8164965510368347, 0.9220733642578125]
+      >
+  """
+
+  deftransform skew(tensor, opts \\ []) do
+    opts = NimbleOptions.validate!(opts, @skew_schema)
+
+    num_samples =
+      Enum.product(Enum.map(opts[:axes] || Nx.axes(tensor), &Nx.axis_size(tensor, &1)))
```
josevalim, line 109: Let's refactor this into a private function called `num_samples(tensor, opts)` and reuse it in the three definitions?

josevalim:

PR #86, opened 2023-03-04
Improve tests

msluszniak:
I converted examples in docs so we can use them as doctests. Also, I refactor the code of tests and add some minor improvements.


josevalim:

PR #89, opened 2023-03-10
Fix logistic regression

msluszniak:


test/scholar/linear/logistic_regression_test.exs
```diff
@@ -5,34 +5,40 @@ defmodule Scholar.Linear.LogisticRegressionTest do
...
+    y_train = Nx.squeeze(y_train, axes: [1])
+    y_test = Nx.squeeze(y_test, axes: [1])
 
     model =
       LogisticRegression.fit(x_train, y_train,
         num_classes: 2,
-        iterations: 100,
-        learning_rate: 0.0085
+        iterations: 14,
+        learning_rate: 0.01
```
msluszniak, line 15: I don't know why, but when I set here the number of iterations bigger than 14, then there is 0 division in grad.

test/scholar/linear/logistic_regression_test.exs
```diff
@@ -5,34 +5,40 @@ defmodule Scholar.Linear.LogisticRegressionTest do
...
+    y_train = Nx.squeeze(y_train, axes: [1])
+    y_test = Nx.squeeze(y_test, axes: [1])
 
     model =
       LogisticRegression.fit(x_train, y_train,
         num_classes: 3,
-        iterations: 100,
-        learning_rate: 0.0085
+        iterations: 14,
+        learning_rate: 0.01
```
msluszniak, line 31: Same here

josevalim:
@msluszniak I think the "naive" softmax implementation has underflow/overflow issues. You can try this version (from Axon):

  defn softmax(x, opts \\ []) do
    opts = keyword!(opts, axis: -1)
    axes = wrap(opts[:axis])

    # This is a scaling term designed to prevent over/under flow when x is very
    # large. Consider cases where the intermediate value e^x with large positive
    # x, e^x tends towards infinity or 0. This poisons the rest of the
    # calculation which would otherwise be normalized with the division by sum(e^x).
    # Thus we can scale by the max value in the tensor which guarantees all values
    # are smaller than 0.
    #
    # Given the expression is essentially:
    #
    # e^(x - C) / sum(e^(x - C))
    #
    # We are essentially treating the max value as a constant term, C. Thus there
    # is no need to differentiate through the max. See also: https://github.com/google/jax/pull/2260
    # for a note on performance.
    max_val = stop_grad(Nx.reduce_max(x, axes: axes, keep_axes: true))

    stable_exp =
      x
      |> Nx.subtract(max_val)
      |> Nx.exp()

    stable_exp
    |> Nx.sum(axes: axes, keep_axes: true)
    |> then(&Nx.divide(1, &1))
    |> Nx.multiply(stable_exp)
  end
```

File: lib/scholar/linear/logistic_regression.ex [hunk 3]
```diff
@@ -85,127 +83,92 @@ defmodule Scholar.Linear.LogisticRegression do
...
+      iex> model = Scholar.Linear.LogisticRegression.fit(x, y, num_classes: 2)
+      iex> Scholar.Linear.LogisticRegression.predict_proba(model, Nx.tensor([[-3.0, 5.0]]))
+      #Nx.Tensor<
+        f32[1][2]
+        [
+          [1.650987258017267e-7, 0.9999998807907104]
+        ]
+      >
+  """
+  defn predict_proba(%__MODULE__{coefficients: coeff, bias: bias}, x) do
```
josevalim, line 171: Do we use the "proba" abbreviation anywhere else? Maybe we shouldn't abbreviate it?

josevalim:
There is also a cross entropy implementation in Axon if you think it can be helpful!

msluszniak:
> @msluszniak I think the "naive" softmax implementation has underflow/overflow issues. You can try this version (from Axon):
> 
> ```elixir
>   defn softmax(x, opts \\ []) do
>     opts = keyword!(opts, axis: -1)
>     axes = wrap(opts[:axis])
> 
>     # This is a scaling term designed to prevent over/under flow when x is very
>     # large. Consider cases where the intermediate value e^x with large positive
>     # x, e^x tends towards infinity or 0. This poisons the rest of the
>     # calculation which would otherwise be normalized with the division by sum(e^x).
>     # Thus we can scale by the max value in the tensor which guarantees all values
>     # are smaller than 0.
>     #
>     # Given the expression is essentially:
>     #
>     # e^(x - C) / sum(e^(x - C))
>     #
>     # We are essentially treating the max value as a constant term, C. Thus there
>     # is no need to differentiate through the max. See also: https://github.com/google/jax/pull/2260
>     # for a note on performance.
>     max_val = stop_grad(Nx.reduce_max(x, axes: axes, keep_axes: true))
> 
>     stable_exp =
>       x
>       |> Nx.subtract(max_val)
>       |> Nx.exp()
> 
>     stable_exp
>     |> Nx.sum(axes: axes, keep_axes: true)
>     |> then(&Nx.divide(1, &1))
>     |> Nx.multiply(stable_exp)
>   end
> ```

I use also the stable implementation of softmax. The only difference is that I dont use stop_grad function.

josevalim:
Yeah, you are right. I can't see any difference. The order between division and multiplication should not matter here I think. Does `stop_grap` help with it or not? /cc @seanmor5

test/scholar/linear/logistic_regression_test.exs
msluszniak: Ok, the problem was with `Nx.log` in cross entropy. Some probabilities were really close to 0, which in grad produce 0 division. When I change to Axon `log_softmax` it works correct.

mix.exs
```diff
@@ -27,7 +27,8 @@ defmodule Scholar.MixProject do
       {:ex_doc, "~> 0.29", only: :dev, runtime: false},
       {:nx, "~> 0.5.1"},
       {:explorer, "~> 0.5.1", only: [:test, :dev]},
-      {:nimble_options, "~> 0.5.2 or ~> 1.0"}
+      {:nimble_options, "~> 0.5.2 or ~> 1.0"},
+      {:axon, "~> 0.5.1", override: true}
```
josevalim, line 31: No need to depend on Axon for now. Let's copy and paste log_softmax for now and we will figure out how to merge them later. :)

lib/scholar/linear/logistic_regression.ex
```diff
@@ -85,127 +83,99 @@ defmodule Scholar.Linear.LogisticRegression do
...
+  defnp grad_loss(coeff, bias, xs, ys) do
+    grad({coeff, bias}, fn {coeff, bias} -> loss_fn(coeff, bias, xs, ys) end)
+  end
 
-    diff = Nx.dot(x_t, prob - one_hot) / m
+  # Normalized softmax
 
-    coeff - diff * learning_rate
+  defn softmax(t) do
+    max = Nx.Defn.Kernel.stop_grad(Nx.reduce_max(t, axes: [-1], keep_axes: true))
```
josevalim, line 138: everything in Nx.Defn.Kernel is imported inside defn.

```suggestion
    max = stop_grad(Nx.reduce_max(t, axes: [-1], keep_axes: true))
```

lib/scholar/linear/logistic_regression.ex
```diff
@@ -85,127 +83,99 @@ defmodule Scholar.Linear.LogisticRegression do
...
-    prob = softmax(dot_prod)
+  defnp grad_loss(coeff, bias, xs, ys) do
+    grad({coeff, bias}, fn {coeff, bias} -> loss_fn(coeff, bias, xs, ys) end)
+  end
 
-    diff = Nx.dot(x_t, prob - one_hot) / m
+  # Normalized softmax
 
-    coeff - diff * learning_rate
+  defn softmax(t) do
```
josevalim, line 137: Should this be `defnp`?

josevalim:

lib/scholar/linear/logistic_regression.ex
```diff
@@ -85,127 +83,110 @@ defmodule Scholar.Linear.LogisticRegression do
...
-    coeff_diff = Nx.dot(x_t, diff) / m
-
-    bias_diff = Nx.sum(diff) / m
-
-    new_coeff = coeff - coeff_diff * learning_rate
-    new_bias = bias - bias_diff * learning_rate
-
-    {new_coeff, new_bias}
+  defnp loss_fn(coeff, bias, xs, ys) do
+    probs = log_softmax(Nx.dot(xs, coeff) + bias)
```
seanmor5, line 127: One thing I found in Axon is that it is much more stable to optimize the loss function which uses logits. You can avoid doing this recomputation by "caching" the logits as metadata when computing the softmax. This is what we do in Axon, and is also what they do in Keras. I think here since you have more control internally you can actually just avoid computing softmax altogether, compute the loss from logits (there is a cross_entropy_from_logits loss in Axon), and then only compute softmax in `predict_probabilities`

msluszniak: @seanmor5 you mean something like this? 
https://github.com/msluszniak/scholar/blob/add_cross_entropy_from_logits/lib/scholar/linear/logistic_regression.ex#L120

PR #91, opened 2023-03-14
Cross Entropy from logits

msluszniak:
@seanmor5 this is my proposition according to your suggestions.


josevalim:

PR #92, opened 2023-03-20
Make option :num_classes required only for task :classification in KNN

msluszniak:
Except the :num_classes correction this PR bump version of `Explorer` and exclude one slow test in the test suite.


test/test_helper.exs
```diff
@@ -1,4 +1,4 @@
-ExUnit.start()
+ExUnit.start(exclude: [:slow])
```
polvalente, line 1: There should be a separate CI task that runs these with `--only slow` I think

jonatanklosko: Btw shouldn't we be running tests with exla?

mix.exs
```diff
@@ -26,7 +26,7 @@ defmodule Scholar.MixProject do
     [
       {:ex_doc, "~> 0.29", only: :dev, runtime: false},
       {:nx, "~> 0.5.1"},
-      {:explorer, "~> 0.5.1", only: [:test, :dev]},
+      {:explorer, "~> 0.5.5", only: [:test, :dev]},
```
josevalim, line 29: Btw, in the future we want to avoid updating dependencies requirements. That's because doing so will force everyone to use latest Explorer, forcing them to update, and we should only do so if strictly necessary. If you want to update your own Explorer only, you can do `mix deps.update explorer`. :)

PR #93, opened 2023-03-21
Add Notebook about KNNs

msluszniak:
Depends on #92 and [#1131](https://github.com/elixir-nx/nx/pull/1131)

### DIFF
--- a/notebooks/k_nearest_neighbors.livemd
+++ b/notebooks/k_nearest_neighbors.livemd
@@ -3,13 +3,13 @@
 ```elixir
 Mix.install([
   {:scholar, github: "elixir-nx/scholar"},
-  {:explorer, "~> 0.5.5"},
+  {:explorer, github: "elixir-nx/explorer", override: true},
   {:exla, "~> 0.5.2"},
   {:req, "~> 0.3.6"},
   {:kino_vega_lite, "~> 0.1.8"},
   {:kino, "~> 0.9.0"},
   {:scidata, "~> 0.1.9"},
-  {:kino_explorer, "~> 0.1.2"}
+  {:kino_explorer, "~> 0.1.2", override: true}
 ])
 ```
 
@@ -677,16 +677,13 @@ Vl.new(
 The Suction Side Displacement Thickness and Frequency features have asymmetric distributions with many observations having small values. To make these distributions less skewed, we will use the logarithms of these two features.
 
 ```elixir
-freq_log = data["frequency"] |> S.to_tensor() |> Nx.log() |> Nx.as_type(:f64)
-
-suction_side_displacement_thickness_log =
-  data["suction_side_displacement_thickness"] |> S.to_tensor() |> Nx.log() |> Nx.as_type(:f64)
+data =
+  DF.mutate(data,
+    frequency_log: log(frequency),
+    suction_side_displacement_thickness_log: log(suction_side_displacement_thickness)
+  )
 
 data = DF.discard(data, ["frequency", "suction_side_displacement_thickness"])
-data = DF.put(data, :frequency_log, freq_log)
-
-data =
-  DF.put(data, :suction_side_displacement_thickness_log, suction_side_displacement_thickness_log)
 ```
 
 Scale data and convert them to tensors.
--- a/notebooks/k_nearest_neighbors.livemd
+++ b/notebooks/k_nearest_neighbors.livemd
@@ -3,13 +3,13 @@
 ```elixir
 Mix.install([
   {:scholar, github: "elixir-nx/scholar"},
-  {:explorer, github: "elixir-nx/explorer", override: true},
+  {:explorer, "~> 0.5.6"},
   {:exla, "~> 0.5.2"},
   {:req, "~> 0.3.6"},
   {:kino_vega_lite, "~> 0.1.8"},
   {:kino, "~> 0.9.0"},
   {:scidata, "~> 0.1.9"},
-  {:kino_explorer, "~> 0.1.2", override: true}
+  {:kino_explorer, "~> 0.1.2"}
 ])
 ```
 
@@ -828,6 +828,9 @@ Vl.new(
   |> Vl.mark(:text)
   |> Vl.encode(:x, value: 11)
   |> Vl.encode(:y, value: -3.5)
-  |> Vl.encode_field(:text, "filter", type: :quantitative, aggregate: :count)
+  |> Vl.encode_field(:text, "filter",
+    type: :quantitative,
+    aggregate: :count
+  )
 ])
 ```
--- a/notebooks/k_nearest_neighbors.livemd
+++ b/notebooks/k_nearest_neighbors.livemd
@@ -123,9 +123,8 @@ defmodule KFold do
     {splits_data, splits_labels, {test_data, test_labels}} =
       split_data(data, labels, num_splits, seed, train_size)
 
-    res =
-      Enum.reduce_while(0..(param_space_size * num_splits - 1), {0, []}, fn x,
-                                                                            {idx, vals} = acc ->
+    {_, res} =
+      Enum.reduce_while(0..(param_space_size * num_splits - 1), {0, []}, fn x, {idx, vals} ->
         if x < param_space_size * num_splits,
           do:
             {:cont,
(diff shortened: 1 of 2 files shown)


josevalim:
@msluszniak i have released Nx 0.5.2 with the mode fix, so you can update the notebook :)

notebooks/k_nearest_neighbors.livemd
```diff
@@ -0,0 +1,812 @@
...
+freq_log = data["frequency"] |> S.to_tensor() |> Nx.log() |> Nx.as_type(:f64)
+
+suction_side_displacement_thickness_log =
+  data["suction_side_displacement_thickness"] |> S.to_tensor() |> Nx.log() |> Nx.as_type(:f64)
+
+data = DF.discard(data, ["frequency", "suction_side_displacement_thickness"])
+data = DF.put(data, :frequency_log, freq_log)
+
+data =
+  DF.put(data, :suction_side_displacement_thickness_log, suction_side_displacement_thickness_log)
```
msluszniak, line 662: This part is not elegant, but I didn't know how to avoid casting to tensors. Both `:math.log` and `Nx.log` currently don't work inside `DF.mutate`. It will be great if `Explorer` support math operations such as log, exp, etc. Should I send an issue @josevalim?

notebooks/k_nearest_neighbors.livemd
```diff
@@ -0,0 +1,812 @@
...
+      Enum.reduce_while(0..(param_space_size * num_splits - 1), {0, []}, fn x,
+                                                                            {idx, vals} = acc ->
+        if x < param_space_size * num_splits,
+          do:
+            {:cont,
+             {idx + 1,
+              [one_round_fold(x, splits_data, splits_labels, num_splits, task, num_classes)] ++
+                vals}},
+          else: {:halt, acc}
+      end)
```
msluszniak, line 136:
```suggestion
    {_, res} =
      Enum.reduce_while(0..(param_space_size * num_splits - 1), {0, []}, fn x,
                                                                            {idx, vals} = acc ->
        if x < param_space_size * num_splits,
          do:
            {:cont,
             {idx + 1,
              [one_round_fold(x, splits_data, splits_labels, num_splits, task, num_classes)] ++
                vals}},
          else: {:halt, acc}
      end)
```
```suggestion
    res =
      Enum.reduce_while(0..(param_space_size * num_splits - 1), {0, []}, fn x,
                                                                            {idx, vals} = acc ->
        if x < param_space_size * num_splits,
          do:
            {:cont,
             {idx + 1,
              [one_round_fold(x, splits_data, splits_labels, num_splits, task, num_classes)] ++
                vals}},
          else: {:halt, vals}
      end)
```

notebooks/k_nearest_neighbors.livemd
```diff
@@ -0,0 +1,807 @@
...
+## Introduction
+
+This notebook will cover the three primary applications of K-Nearest Neighbors: classification, regression, and
+anomaly detection.
+
+We will also use a different approach for this notebook. First, we are going to implement KFold, and then we will explain
+how it works later on. So feel free to skip the implementation below for now and come back later once you grasp the concepts.
+
+```elixir
+defmodule KFold do
```
josevalim, line 43: Should we provide KFold as part of Scholar itself? I think scikit-learn ships with its own version too, right?

notebooks/k_nearest_neighbors.livemd
```diff
@@ -0,0 +1,807 @@
...
+    task = opts[:task]
+    num_classes = opts[:num_classes]
+    {_, num_features} = Nx.shape(data)
+
+    {splits_data, splits_labels, {test_data, test_labels}} =
+      split_data(data, labels, num_splits, seed, train_size)
+
+    {_, res} =
+      Enum.reduce_while(0..(param_space_size * num_splits - 1), {0, []}, fn x, {idx, vals} ->
+        if x < param_space_size * num_splits,
```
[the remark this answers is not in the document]
josevalim: This could also be implemented in pure Nx using `while`, which may be important if we move it to Scholar itself!

notebooks/k_nearest_neighbors.livemd
josevalim: Yes, we should support `log`. Please open up an issue!

notebooks/k_nearest_neighbors.livemd
```diff
@@ -0,0 +1,807 @@
...
+best_model =
+  KNearestNeighbors.fit(train_data, train_labels, num_neighbors: best_neighbors, num_classes: 3)
+
+final_prediction = KNearestNeighbors.predict(best_model, test_data)
+Scholar.Metrics.accuracy(test_labels, final_prediction)
+```
+
+Pretty well, now let's move on to a regression task using KNN.
+
+<!-- livebook:{"branch_parent_index":0} -->
```
jonatanklosko, line 505:
```suggestion
<!-- livebook:{"branch_parent_index":1} -->
```

notebooks/k_nearest_neighbors.livemd
msluszniak: It would be great if we provide this as a part of Scholar, but I think it might be a little bit problematic. Cross-validation goes through a given set of hyperparameters which very often are numbers because tensors' shapes depend on them. In this case, I also encounter the problem, because we iterate through different values of num_neighbors which must be a number. When I tried implementation with pure Nx and while I cannot implement this to pass num_neighbors as a number.

josevalim: @msluszniak I see. I think it is fine then to implement the outer loop in Elixir, as you have done here, as long as the inner of the loop is in Nx. As long as we document it and we don't name the function `fit/predict`, which is not the case in scikit anyway, i think we are fine. :)

notebooks/k_nearest_neighbors.livemd
msluszniak: I added `DF.mutate`, but currently compilation of Explorer takes a lot of time, so I can revert changes and add them with a new version of Explorer.

notebooks/k_nearest_neighbors.livemd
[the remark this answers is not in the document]
msluszniak: I changed to `Enum.reduce` and converted other functions into `defn`.

notebooks/k_nearest_neighbors.livemd
```diff
@@ -0,0 +1,833 @@
...
+    value: :gray
+  ),
+  Vl.new()
+  |> Vl.transform(filter: [param: "brush", field: "is_outlier", equal: true])
+  |> Vl.mark(:text)
+  |> Vl.encode(:x, value: 11)
+  |> Vl.encode(:y, value: -3.5)
+  |> Vl.encode_field(:text, "filter", type: :quantitative, aggregate: :count)
+])
+```
```
msluszniak, line 833: I'm wondering how to implement a plot in which there will be a counter of outliers inside a brush. Currently, the counter shows all samples inside a brush.

notebooks/k_nearest_neighbors.livemd
```diff
@@ -0,0 +1,833 @@
...
+
+Below are the KDE plots of the dataset's features. We will analyze them in the next step.
+
+```elixir
+Vl.new(
+  title: [
+    text: "KDE plots of Features",
+    offset: 20
+  ]
+)
```
msluszniak, line 589: I will simplify the code of KDE plots, but to be honest, I don't have any idea right now.

josevalim:
@msluszniak notebook is good to go! Do you want to merge it as is and then extract KFold or do you want to extract KFold and only then merge it?

msluszniak:
> @msluszniak notebook is good to go! Do you want to merge it as is and then extract KFold or do you want to extract KFold and only then merge it?

I would rather merge it as it is and then separate KFold. In the notebook, the Kfold module implements cross_validation with a splitting strategy set to KFold [cross_validation](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_validate.html#sklearn.model_selection.cross_validate), rather than just [KFold](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html). We can offer a KFold split, but I think that without a dynamic shape, it will be tough to implement general cross-validation. I also wonder about the two comments above about plotting and one earlier about the version of Explorer.

notebooks/k_nearest_neighbors.livemd
josevalim: We will ship v0.5.6 soon!

josevalim:
@msluszniak i think the current version of KFold is fine to bring into Scholar. We may also want to do the splitting on Enumerables instead of tensors, which gives more flexibility. This is also how Axon works with Axon.Loop :)

josevalim:
@msluszniak explorer 0.5.6 is out!

josevalim:

PR #99, opened 2023-04-23
Add affinity propagation

msluszniak:


test/scholar/cluster/affinity_propagation_test.exs
```diff
@@ -0,0 +1,132 @@
...
+               1,
+               3,
+               7,
+               6,
+               2,
+               2,
+               2,
+               7,
+               6,
+               2,
```
josevalim, line 82: Can you please write this by concatenating lists, something like:

    [4, 5, 6, 7, 3, 3, 5, 6] ++
      [3, 5, 10, 7, 9]

So it takes 2 or 3 lines only? :)

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,321 @@
...
+
+        temp = Nx.clip(temp, 0, inf)
+
+        temp = Nx.put_diagonal(temp, a_change)
+
+        temp = temp * (1 - damping_factor)
+
+        a = a * damping_factor - temp
+
+        diagonals = Nx.take_diagonal(a) + Nx.take_diagonal(r) > 0
```
josevalim, line 174: Since diagonal does not seem to be used throughout the loop, I think you can compute this outside of the `while` by using the last `a` and `r`?

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,321 @@
...
+          |> Nx.subtract(1)
+          |> Nx.as_type({:s, 64})
+
+        c = Nx.argmax(Nx.take(s, indices, axis: 1), axis: 1)
+        c = Nx.indexed_put(c, Nx.new_axis(indices, -1), Nx.iota({k}))
+        {Nx.take(data, indices), indices, c}
+      else
+        {n, _} = Nx.shape(s)
+
+        {Nx.tensor([-1], type: Nx.type(s)), Nx.broadcast(Nx.tensor(-1, type: Nx.type(s)), {n}),
```
josevalim, line 240: Is `indices` always meant to have size `{n}`? Or it may change?

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,321 @@
...
+          [
+            [0.0, -6062.0, -5310.0, -5459.0],
+            [-6062.0, 0.0, -966.0, -913.0],
+            [-5310.0, -966.0, 0.0, -3.0],
+            [-5459.0, -913.0, -3.0, 0.0]
+          ]),
+        cluster_centers: Nx.tensor(
+          [
+            [12, 5, 78, 2],
+            [1, 2, 5, 2]
```
josevalim, line 216: Are those values never negative?

lib/scholar/cluster/affinity_propagation.ex
msluszniak: No, I'll introduce some negative values.

lib/scholar/cluster/affinity_propagation.ex
msluszniak: No, I'll introduce some negative values.

lib/scholar/cluster/affinity_propagation.ex
msluszniak: No, it may change. The case in the else part indicates that the algorithm does not converge to any solution and we need to return some "trash".

josevalim: I see. I have a question: could we then implement this whole function as a `defn`? Then in the `if true` case, we may have some `-1` values in the indexes. Then an operation, outside of `defn`, could do the `take` only on the indices that are not negative?

msluszniak: I abstract the function compute_values from fit because I couldn't implement this part in `defn`, otherwise there wouldn't be that function. The problem with defn in this situation is that algorithm decides on its own how many clusters it predict, and that introduces dynamic shapes (that is why I use Enum.filter).

josevalim: Yes, what I propose is for us to compute and return fixed size tensors in `fit`, using only `defn`, and use `-1` indexes to say which entries are dummies. Something like this:

   %Scholar.Cluster.AffinityPropagation{
        labels: Nx.tensor([0, 1, 1, 1]),
        cluster_centers_indices: Nx.tensor([0, -1, -1, 3]),
        affinity_matrix: Nx.tensor(
          [
            [0.0, -6162.0, -5358.0, -5499.0],
            [-6162.0, 0.0, -1030.0, -913.0],
            [-5358.0, -1030.0, 0.0, -31.0],
            [-5499.0, -913.0, -31.0, 0.0]
          ]),
        cluster_centers: Nx.tensor(
          [
            [12, 5, 78, 2],
            [0, 0, 0, 0],
            [0, 0, 0, 0],
            [1, -2, 5, 2]
          ]
        )
      }

We could even have a field called `num_clusters` which returns the number of actual clusters. Then we can have a function called `prune`, which is not a `defn`, that receives the struct above and remove all negative indexes and the dummy clusters.

msluszniak: Ok, I can recreate `cluster_centers_indices` and `cluster_centers` in the proposed manner, but labels cannot be calculated inside defn.

josevalim: Where would we use the labels? They are not used in `predict` right now, are they?

msluszniak: Well, labels are the 'final product' and they are not used in `predict`. I thought I implemented an alternative version with static shapes to compute labels, but some labels are wrong (And tbh I don't know why). 

working version, but with dynamic shapes
```elixir
labels = Nx.argmax(Nx.take(s, cluster_centers_indices, axis: 1), axis: 1)
labels = Nx.indexed_put(labels, Nx.new_axis(cluster_centers_indices, -1), Nx.iota({k}))
```

my alternative version
```elixir
  c =
    Nx.select(
      Nx.broadcast(Nx.new_axis(Nx.flatten(diagonals) != 0, -1), Nx.shape(s)),
      s,
      Nx.Constants.neg_infinity(Nx.type(s))
    )

  c = Nx.argmax(c, axis: 0) |> Nx.as_type({:s, 64})
```

josevalim: What if make this in steps? What if we do this `fit |> compute_values |> predict`, but `fit` already computes the struct and `compute_values` is a regular function that simply removes the negative indexes?

josevalim: It is a good first step. We can look into improving it further next :)

msluszniak: Btw @josevalim, are the warnings that I've got after the most recent pushes correct? Because I don't think so, if I change -1 to 1 in tensors then warnings magically disappear.

polvalente: Inside defn the unary - operator becomes the negate function. I think we could get around this by using Kernel.- if the input is a number

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,378 @@
...
+  ## Examples
+
+      iex> seed = 42
+      iex> x = Nx.tensor([[12,5,78,2], [1,5,7,32], [1,3,6,1], [1,2,5,2]])
+      iex> model = Scholar.Cluster.AffinityPropagation.fit(x, seed: seed)
+      iex> model = Scholar.Cluster.AffinityPropagation.prune(model)
+      iex> Scholar.Cluster.AffinityPropagation.predict(model, Nx.tensor([[1,6,2,6], [8,3,8,2]]))
+      #Nx.Tensor<
+        s64[2]
+        [1, 1]
```
josevalim, line 321: So what is the meaning of this result? The 1 refers to positions in `x`? Can we clarify it?

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,378 @@
...
+
+    labels = Nx.argmax(Nx.take(similarity_matrix, cluster_centers_indices, axis: 1), axis: 1)
+    labels = Nx.indexed_put(labels, Nx.new_axis(cluster_centers_indices, -1), Nx.iota({k}))
+
+    %__MODULE__{
+      model
+      | cluster_centers_indices: cluster_centers_indices,
+        cluster_centers: cluster_centers,
+        labels: labels
+    }
```
josevalim, line 306: So here is what I am thinking: could we make it so the labels get `-1` and the cluster centers get NaNs in the positions where indices are `-1`? This way when we compute the euclidian distance and then discard NaNs?

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,378 @@
...
+
+    diagonals = Nx.take_diagonal(a) + Nx.take_diagonal(r) > 0
+
+    k = Nx.sum(diagonals, axes: [0])
+    {n, _} = shape = Nx.shape(data)
+
+    {cluster_centers, cluster_centers_indices, labels} =
+      if k > 0 do
+        indices =
+          Nx.select(Nx.flatten(diagonals) != 0, Nx.iota(Nx.shape(diagonals)), -1)
```
josevalim, line 197: `Nx.flatten(diagonals) != 0` is computed several times. I would extract it to its own variable. :)

lib/scholar/cluster/affinity_propagation.ex
msluszniak: This means that these two samples were labelled to the first centroid (at index 1 in clusters_centers after calling function prune)

lib/scholar/cluster/affinity_propagation.ex
msluszniak: Labels are the labels of data, so this wouldn't make sense. For cluster centers I originally pick nans, but eventually resign, because nans are available only as floats. If we fill them with 0 we can preserve the correct type of data.

msluszniak: I mean at this point we can put either all -1 as labels or try to calculate them correctly.

josevalim: The issue with filling them with `0` is that then the eucledian distance can accidentally match on them, no? Or would we do a pass to discard them later?

msluszniak: After the `prune` function all parameters are ok, so our problem is to compute labels without knowing explicitly how many clusters do we have. 

For example, suppose we have given from the algorithm that the clusters are samples on indices [2, 6, 7, 11] and num_samples = 15. Ideally, if the data to fit has the following closest centroids [2, 2, 6, 7, 11, 2 ...] we want to label them as [0, 0, 1, 2, 3, 0 ... ], but we couldn't achieve this because of static shapes. If we could somehow return labels [2, 2, 6, 7, 11, 2 ...] this will also be great and we relabel the data in `prune` to  [0, 0, 1, 2, 3, 0 ... ]. Currently, in `fit` I tried to achieve this, but some labels are mismatched.

msluszniak: > The issue with filling them with `0` is that then the eucledian distance can accidentally match on them, no? Or would we do a pass to discard them later?

Since we got the whole information in `cluster_centers_indices`, we never make a mistake in `prune` function. However, 0s are confusing because someone may think that those 0s are legit clusters with position [0, 0, ..., 0]

josevalim: Maybe we force affinity propagation to convert to floats for now?

msluszniak: Yeah, then we could use Nans, I think that's a good idea.

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,382 @@
...
+      type: {:or, [:float, :boolean, :integer]},
+      doc: "Self preference."
+    ],
+    seed: [
+      type: :integer,
+      doc: """
+      Determines random number generation for centroid initialization.
+      If the seed is not provided, it is set to `System.system_time()`.
+      """
+    ]
```
polvalente, line 51: I think it's better if we receive a `key` tensor instead.
This allows the algorithm to play nicer with other functions that also use Nx.Random

polvalente: The algorithm should also return the key -- or make it clear that the key should be manually split beforehand

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,382 @@
...
+
+        y = Nx.reduce_max(temp, axes: [1])
+
+        neg_inf = Nx.Constants.neg_infinity(Nx.Type.to_floating(Nx.type(a)))
+        inf = Nx.Constants.infinity(Nx.Type.to_floating(Nx.type(a)))
+
+        neg_infinities = Nx.broadcast(neg_inf, {n})
+
+        max_indices =
+          Nx.stack([range, indices])
```
polvalente, line 152: Shouldn't the `range` also be included in the `while` args?

Also, if you do the stack along axis: 1 you can avoid the transposition

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,382 @@
...
+
+        y2 = Nx.reduce_max(temp, axes: [1])
+
+        temp = s - Nx.new_axis(y, -1)
+
+        temp =
+          Nx.indexed_put(
+            temp,
+            max_indices,
+            Nx.flatten(Nx.gather(s, max_indices) - y2)
```
polvalente, line 165: Why is the `flatten` needed?

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,382 @@
...
+
+        temp = Nx.max(r, 0)
+
+        temp = Nx.put_diagonal(temp, Nx.take_diagonal(r))
+
+        temp = temp - Nx.sum(temp, axes: [0])
+
+        a_change = Nx.take_diagonal(temp)
+
+        temp = Nx.clip(temp, 0, inf)
```
polvalente, line 180:
```suggestion
        temp = Nx.max(temp, 0)
```

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,382 @@
...
+      end
+
+    diagonals = Nx.take_diagonal(a) + Nx.take_diagonal(r) > 0
+
+    k = Nx.sum(diagonals, axes: [0])
+    {n, _} = shape = Nx.shape(data)
+
+    {cluster_centers, cluster_centers_indices, labels} =
+      if k > 0 do
+        mask = Nx.flatten(diagonals) != 0
```
polvalente, line 198: Is the flatten necessary here?

lib/scholar/cluster/affinity_propagation.ex
polvalente: How about adding a fake cluster at the end of the cluster list?
Then pruning is just a matter of dropping that last position from the shape and all related data.

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,382 @@
...
+  defn initialize_similarities(data, opts \\ []) do
+    {n, dims} = Nx.shape(data)
+    self_preference = opts[:self_preference]
+    t1 = Nx.reshape(data, {1, n, dims})
+    t2 = Nx.reshape(data, {n, 1, dims})
+
+    dist =
+      (t1 - t2)
+      |> Nx.pow(2)
+      |> Nx.sum(axes: [2])
```
polvalente, line 364: isn't this Scholar.Metrics.Distance.euclidean_squared?

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,382 @@
...
+    fill_in =
+      case self_preference do
+        false ->
+          Nx.broadcast(Nx.median(dist), {n})
+
+        _ ->
+          if Nx.size(self_preference) == 1,
+            do: Nx.broadcast(self_preference, {n}),
+            else: self_preference
+      end
```
polvalente, line 377: this feels like a `cond`

lib/scholar/cluster/affinity_propagation.ex
josevalim: Let's keep seed for now, as this is a larger change, and we figure out a way of injecting either key or seed later?

lib/scholar/cluster/affinity_propagation.ex
msluszniak: If I broadcast t1 and t2 correctly beforehand, then yes

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,376 @@
...
+    s =
+      s +
+        normal *
+          (Nx.Constants.smallest_positive_normal(Nx.type(s)) * 100 +
+             Nx.Constants.epsilon(Nx.type(data)) / 10 * s)
+
+    range = Nx.iota({n})
+
+    {a, r, _, _, _} =
+      while {a = initial_a, r = initial_r, s = s, range = range, i = 0},
```
josevalim, line 138: I think this also works:

```suggestion
      while {a = initial_a, r = initial_r, s, range, i = 0},
```

josevalim:
@msluszniak I pushed a commit that makes pruning optional. We first compute the labels pointing only to valid clusters and then we may prune it if desired. I also changed NaNs to Infinity to make it work. Please check and if you are happy, we can merge. :)

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -0,0 +1,331 @@
...
+          |> Nx.as_type({:s, 64})
+
+        {labels, _} =
+          while {labels, j = 0}, i <- indices do
+            if i >= 0 do
+              {Nx.put_slice(labels, [i], Nx.new_axis(j, -1)), j + 1}
+            else
+              {labels, j + 1}
+            end
+          end
```
jonatanklosko, line 189: I have no idea what's going on, but indices are based off of the mask, so this should be equivalent:

```suggestion
        labels = Nx.select(mask, Nx.iota(Nx.shape(labels)), labels)
```

msluszniak:
> @msluszniak I pushed a commit that makes pruning optional. We first compute the labels pointing only to valid clusters and then we may prune it if desired. I also changed NaNs to Infinity to make it work. Please check and if you are happy, we can merge. :)

Everything LGTM!

josevalim (approved):

josevalim:

PR #104, opened 2023-05-20
Added polynomial regression with tests and livebook

matiascr:
Continuation of PR https://github.com/elixir-nx/scholar/pull/69


lib/scholar/linear/polynomial_regression.ex
```diff
@@ -0,0 +1,226 @@
...
+      #Nx.Tensor<
+        s64[3][5]
+        [
+          [0, 1, 0, 0, 1],
+          [2, 3, 4, 6, 9],
+          [4, 5, 16, 20, 25]
+        ]
+      >
+  """
+  deftransform transform(x, opts \\ []) do
```
josevalim, line 172: Why is it called `transform` instead of `predict` like in linear regression?

matiascr: `predict` is still there. `transform` only computes the feature matrix.

josevalim: I see, thanks. Why is it called `transform`? Is this a convention that comes from scikit-learn? Or should we call it `feature_matrix` or similar? Thoughts? /cc @polvalente

polvalente: yup, convention from sklearn (https://towardsdatascience.com/what-and-why-behind-fit-transform-vs-transform-in-scikit-learn-78f915cf96fe)

josevalim (approved):

polvalente (approved):

PR #105, opened 2023-05-23
Add key as an option to seed

### LINKED ISSUE #103: Pass key instead of seed to algorithms with randomization
If we want this functionality I can implement this.

msluszniak:
Closes #103

### DIFF
--- a/lib/scholar/cluster/affinity_propagation.ex
+++ b/lib/scholar/cluster/affinity_propagation.ex
@@ -41,7 +41,7 @@ defmodule Scholar.Cluster.AffinityPropagation do
       doc: "Self preference."
     ],
     seed: [
-      type: :integer,
+      type: {:custom, Scholar.Options, :seed, []},
       doc: """
       Determines random number generation for centroid initialization.
       If the seed is not provided, it is set to `System.system_time()`.
@@ -111,7 +111,8 @@ defmodule Scholar.Cluster.AffinityPropagation do
       initialize_matrices(data, self_preference: self_preference)
 
     {n, _} = Nx.shape(initial_a)
-    key = Nx.Random.key(seed)
+    seed_size = Nx.size(seed)
+    key = if seed_size == 2, do: seed, else: Nx.Random.key(seed)
     {normal, _new_key} = Nx.Random.normal(key, 0, 1, shape: {n, n}, type: Nx.type(s))
 
     s =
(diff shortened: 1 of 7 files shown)


lib/scholar/cluster/k_means.ex
```diff
@@ -73,7 +73,7 @@ defmodule Scholar.Cluster.KMeans do
       """
     ],
     seed: [
-      type: :integer,
+      type: {:custom, Scholar.Options, :seed, []},
```
polvalente, line 76: I think we should fully replace `:seed` with just accepting valid `Nx.Random.key`s

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -97,11 +97,13 @@ defmodule Scholar.Cluster.AffinityPropagation do
   deftransform fit(data, opts \\ []) do
     opts = NimbleOptions.validate!(opts, @opts_schema)
     opts = Keyword.update(opts, :self_preference, false, fn x -> x end)
-    seed = Keyword.get_lazy(opts, :seed, &System.system_time/0)
-    fit_n(data, seed, NimbleOptions.validate!(opts, @opts_schema))
+    key = Keyword.get_lazy(opts, :key, &System.system_time/0)
+    key_size = Nx.size(key)
+    key = if key_size == 2, do: key, else: Nx.Random.key(Nx.as_type(key, :s64))
```
josevalim, line 102: Maybe?

```suggestion
    key = Keyword.get_lazy(opts, :key, fn -> Nx.Random.key(System.system_time()) end)
```

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -40,11 +40,11 @@ defmodule Scholar.Cluster.AffinityPropagation do
...
       doc: "Self preference."
     ],
-    seed: [
-      type: {:custom, Scholar.Options, :seed, []},
+    key: [
+      type: {:custom, Scholar.Options, :key, []},
       doc: """
       Determines random number generation for centroid initialization.
-      If the seed is not provided, it is set to `System.system_time()`.
+      If the key is not provided, it is set to `System.system_time()`.
```
josevalim, line 47:
```suggestion
      If the key is not provided, it is set to `Nx.Random.key(System.system_time())`.
```

lib/scholar/cluster/affinity_propagation.ex
josevalim: We need to apply this diff on the other examples too :)

josevalim (approved):

josevalim:

PR #106, opened 2023-05-26
Add unrolling loops

### LINKED ISSUE #102: Optional unrolling while loops in Scholar
As proposed in #101, we might create a separate optional variable `:unroll_while` for every functionality that is implemented using the `while` loop. Then, if set to `true` we unroll the while loop. The default value would be `false`. I will implement this one if we decide.

msluszniak:
Closes #102


josevalim:

Issue #75, opened 2023-01-31
Roadmap for Scholar


msluszniak:
This issue presents the roadmap of the most important functionalities that should be implemented in the first place.

- [ ] **Clustering Algorithms** - We need to implement more clustering algorithms especially density-based e.g. *DBSCAN*
- [ ] **SVM** 
- [x] **K-Nearest Neighbors**
- [ ] **More Classification and Regression Metrics**
- [ ] **Numerical Solvers** - They won't be functionalities in Scholar per se, but they will be exceptionally helpful. Using them, we could improve implementations of current functionalities like *Logistic Regression*, *Ridge Regression* and so on. We should start with *LBFGS*
- [ ] **KFold**

seanmor5:
@msluszniak On the note of solvers, it may be useful to look at the implementations in [jaxopt](https://github.com/google/jaxopt)! JAX also supports LBFGS out of the box I believe, but the XLA version is significantly slower than the SciPy version

josevalim:
@seanmor5 do you solvers should be in a separate library then? Or here is fine?

Some additional thoughts:

 * Splitting data may also be worth moving to Nx, since it will be shared by both Scholar and Axon. So I would postpone tackling this for now

 * We most likely cannot implement training of decision trees - this poses the question if this library should implement decision trees in the first place. However, [the HummingBird project](https://github.com/microsoft/hummingbird) shows it can be efficient to compile trained decision trees into tensors

tiagodavi:
@msluszniak Is this priority list updated?

msluszniak:
> @msluszniak Is this priority list updated?

@tiagodavi I think that this list is still up to date.

cigrainger:
Is anyone working on SVM? What about saving/loading fitted models? I've got both in my near-term roadmap at work and can jump on them if they're not being worked on already.

josevalim:
@msluszniak is working on SVM. :)

PR #107, opened 2023-06-01
add train_test_split utility function

tiagodavi:
I created a simple train_test_split function based on the [roadmap](https://github.com/elixir-nx/scholar/issues/75)

I had some issues with the `doctest `related to syntax error, so I temporally kept it commented.
Screenshot 2023-05-31 at 22 34 30

I also would like some information about `NimbleOptions`, I tried to follow the other examples, but I am unsure whether or not this is correct. 

@josevalim
@msluszniak 
@polvalente

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ scholar-*.tar
 
 # Temporary files, for example, from tests.
 /tmp/
+
+# ASDF files
+.tool-versions
--- a/lib/scholar/utilities.ex
+++ b/lib/scholar/utilities.ex
@@ -1,0 +1,55 @@
+defmodule Scholar.Utilities do
+  @moduledoc """
+  Utility functions for general purposes.
+  """
+
+  import Nx.Defn
+
+  train_test_split_schema = [
+    train_size: [
+      required: true,
+      type: :float,
+      doc: """
+      Percentual size for the training set.
+      """
+    ]
+  ]
+
+  @train_test_split_schema NimbleOptions.new!(train_test_split_schema)
+
+  @doc ~S"""
+  Split tensors into random train and test subsets.
+
+  ## Options
+
+  #{NimbleOptions.docs(@train_test_split_schema)}
+
+  ## Example
+
+      Split into 80% for training and 20% for testing.
+
+      iex> {_train, _test} = Scholar.Utilities.train_test_split(Nx.tensor([[3, 6, 5], [26, 75, 3], [23, 4, 1]]), train_size: 0.8)
+      {#Nx.Tensor<
+        s64[2][3]
+        [
+          [3, 6, 5],
+          [26, 75, 3]
+        ]
+      >,
+      #Nx.Tensor<
+        s64[1][3]
+        [
+          [23, 4, 1]
+        ]
+      >}
+  """
+  deftransform train_test_split(%Nx.Tensor{} = features_tensor, opts \\ []) do
+    opts = NimbleOptions.validate!(opts, @train_test_split_schema)
+
+    features = Nx.to_list(features_tensor)
+    num_samples = Enum.count(features)
+    split_size = floor(opts[:train_size] * num_samples)
+    {train, test} = Enum.split(features, split_size)
+    {Nx.tensor(train), Nx.tensor(test)}
+  end
+end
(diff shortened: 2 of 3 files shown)


josevalim:
Thank you! I am not sure if we should merge this yet because a common question is: are you splitting Nx tensors or do you split data before they become tensors? In my experience, it is often the latter. You split dataframes, streams, etc, which means a solution would not be Nx specific. We really need to look at practical examples here. Our guides, currently, all split on Explorer itself. :)

tiagodavi:
Humm not sure either, but I know is a quite common scenario to have a list or tensor of features where you remove one of those columns, the “target” you want to predict and then want to split the remaining tensor to use part of the data for training and the rest for testing. 

I imagine this tensor of features as a single tensor where each column is a feature and I have multiple rows, this is the argument of this function, then I want 80% of it for training and 20% for testing, this is the second argument.

And sometimes I don’t want to install anything extra like Explorer, I just want to use the simplest library.

tiagodavi:
Maybe the simplest solution is receiving the list instead of tensor and let the user decide whether they want to convert it to tensor or not.

josevalim:
> And sometimes I don’t want to install anything extra like Explorer, I just want to use the simplest library.

Exactly but sometimes you want to split before you even have a tensor as well and it looks wasteful to create a tensor only to split it, rather than create tensors from previously split data. :)

tiagodavi:
How about a protocol? It can handle both cases and even more in the future. If gets a tensor split the tensor, if gets list split the list and so on ..

polvalente:
@josevalim I can see value in this applying to tensors as well.

I have an use case which I faced in reinforcement learning where I would take a collection of samples that was generated in a tensor by running simulations. 

And with that collection of samples, I pretrain a neural network as if it was a supervised learning problem before actually training with RL techniques. 

In this application, the data is fully generated into Nx tensors directly

josevalim:
FWIW, I would be totally happy with adding this as `Nx.split` too.

polvalente:
@josevalim I meant Nx.slice or slice_along_axis directly, not Access, but I'm fine with having it work on tensors directly too :)

josevalim:
@polvalente we could support containers on them as well but then it begs the question of when to stop. Should we allow `Nx.negate(containers)`? :)

tiagodavi:
Nice. I will push a PR to NX repo then with something like:

`Nx.split(tensor, split: float, axis: axis)`

not sure about the `axis: axis` option yet. 

Thank you!

tiagodavi:
What's the problem with this doctest? 

```
@doc ~S"""
  Split tensor or list into train and test subsets.

  ## Examples
    Split into 80% for training and 20% for testing.

    iex> {_train, _test} = Nx.split(Nx.tensor([[3, 6, 5], [26, 75, 3], [23, 4, 1]]), split: 0.8)
    {
      #Nx.Tensor<
        s64[2][3]
        [
          [3, 6, 5],
          [26, 75, 3]
        ]
      >,
      #Nx.Tensor<
        s64[1][3]
        [
          [23, 4, 1]
        ]
      >
    }

    iex> {_train, _test} = Nx.split([[3, 6, 5], [26, 75, 3], [23, 4, 1]], split: 0.8)
    {
      [[3, 6, 5], [26, 75, 3]],
      [[23, 4, 1]]
    }
  """
```
  
```
1) doctest Nx.split/2 (165) (Nx.DoctestTest)
     test/nx/doctest_test.exs:4
     Doctest did not compile, got: (SyntaxError) lib/nx.ex:3277:5: syntax error before: '['
           |
      3277 |     [
           |     ^
     doctest:
```

josevalim:
> I believe we could draw the line at "let's support shape manipulation Nx functions" (slice, take, stack, concatenate and friends) but not any creation or value-manipulation functions

That's a good line but be careful it may get complex in some cases where we accept options (because they can apply slightly different on each tensor based on their shape).

> Nx.split(tensor, split: float, axis: axis)

Let's go with `Nx.split(tensor, integer_or_float, opts)`.

> Re: doctests

`#Nx.Tensor` is being treated as a code comment. See how we test for this elsewhere in the library. :)

polvalente:
@josevalim what's the use case for the `integer` argument? It would be the index at which we would split?

josevalim:
Yeah.

tiagodavi:
I am closing this one in favor of this on Nx repo -> https://github.com/elixir-nx/scholar/pull/107

PR #108, opened 2023-06-07
feat: support custom optimizers in logistic regression

polvalente:


mix.exs
```diff
@@ -32,8 +32,9 @@ defmodule Scholar.MixProject do
       {:ex_doc, "~> 0.29", only: :docs},
       {:nx, github: "elixir-nx/nx", sparse: "nx", override: true},
       {:nimble_options, "~> 0.5.2 or ~> 1.0"},
-      {:exla, ">= 0.0.0", only: [:test, :dev]},
-      {:explorer, "~> 0.5.1", only: [:test, :dev]}
+      {:exla, github: "elixir-nx/exla", sparse: "exla", only: [:test, :dev]},
```
josevalim, line 35: Why this change? :)

josevalim (approved):

mix.exs
polvalente: I was getting some weird C++ raise error that went away with this update ¯\_(ツ)_/¯

PR #111, opened 2023-06-09
feat: add early-stopping criterion to logistic reg

polvalente:


lib/scholar/linear/logistic_regression.ex
```diff
@@ -156,17 +163,20 @@ defmodule Scholar.Linear.LogisticRegression do
...
-        {{coef, bias}, {x, iterations, y, coef_optimizer_state, bias_optimizer_state}}
+        has_converged = Nx.sum(Nx.abs(loss)) < Nx.size(x) * opts[:eps]
+
+        {{coef, bias},
+         {x, iterations, y, coef_optimizer_state, bias_optimizer_state, has_converged, iter + 1}}
       end
 
     %__MODULE__{
-      coefficients: Nx.transpose(final_coef),
+      coefficients: final_coef,
```
polvalente, line 173: unrelated change, but given we have Nx.dot/4, we can avoid the transpose here by changing the contraction axis below

PR #114, opened 2023-06-12
Add auc_roc, auc and roc_curve

### LINKED ISSUE #112: Scholar.Metrics exposing more metrics
hey folks, as usual great work on the lib. 

I am using it and was missing ROC-AUC, which would be awesome if the lib already provided it. 
Here is a good reference on useful metrics, formulae and meanings: https://en.wikipedia.org/wiki/Receiver_operating_characteristic

Maybe it would be nice to have all these metrics on scholar? 

thanks

msluszniak:
Closes #112 

I implemented auc_roc_socore, roc_curve and auc functions. Unfortunately, for first two I needed separate regular def function because sub-resultant tensor has unknown shape. Probably, after I implement `Nx.diff` in Nx I will add some other changes.


lib/scholar/metrics.ex
```diff
@@ -481,6 +481,191 @@ defmodule Scholar.Metrics do
...
+  @doc ~S"""
+  Compute Receiver operating characteristic (ROC).
+
+  Note: this implementation is restricted to the binary classification task.
+
+  ## Examples
+
+      iex> y_true = Nx.tensor([0, 0, 1, 1])
+      iex> scores = Nx.tensor([0.1, 0.4, 0.35, 0.8])
+      iex> distinct_value_indices = Scholar.Metrics.calculate_distinct_value_indices(scores)
```
josevalim, line 582: Ideally we would figure out how to implement this with `defn`, I can play with this later. :)

lib/scholar/metrics.ex
```diff
@@ -481,6 +481,191 @@ defmodule Scholar.Metrics do
...
+      iex> {fpr, tpr, _thresholds} = Scholar.Metrics.roc_curve(y, pred, distinct_value_indices)
+      iex> Scholar.Metrics.auc(fpr, tpr)
+      #Nx.Tensor<
+        f32
+        0.75
+      >
+  """
+  defn auc(x, y) do
+    check_shape(x, y)
+    dx = x[[1..-1//1]] - x[[0..-2//1]]
```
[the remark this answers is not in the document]
msluszniak: I don't understand why this will be beneficial

lib/scholar/metrics.ex
```diff
@@ -481,6 +481,191 @@ defmodule Scholar.Metrics do
...
+        Nx.all(dx >= 0) -> 1
+        true -> 0
+      end
+
+    direction * trapz1d(y, x, use_x?: true)
+  end
+
+  # TODO add support for multi-dimensional x and y and move to Nx
+  defnp trapz1d(y, x, opts \\ []) do
+    opts = keyword!(opts, use_x?: false, dx: 1.0, axis: -1)
```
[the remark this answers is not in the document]
msluszniak: This is an artefact from the attempt to implement more general version of this function. I'll change this.

lib/scholar/metrics.ex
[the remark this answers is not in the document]
polvalente: It allows for users to just pass the dataset in whatever order they have it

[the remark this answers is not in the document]
polvalente: Makes sense. I'd return `:nan` in the error case instead of `0`, then

lib/scholar/metrics.ex
polvalente: This is a more general question in Nx. If you can solve this somehow, maybe the solution can be used to internally handle dynamic shapes.

josevalim: We may be able to solve this in an algorithmic way, keeping the shape full, and removing the zeros on the computation itself.

lib/scholar/metrics.ex
```diff
@@ -481,6 +481,190 @@ defmodule Scholar.Metrics do
...
+
+    Nx.sum(d * (y[[0..-2//1]] + y[[1..-1//1]]) / 2.0)
+  end
+
+  @doc ~S"""
+  It's a helper function for `Scholar.Metrics.roc_curve` and `Scholar.Metrics.roc_auc_score` functions.
+  You should call it and use as follows:
+
+    `iex> distinct_value_indices = Scholar.Metrics.calculate_distinct_value_indices(scores)`
+    `iex> {fpr, tpr, thresholds} = Scholar.Metrics.roc_curve(y_true, scores, distinct_value_indices, weights)`
```
polvalente, line 535:
```suggestion
    distinct_value_indices = Scholar.Metrics.calculate_distinct_value_indices(scores)
    {fpr, tpr, thresholds} = Scholar.Metrics.roc_curve(y_true, scores, distinct_value_indices, weights)
```

lib/scholar/metrics.ex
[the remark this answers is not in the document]
msluszniak: I leave this until we have [Nx.trapz](https://github.com/elixir-nx/nx/pull/1255)

lib/scholar/metrics.ex
```diff
@@ -481,6 +481,189 @@ defmodule Scholar.Metrics do
...
+      Nx.concatenate([distinct_value_indices, Nx.new_axis(Nx.size(y_true) - 1, -1)], axis: 0)
+
+    tps = Nx.take(Nx.cumulative_sum(y_true * weight, axis: 0), threshold_idxs)
+
+    fps = Nx.take(Nx.cumulative_sum((1 - y_true) * weight, axis: 0), threshold_idxs)
+
+    {fps, tps, Nx.take(y_score, threshold_idxs)}
+  end
+
+  # TODO implement :drop_intermediate option when function Nx.diff is implemented
```
msluszniak, line 570:
```suggestion
```
This would introduce yet another tensor shape change in the middle of a function

msluszniak:
```suggestion
  # TODO implement :drop_intermediate option when dynamic shapes will be available
```

lib/scholar/integrate/trapezoidal.ex
```diff
@@ -0,0 +1,182 @@
+defmodule Scholar.Integrate.Trapezoidal do
```
josevalim, line 1: Do we want to have one module per integration or a single `Scholar.Integrate` module with all integration methods?

lib/scholar/metrics.ex
```diff
@@ -481,6 +482,176 @@ defmodule Scholar.Metrics do
...
+  end
+
+  @doc ~S"""
+  It's a helper function for `Scholar.Metrics.roc_curve` and `Scholar.Metrics.roc_auc_score` functions.
+  You should call it and use as follows:
+
+      distinct_value_indices = Scholar.Metrics.calculate_distinct_value_indices(scores)
+      {fpr, tpr, thresholds} = Scholar.Metrics.roc_curve(y_true, scores, distinct_value_indices, weights)
+  """
+  def calculate_distinct_value_indices(y_score) do
```
josevalim, line 524: Let's rename this function to `distinct_value_indices` (no need to `calculate`). I have also noticed that none of our tests have duplicate values, so this behaviour itself is not being tested.

lib/scholar/metrics.ex
```diff
@@ -481,6 +482,176 @@ defmodule Scholar.Metrics do
...
+
+    distinct_value_indices_mask = Nx.not_equal(y_score[[1..-1//1]], y_score[[0..-2//1]])
+
+    Nx.iota({Nx.size(y_score) - 1})
+    |> Nx.add(1)
+    |> Nx.multiply(distinct_value_indices_mask)
+    |> Nx.to_flat_list()
+    |> Enum.filter(fn x -> x != 0 end)
+    |> Nx.tensor()
+    |> Nx.subtract(1)
```
josevalim, line 536:
```suggestion
    |> Nx.multiply(distinct_value_indices_mask)
    |> Nx.subtract(-1)
    |> Nx.to_flat_list()
    |> Enum.filter(fn x -> x != -1 end)
    |> Nx.tensor()
```

lib/scholar/integrate/trapezoidal.ex
polvalente: I believe it's best to have one module with the different methods. We could go so far as to have a single function that delegates to different implementations based on an option

lib/scholar/metrics.ex
josevalim: I play with this a little and the only idea I had so far was to return the value of the previous index (instead of discarding them), but I think doing this would be incorrect on the trapeizodal function.

lib/scholar/integrate/trapezoidal.ex
polvalente: In general, the choice for one integration method over the other depends on the type of function you're integrating.

For instance, in circuit simulators it is common to have a drop-down menu in which you choose  the integration method. Then, if one gives artifacts, another might work better.

lib/scholar/metrics.ex
polvalente: Shouldn't it be subtract 1?

lib/scholar/integrate/trapezoidal.ex
msluszniak: I almost implemented the Simpson method in a separate file.

polvalente: I'm ok with having private separate files for them, but I really think that it would be best to have the public interface more tightly bound

lib/scholar/metrics.ex
josevalim: Sorry, fixed.

msluszniak: I don't get this change

polvalente: I believe the idea is that if we keep the operations close together we maximize the chance of operation fusion at compile-time

msluszniak: Now if first index in mask was set to `1` it will be filtered, but it shouldn't be

josevalim: We are filtering by -1 now. so if the mask is 1, we won't filter it. :)

msluszniak: First element of iota is 0, 0 * 1 if is 0, and then when you subtract it will give -1, so you fillter it, but it shouldn't be filltered. To problem here is caused by the 0 from iota

josevalim: But don’t we do Nx.add(1)? I am struggling to see for which inputs the results would differ. :(

msluszniak: Ok sorry, I completely forget about that add(1)

lib/scholar/metrics.ex
msluszniak: Yeah, the only way we can deal with the dynamic shapes currently is to preserve the original shapes and fill the rest of the indices with some "filling'. But in 95% of cases, this filling is unwanted so we need either way a filtering function at the end of the pipeline

josevalim (approved):

josevalim:

PR #116, opened 2023-06-17
Add simpson rule

msluszniak:


lib/scholar/integrate/simpson.ex
```diff
@@ -0,0 +1,446 @@
+defmodule Scholar.Integrate.Simpson do
```
josevalim, line 1: LGTM! We can keep the `Scholar.Integrate.Simpson` private (with `@moduledoc false`) and then we move the `deftransform simplson` to `Scholar.Integrate` and have it call `Scholar.Integrate.Simpson.simpson_n`.

msluszniak: Yeah, that's really good idea to me

lib/scholar/integrate/simpson.ex
```diff
@@ -0,0 +1,446 @@
...
+             else:
+               Nx.slice_along_axis(
+                 diff,
+                 start + 1,
+                 len,
+                 axis: axis,
+                 strides: 2
+               )
+
+        {diff_0, diff_1}
```
polvalente, line 357:
```suggestion
```

lib/scholar/integrate/simpson.ex
msluszniak: I think that now, we can move both Simpson and trapezoidal to `integrate.ex` and have general interface `Scholar.Integrate.integrate` and more specific calls like `Scholar.Integrate.simpson`. This will save all the documentation.

lib/scholar/integrate/integrate.ex
```diff
@@ -0,0 +1,61 @@
...
+        [2.0, 8.0]
+      >
+
+      iex> Scholar.Integrate.integrate(Nx.tensor([[0, 1, 2], [3, 4, 5]]), :simpson, Nx.tensor([[1, 1, 1], [2, 2, 2]]), axis: 0)
+      #Nx.Tensor<
+        f32[3]
+        [1.5, 2.5, 3.5]
+      >
+  """
+  deftransform integrate(y, method, x, opts \\ []) do
```
josevalim, line 36: I would have the `method` as first argument, given it is the most important (and the most different argument).

lib/scholar/integrate/integrate.ex
```diff
@@ -0,0 +1,61 @@
...
+    end
+  end
+
+  deftransform integrate_uniform(y, method, opts \\ []) do
+    case method do
+      :trapezoidal ->
+        Scholar.Integrate.Trapezoidal.trapezoidal_uniform(y, opts)
+
+      :simpson ->
+        Scholar.Integrate.Simpson.simpson(y, opts)
```
josevalim, line 55:
```suggestion
        Scholar.Integrate.Simpson.simpson_uniform(y, opts)
```

josevalim:
@polvalente are you happy with this "both worlds" approach?

lib/scholar/integrate/integrate.ex
polvalente: Having `y` as the first argument lends the function more easily to pipe-chains. I would have the method as the last mandatory argument to keep all data arguments together, though

josevalim: I would prefer `(method, y, x)` but `(y, x, method)` works for me too.

polvalente: In isolation `(method, y, x)` does look better, but it makes for clunky pipes with the need for `then` if you're piping `y` into `integrate`

lib/scholar/integrate/integrate.ex
```diff
@@ -49,13 +104,529 @@ defmodule Scholar.Integrate do
...
+        trapezoidal_uniform(y, opts)
 
       :simpson ->
-        Scholar.Integrate.Simpson.simpson_uniform(y, opts)
+        simpson_uniform(y, opts)
 
       _ ->
         raise ArgumentError, "Unknown integration method: #{inspect(method)}"
     end
   end
```
josevalim, line 115: If we are having the functions in this module, I no longer think we need integrate, because you can get the same behaviour by doing:

    apply(Scholar.Integrate, :trapezoidal, [y, x])

Thoughts?

msluszniak: Yes, valid point. Now, `Scholar.Integrate.integrate/integrate_uniform` is a little bit redundant.

polvalente: Same issue I have with pipes. Needing apply will imply needing then and captures.

IMO we should have only integrate and integrate_uniform

josevalim: @polvalente I am really not in favor of arguing for the existence of functions for "pipe" reasons. A function needs to exist by itself. It is very easy to wrap apply into something pipeable.

lib/scholar/integrate/integrate.ex
```diff
@@ -0,0 +1,632 @@
...
+      :trapezoidal ->
+        trapezoidal_uniform(y, opts)
+
+      :simpson ->
+        simpson_uniform(y, opts)
+
+      _ ->
+        raise ArgumentError, "Unknown integration method: #{inspect(method)}"
+    end
+  end
```
josevalim, line 115:
```suggestion
```

josevalim:

lib/scholar/integrate/integrate.ex
polvalente: pipe usability IMO should be considered because it is one major use case of these functions. Needing apply will force people away from piping into these functions

PR #117, opened 2023-06-20
Prepare notebooks to release

msluszniak:
After the new release we need to change `Mix.install` in all the notebooks.


josevalim (approved):
LGTM but let's update Kino and Explorer too, WDYT?

PR #126, opened 2023-06-28
Upgrade confusion matrix function

msluszniak:
This adds some cool features, but has a small drawback. When user doesn't set weights the output will have float type instead of unsigned integer. But imo, it's still worth to add it.

### DIFF
--- a/lib/scholar/metrics.ex
+++ b/lib/scholar/metrics.ex
@@ -46,7 +46,26 @@ defmodule Scholar.Metrics do
         ]
       ]
 
+  confusion_matrix_schema =
+    general_schema ++
+      [
+        sample_weights: [
+          type: {:custom, Scholar.Options, :weights, []},
+          doc: """
+          Sample weights of the observations.
+          """
+        ],
+        normalize: [
+          type: {:in, [true, :predicted, :all]},
+          doc: """
+          Normalizes confusion matrix over the `:true` (rows), `:predicted` (columns)
+          conditions or `:all` the population. If `nil`, confusion matrix will not be normalized.
+          """
+        ]
+      ]
+
   @general_schema NimbleOptions.new!(general_schema)
+  @confusion_matrix_schema NimbleOptions.new!(confusion_matrix_schema)
   @f1_score_schema NimbleOptions.new!(f1_score_schema)
 
   # Standard Metrics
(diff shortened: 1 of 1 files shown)


lib/scholar/metrics.ex
```diff
@@ -332,28 +351,59 @@ defmodule Scholar.Metrics do
...
   deftransform confusion_matrix(y_true, y_pred, opts \\ []) do
-    confusion_matrix_n(y_true, y_pred, NimbleOptions.validate!(opts, @general_schema))
+    confusion_matrix_n(y_true, y_pred, NimbleOptions.validate!(opts, @confusion_matrix_schema))
   end
 
   defnp confusion_matrix_n(y_true, y_pred, opts) do
     check_shape(y_pred, y_true)
+    weights = validate_weights(opts[:sample_weights], Nx.axis_size(y_true, 0))
+    weights = if Nx.rank(weights) == 0, do: Nx.broadcast(weights, y_true), else: weights
+    non_uniform_weights = Nx.any(Nx.diff(weights) != 0.0)
```
msluszniak, line 383:
```suggestion
```

lib/scholar/metrics.ex
```diff
@@ -332,28 +351,59 @@ defmodule Scholar.Metrics do
...
 
     num_classes = check_num_classes(opts[:num_classes])
 
     zeros = Nx.broadcast(Nx.u64(0), {num_classes, num_classes})
     indices = Nx.stack([y_true, y_pred], axis: 1)
     updates = Nx.broadcast(Nx.u64(1), y_true)
+    updates = if non_uniform_weights, do: updates * weights, else: Nx.as_type(updates, :u64)
```
msluszniak, line 390:
```suggestion
    updates = updates * weights
```

lib/scholar/metrics.ex
```diff
@@ -332,28 +351,58 @@ defmodule Scholar.Metrics do
...
   """
   deftransform confusion_matrix(y_true, y_pred, opts \\ []) do
-    confusion_matrix_n(y_true, y_pred, NimbleOptions.validate!(opts, @general_schema))
+    confusion_matrix_n(y_true, y_pred, NimbleOptions.validate!(opts, @confusion_matrix_schema))
   end
 
   defnp confusion_matrix_n(y_true, y_pred, opts) do
     check_shape(y_pred, y_true)
+    weights = validate_weights(opts[:sample_weights], Nx.axis_size(y_true, 0))
+    weights = if Nx.rank(weights) == 0, do: Nx.broadcast(weights, y_true), else: weights
```
josevalim, line 382: Validate weights already does this broadcast, no?

lib/scholar/metrics.ex
```diff
@@ -332,28 +351,58 @@ defmodule Scholar.Metrics do
...
-          [0, 0, 2]
+          [0.6666666865348816, 1.0, 0.0],
+          [0.3333333432674408, 0.0, 0.1304347813129425],
+          [0.0, 0.0, 0.8695651888847351]
         ]
       >
   """
   deftransform confusion_matrix(y_true, y_pred, opts \\ []) do
-    confusion_matrix_n(y_true, y_pred, NimbleOptions.validate!(opts, @general_schema))
+    confusion_matrix_n(y_true, y_pred, NimbleOptions.validate!(opts, @confusion_matrix_schema))
```
josevalim, line 376: If you want, you can make weights default to an integer if both inputs are integers, is that what you want?

lib/scholar/metrics.ex
msluszniak: No, since validate weights return scalar 1.0 if weights are initially `nil`

lib/scholar/metrics.ex
msluszniak: Ideally, I would make floats always if weights are passed (but this is impossible). So yes, if both data and weights are integers, then I want them to preserve the type.

msluszniak: Oh, but actually we can do this in deftransform

lib/scholar/metrics.ex
msluszniak: https://github.com/elixir-nx/scholar/blob/main/lib/scholar/shared.ex#L54-L56

josevalim: Exactly. Instead of having it as `nil`, check the input types in `deftransform` and make it so the default value is either `1` or `1.0`. This way you get the integer/float behaviour and you can also remove the broadcast. :)

josevalim:

PR #131, opened 2023-07-05
Allow broadcast in distances

msluszniak:
@seanmor5, let me know if this is sufficient in terms of distances. I will focus on classification metrics tomorrow.

### DIFF
--- a/lib/scholar/metrics/distance.ex
+++ b/lib/scholar/metrics/distance.ex
@@ -63,7 +63,7 @@ defmodule Scholar.Metrics.Distance do
       iex> x = Nx.tensor([1, 2])
       iex> y = Nx.tensor([1, 2, 3])
       iex> Scholar.Metrics.Distance.euclidean(x, y)
-      ** (ArgumentError) cannot broadcast tensor of dimensions {2} to {3}
+      ** (ArgumentError) Tensors must be broadcast compatible, got tensors with shapes {2} and {3}
 
       iex> x = Nx.tensor([[1, 2, 5], [3, 4, 3]])
       iex> y = Nx.tensor([[8, 3, 1], [2, 5, 2]])
@@ -126,7 +126,7 @@ defmodule Scholar.Metrics.Distance do
       iex> x = Nx.tensor([1, 2])
       iex> y = Nx.tensor([1, 2, 3])
       iex> Scholar.Metrics.Distance.squared_euclidean(x, y)
-      ** (ArgumentError) cannot broadcast tensor of dimensions {2} to {3}
+      ** (ArgumentError) Tensors must be broadcast compatible, got tensors with shapes {2} and {3}
 
       iex> x = Nx.tensor([[1, 2, 5], [3, 4, 3]])
       iex> y = Nx.tensor([[8, 3, 1], [2, 5, 2]])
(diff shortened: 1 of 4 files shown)


seanmor5:
@msluszniak this worked for what I needed! Thank you!

lib/scholar/metrics/distance.ex
```diff
@@ -468,20 +521,29 @@ defmodule Scholar.Metrics.Distance do
...
   defnp hamming_unweighted(x, y, opts) do
-    assert_same_shape!(x, y)
+    valid_broadcast?(Nx.rank(x), Nx.shape(x), Nx.shape(y))
     result_type = Nx.Type.to_floating(Nx.Type.merge(Nx.type(x), Nx.type(y)))
     Nx.mean(x != y, axes: opts[:axes]) |> Nx.as_type(result_type)
   end
 
   defnp hamming_weighted(x, y, w, opts) do
-    assert_same_shape!(x, y)
+    valid_broadcast?(Nx.rank(x), Nx.shape(x), Nx.shape(y))
```
josevalim, line 546: Perhaps we should instead check if they have the same `rank`? And then, if broadcast fails, it will naturally fail in a later operation?

Otherwise, if we want to keep the current code, it doesn't currently raise if the broadcast shape do not match, so this is not checking much (and it also fails with an unrelated exception if they don't have the same rank).

msluszniak: I'll add a check on rank and raise an error in both cases in `valid_broadcast?`

msluszniak: I added some changes, WDYT @josevalim?

lib/scholar/shared.ex
```diff
@@ -66,4 +66,27 @@ defmodule Scholar.Shared do
               "invalid value for :weights option: expected list or tensor of positive numbers of size #{num_samples}, got: #{inspect(weights)}"
     end
   end
+
+  deftransform valid_broadcast?(n_dims, shape1, shape2) do
```
josevalim, line 70: I think the function should be called `validate_broadcast!`, as its main goal is to raise if something is wrong, rather than question mark.

lib/scholar/shared.ex
```diff
@@ -66,4 +66,27 @@ defmodule Scholar.Shared do
...
+    valid_broadcast(Enum.to_list(0..(n_dims - 1)), shape1, shape2)
+  end
+
+  deftransform valid_broadcast([head | tail], shape1, shape2) do
+    dim1 = elem(shape1, head)
+    dim2 = elem(shape2, head)
+
+    if not (dim1 == 1 or dim2 == 1 or dim2 == dim1) do
+      raise ArgumentError,
+            "Tensors must be broadcast compatible, got tensors with shapes #{inspect(shape1)} and #{inspect(shape2)}"
```
josevalim, line 85:
```suggestion
            "tensors must be broadcast compatible, got tensors with shapes #{inspect(shape1)} and #{inspect(shape2)}"
```

lib/scholar/shared.ex
```diff
@@ -66,4 +66,27 @@ defmodule Scholar.Shared do
...
     end
   end
+
+  deftransform valid_broadcast?(n_dims, shape1, shape2) do
+    if tuple_size(shape1) != tuple_size(shape2) do
+      raise ArgumentError,
+            "expected shapes to have same rank, got #{inspect(tuple_size(shape1))} and #{inspect(tuple_size(shape2))}"
+    end
+
+    valid_broadcast(Enum.to_list(0..(n_dims - 1)), shape1, shape2)
```
josevalim, line 76: Instead of creating a list, start with `n_dims` and decrement one while `n_dims >= 0`.

msluszniak: Yeah, now I'm wondering what I was thinking when I write this function using list 😅

josevalim:

PR #137, opened 2023-07-13
Adding Gaussian Mixture Model

krstopro:
Added Gaussian Mixture Model implementation, similar to the one in scikit-learn Python module. Currently it supports only full covariances and uses k-means for the initialisation. Other covariance types (diagonal, tied, spherical) and initialisation methods (random, k-means++) can easily be added.

### DIFF
--- a/lib/scholar/cluster/gmm.ex
+++ b/lib/scholar/cluster/gmm.ex
@@ -9,7 +9,8 @@ defmodule Scholar.Cluster.GaussianMixture do
   expectation of the Gaussian assignment for each data point x and the M-step which updates the
   parameters to maximize the expectations found in E-step. While every iteration of the algorithm
   is guaranteed to improve the log-likelihood, the final result depends on the initial values of
-  the parameters and the entire procedure should be repeated several times.
+  the parameters. Thus the procedure consists of repeating the algorithm several times
+  and taking the best obtained result.
 
   References:
 
--- a/lib/scholar/cluster/gmm.ex
+++ b/lib/scholar/cluster/gmm.ex
@@ -2,7 +2,7 @@ defmodule Scholar.Cluster.GaussianMixture do
   @moduledoc """
   Gaussian Mixture Model.
 
-  Gaussian Mixture model is a probabilistic model that assumes every data point is generated
+  Gaussian Mixture Model is a probabilistic model that assumes every data point is generated
   by randomly choosing one of the several fixed Gaussian distributions and then sampling from it.
   Its parameters are estimated using the Expectation-Maximization (EM) algorithm, which is an
   iterative algorithm alternating between the two steps: the E-step which computes the
--- a/lib/scholar/cluster/gmm.ex
+++ b/lib/scholar/cluster/gmm.ex
@@ -74,37 +74,26 @@ defmodule Scholar.Cluster.GaussianMixture do
 
     The function returns a struct with the following parameters:
 
-    * `:weights` - ..
+    * `:weights` - The fractions of data sampled from each Gaussian, respectively.
 
     * `:means` - Means of the Gaussian components.
 
     * `:covariances` - Covariance matrices of the Gaussian components.
 
     * `:precisions_cholesky` - Cholesky decomposition of the precision matrices
-  	(inverse of covariance). This is useful for the inference.
+  	(inverses of covariances). This is useful for the model inference.
 
   ## Examples
 
       iex> key = Nx.Random.key(12)
-      iex> x = Nx.tensor([[1, 2], [2, 4], [1, 3], [2, 5]])
-      iex> Scholar.Cluster.KMeans.fit(x, num_clusters: 2, key: key)
-      %Scholar.Cluster.KMeans{
-        num_iterations: Nx.tensor(
-          2
-        ),
-        clusters: Nx.tensor(
-          [
-            [1.0, 2.5],
-            [2.0, 4.5]
-          ]
-        ),
-        inertia: Nx.tensor(
-          1.0
-        ),
-        labels: Nx.tensor(
-          [0, 1, 0, 1]
-        )
-      }
+      iex> x = Nx.tensor([[1, 2], [1, 4], [1, 0], [10, 2], [10, 4], [10, 0]])
+      iex> Scholar.Cluster.GaussianMixture.fit(x, num_gaussians: 2, key: key).means
+      Nx.tensor(
+        [
+					[1.0, 2.0],
+					[10.0, 2.0]
+				]
+      )
   """
   deftransform fit(x, opts \\ []) do
     if Nx.rank(x) != 2 do
@@ -198,14 +187,14 @@ defmodule Scholar.Cluster.GaussianMixture do
         }
       end
 
-    # %__MODULE__{
-    #   weights: weights,
-    #   means: means,
-    #   covariances: covariances,
-    #   precisions_cholesky: precisions_cholesky
-    # }
+    {weights, means, covariances, precisions_cholesky} = best_params
 
-    best_params
+    %__MODULE__{
+      weights: weights,
+      means: means,
+      covariances: covariances,
+      precisions_cholesky: precisions_cholesky
+    }
   end
 
   defnp init_params(key, x, opts) do
@@ -214,7 +203,6 @@ defmodule Scholar.Cluster.GaussianMixture do
     num_gaussians = opts[:num_gaussians]
     reg_covar = opts[:reg_covar]
 
-    # k_means_model = Scholar.Cluster.KMeans.fit(x, num_clusters: num_gaussians, num_runs: 1, key: key)
     # add key here
     k_means_model = Scholar.Cluster.KMeans.fit(x, num_clusters: num_gaussians, num_runs: 1)
     labels = k_means_model.labels
@@ -241,7 +229,6 @@ defmodule Scholar.Cluster.GaussianMixture do
   end
 
   defnp estimate_gaussian_parameters(x, resp, reg_covar) do
-    # consider adding a small number here as in scikit_learn
     nk = Nx.sum(resp, axes: [0])
     means = Nx.dot(Nx.transpose(resp), x) / nk
     covariances = estimate_gaussian_covariances(x, resp, nk, means, reg_covar)
@@ -304,13 +291,7 @@ defmodule Scholar.Cluster.GaussianMixture do
 
   defnp estimate_log_prob_resp(x, weights, means, precisions_cholesky) do
     weighted_log_prob = estimate_weighted_log_prob(x, weights, means, precisions_cholesky)
-
-    log_prob_norm =
-      weighted_log_prob
-      |> Nx.exp()
-      |> Nx.sum(axes: [1])
-      |> Nx.log()
-
+    log_prob_norm = logsumexp(weighted_log_prob)
     log_resp = weighted_log_prob - Nx.new_axis(log_prob_norm, 1)
     {log_prob_norm, log_resp}
   end
@@ -336,7 +317,7 @@ defmodule Scholar.Cluster.GaussianMixture do
         mean = means[k]
         precision_cholesky = precisions_cholesky[k]
         y = Nx.dot(x, precision_cholesky) - Nx.dot(mean, precision_cholesky)
-        slice = Nx.new_axis(Nx.sum(Nx.multiply(y, y), axes: [1]), 0)
+        slice = Nx.new_axis(Nx.sum(y * y, axes: [1]), 0)
 
         {
           Nx.put_slice(log_prob, [k, 0], slice),
@@ -345,11 +326,16 @@ defmodule Scholar.Cluster.GaussianMixture do
       end
 
     -0.5 * (num_features * Nx.log(2 * Nx.Constants.pi()) + Nx.transpose(log_prob)) + log_det
-    # log_prob
-    # 	|> Nx.transpose
-    # 	|> Nx.add(num_features * Nx.log(2 * Nx.Constants.pi))
-    # 	|> Nx.multiply(-0.5)
-    # 	|> Nx.add(log_det)
+  end
+
+  defnp logsumexp(weighted_log_prob) do
+    max = Nx.reduce_max(weighted_log_prob, axes: [1])
+    weighted_log_prob
+    |> Nx.subtract(Nx.new_axis(max, 1))
+    |> Nx.exp
+    |> Nx.sum(axes: [1])
+    |> Nx.log
+    |> Nx.add(max)
   end
 
   defnp compute_log_det_cholesky(precisions_cholesky) do
@@ -364,14 +350,14 @@ defmodule Scholar.Cluster.GaussianMixture do
 
   ## Return Values
 
-    It returns a tensor with clusters corresponding to the input.
+    It returns a tensor with Gaussian assignments for every input point.
 
   ## Examples
 
-      iex> key = Nx.Random.key(42)
-      iex> x = Nx.tensor([[1, 2], [2, 4], [1, 3], [2, 5]])
-      iex> model = Scholar.Cluster.KMeans.fit(x, num_clusters: 2, key: key)
-      iex> Scholar.Cluster.KMeans.predict(model, Nx.tensor([[1.9, 4.3], [1.1, 2.0]]))
+      iex> key = Nx.Random.key(12)
+      iex> x = Nx.tensor([[1, 2], [1, 4], [1, 0], [10, 2], [10, 4], [10, 0]])
+      iex> model = Scholar.Cluster.GaussianMixture.fit(x, num_gaussians: 2, key: key)
+      iex> Scholar.Cluster.GaussianMixture.predict(model, Nx.tensor([[8, 1], [2, 3]]))
       Nx.tensor(
         [1, 0]
       )
@@ -381,17 +367,35 @@ defmodule Scholar.Cluster.GaussianMixture do
          x
        ) do
     assert_same_shape!(x[0], means[0])
-    estimate_weighted_log_prob(x, weights, means, precisions_cholesky) |> Nx.argmax(axis: 1)
+    estimate_weighted_log_prob(to_float(x), weights, means, precisions_cholesky) |> Nx.argmax(axis: 1)
   end
 
   @doc """
+  Makes predictions with the given `model` on inputs `x`.
+
+  ## Return Values
+
+    It returns a tensor probabilities of Gaussian assignments for every input point.
+
+  ## Examples
+
+      iex> key = Nx.Random.key(12)
+      iex> x = Nx.tensor([[1, 2], [1, 4], [1, 0], [10, 2], [10, 4], [10, 0]])
+      iex> model = Scholar.Cluster.GaussianMixture.fit(x, num_gaussians: 2, key: key)
+      iex> Scholar.Cluster.GaussianMixture.predict_prob(model, Nx.tensor([[8, 1], [2, 3]]))
+      Nx.tensor(
+        [
+          [0.0, 1.0],
+          [1.0, 0.0]
+        ]
+      )
   """
   defn predict_prob(
          %__MODULE__{weights: weights, means: means, precisions_cholesky: precisions_cholesky},
          x
        ) do
     assert_same_shape!(x[0], means[0])
-    {_, log_resp} = estimate_log_prob_resp(x, weights, means, precisions_cholesky)
+    {_, log_resp} = estimate_log_prob_resp(to_float(x), weights, means, precisions_cholesky)
     Nx.exp(log_resp)
   end
 end
--- a/lib/scholar/cluster/gmm.ex
+++ b/lib/scholar/cluster/gmm.ex
@@ -3,7 +3,7 @@ defmodule Scholar.Cluster.GaussianMixture do
   Gaussian Mixture Model.
 
   Gaussian Mixture Model is a probabilistic model that assumes every data point is generated
-  by randomly choosing one of the several fixed Gaussian distributions and then sampling from it.
+  by choosing one of several fixed Gaussian distributions and then sampling from it.
   Its parameters are estimated using the Expectation-Maximization (EM) algorithm, which is an
   iterative algorithm alternating between the two steps: the E-step which computes the
   expectation of the Gaussian assignment for each data point x and the M-step which updates the
@@ -32,7 +32,7 @@ defmodule Scholar.Cluster.GaussianMixture do
     num_runs: [
       type: :pos_integer,
       default: 1,
-      doc: "The number of initializations to perform."
+      doc: "The number of times to initialize parameters and run the entire EM algorithm."
     ],
     max_iter: [
       type: :pos_integer,
@@ -44,7 +44,7 @@ defmodule Scholar.Cluster.GaussianMixture do
       default: 1.0e-3,
       doc: "The convergence threshold."
     ],
-    reg_covar: [
+    covariance_regularization_eps: [
       type: {:custom, Scholar.Options, :non_negative_number, []},
       default: 1.0e-6,
       doc: """
@@ -127,12 +127,15 @@ defmodule Scholar.Cluster.GaussianMixture do
     num_runs = opts[:num_runs]
     max_iter = opts[:max_iter]
     tol = opts[:tol]
-    reg_covar = opts[:reg_covar]
+    covariance_regularization_eps = opts[:covariance_regularization_eps]
 
     {best_params, _} =
       while {
               best_params =
-                init_params(key, x, num_gaussians: num_gaussians, reg_covar: reg_covar),
+                init_params(key, x,
+                  num_gaussians: num_gaussians,
+                  covariance_regularization_eps: covariance_regularization_eps
+                ),
               {
                 run = 1,
                 max_lower_bound = Nx.Constants.neg_infinity(),
@@ -140,7 +143,7 @@ defmodule Scholar.Cluster.GaussianMixture do
                 key,
                 max_iter,
                 tol,
-                reg_covar
+                covariance_regularization_eps
               }
             },
             run <= num_runs do
@@ -149,7 +152,10 @@ defmodule Scholar.Cluster.GaussianMixture do
                   {
                     lower_bound = Nx.Constants.neg_infinity(),
                     params =
-                      init_params(key, x, num_gaussians: num_gaussians, reg_covar: reg_covar)
+                      init_params(key, x,
+                        num_gaussians: num_gaussians,
+                        covariance_regularization_eps: covariance_regularization_eps
+                      )
                   },
                   {
                     iter = 1,
@@ -158,21 +164,21 @@ defmodule Scholar.Cluster.GaussianMixture do
                     key,
                     max_iter,
                     tol,
-                    reg_covar
+                    covariance_regularization_eps
                   }
                 },
                 iter <= max_iter and not converged do
             prev_lower_bound = lower_bound
             {weights, means, _, precisions_cholesky} = params
             {log_prob_norm, log_resp} = e_step(x, weights, means, precisions_cholesky)
-            params = m_step(x, log_resp, reg_covar)
+            params = m_step(x, log_resp, covariance_regularization_eps)
             lower_bound = log_prob_norm
             change = lower_bound - prev_lower_bound
             converged = iter > 1 and Nx.abs(change) < tol
 
             {
               {lower_bound, params},
-              {iter + 1, converged, x, key, max_iter, tol, reg_covar}
+              {iter + 1, converged, x, key, max_iter, tol, covariance_regularization_eps}
             }
           end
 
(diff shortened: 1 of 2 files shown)


msluszniak (approved):
Thank you for your contribution! The PR LGTM :)

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
+defmodule Scholar.Cluster.GaussianMixture do
+  @moduledoc """
+  Gaussian Mixture Model.
+
+  Gaussian Mixture model is a probabilistic model that assumes every data point is generated
```
polvalente, line 5:
```suggestion
  Gaussian Mixture Model is a probabilistic model that assumes every data point is generated
```

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+  by randomly choosing one of the several fixed Gaussian distributions and then sampling from it.
```
polvalente, line 6: "randomly choosing" might be a loaded expression here. I got stuck on thinking about if this is random uniform, normal or whatever else method of choosing.

```suggestion
  b choosing one of several fixed Gaussian distributions and then sampling from it.
```

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+  Its parameters are estimated using the Expectation-Maximization (EM) algorithm, which is an
+  iterative algorithm alternating between the two steps: the E-step which computes the
+  expectation of the Gaussian assignment for each data point x and the M-step which updates the
+  parameters to maximize the expectations found in E-step. While every iteration of the algorithm
+  is guaranteed to improve the log-likelihood, the final result depends on the initial values of
+  the parameters and the entire procedure should be repeated several times.
```
polvalente, line 12: From reading without looking at the code beforehand:

Does this mean that we need to run the algorithm multiple times, or does the algorithm itself repeat this procedure?

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+  require Nx
+
+  @derive {Nx.Container, containers: [:weights, :means, :covariances, :precisions_cholesky]}
+  defstruct [:weights, :means, :covariances, :precisions_cholesky]
+
+  opts = [
+    num_gaussians: [
+      required: true,
+      type: :pos_integer,
+      doc: "The number of Gaussian components in the mixture."
```
polvalente, line 30:
```suggestion
      doc: "The number of Gaussian model components in the mixture."
```

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+    ],
+    num_runs: [
+      type: :pos_integer,
+      default: 1,
+      doc: "The number of initializations to perform."
```
polvalente, line 35: Not sure if "initializations" is the best word because I assume we're talking about running the full procedure and not just initializing it, right?

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+      type: :pos_integer,
+      default: 100,
+      doc: "The number of EM iterations to perform."
+    ],
+    tol: [
+      type: {:custom, Scholar.Options, :non_negative_number, []},
+      default: 1.0e-3,
+      doc: "The convergence threshold."
+    ],
+    reg_covar: [
```
polvalente, line 47:
```suggestion
    covariance_regularization_eps: [
```

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+      type: {:custom, Scholar.Options, :non_negative_number, []},
+      default: 1.0e-6,
+      doc: """
+      The non-negative number that is added to each element of the diagonal
+      of the covariance matrix to ensure it is positive.
```
polvalente, line 52:
```suggestion
      of the covariance matrix to ensure it is positive. Usually a small number.
```

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+                run = 1,
+                max_lower_bound = Nx.Constants.neg_infinity(),
+                x,
+                key,
+                max_iter,
+                tol,
+                reg_covar
+              }
+            },
+            run <= num_runs do
```
polvalente, line 146: Since there's no early stopping, we might be able to apply vectorization to get rid of this outer while loop

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+
+    {weights, means, covariances, precisions_cholesky} =
+      estimate_gaussian_parameters(x, resp, reg_covar)
+
+    weights = weights / num_samples
+    {weights, means, covariances, precisions_cholesky}
+  end
+
+  defnp e_step(x, weights, means, precisions_cholesky) do
+    {log_prob_norm, log_resp} = estimate_log_prob_resp(x, weights, means, precisions_cholesky)
```
polvalente, line 222: what does _resp mean here?

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+    {weights, means, covariances, precisions_cholesky} =
+      estimate_gaussian_parameters(x, Nx.exp(log_resp), reg_covar)
+
+    weights = weights / Nx.sum(weights)
+    {weights, means, covariances, precisions_cholesky}
+  end
+
+  defnp estimate_gaussian_parameters(x, resp, reg_covar) do
+    nk = Nx.sum(resp, axes: [0])
+    means = Nx.dot(Nx.transpose(resp), x) / nk
```
polvalente, line 236: it might be more numerically stable to do:

```suggestion
    means = Nx.dot(Nx.transpose(resp) / nk, x / nk)
```

polvalente: Also, if `resp` can be complex, you might want Nx.LinAlg.adjoint instead

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+            k < num_gaussians do
+        diff = x - means[k]
+        covariance = Nx.dot(resp[[.., k]] * Nx.transpose(diff), diff) / nk[k]
+        covariance = Nx.put_diagonal(covariance, Nx.take_diagonal(covariance) + reg_covar)
+
+        {
+          Nx.put_slice(covariances, [k, 0, 0], Nx.new_axis(covariance, 0)),
+          {k + 1, x, resp, nk, means, reg_covar}
+        }
+      end
```
polvalente, line 262: This while also reads like something that can benefit from vectorization

polvalente: Also, we can leave usage of vectorization as TO-DO comments instead, unless you really want to tackle this. It might even be the case where we merge this and you tackle it afterwards. Thoughts?

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+
+    covariances
+  end
+
+  defnp compute_precisions_cholesky(covariances) do
+    type = Nx.type(covariances)
+    {num_gaussians, num_features, _} = Nx.shape(covariances)
+
+    {precisions_cholesky, _} =
+      while {
```
polvalente, line 272: Same thing about vectorization here

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+  end
+
+  defnp estimate_log_prob(x, means, precisions_cholesky) do
+    type = Nx.type(x)
+    {num_samples, num_features} = Nx.shape(x)
+    {num_gaussians, _} = Nx.shape(means)
+    log_det = compute_log_det_cholesky(precisions_cholesky)
+
+    {log_prob, _} =
+      while {
```
polvalente, line 313: vectorization too

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+
+  ## Examples
+
+      iex> key = Nx.Random.key(12)
+      iex> x = Nx.tensor([[1, 2], [1, 4], [1, 0], [10, 2], [10, 4], [10, 0]])
+      iex> model = Scholar.Cluster.GaussianMixture.fit(x, num_gaussians: 2, key: key)
+      iex> Scholar.Cluster.GaussianMixture.predict(model, Nx.tensor([[8, 1], [2, 3]]))
+      Nx.tensor(
+        [1, 0]
+      )
```
polvalente, line 367: I don't think this will pass on CI

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+      iex> key = Nx.Random.key(12)
+      iex> x = Nx.tensor([[1, 2], [1, 4], [1, 0], [10, 2], [10, 4], [10, 0]])
+      iex> model = Scholar.Cluster.GaussianMixture.fit(x, num_gaussians: 2, key: key)
+      iex> Scholar.Cluster.GaussianMixture.predict_prob(model, Nx.tensor([[8, 1], [2, 3]]))
+      Nx.tensor(
+        [
+          [0.0, 1.0],
+          [1.0, 0.0]
+        ]
+      )
```
polvalente, line 397: This as well

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,407 @@
...
+                  }
+                },
+                iter <= max_iter and not converged do
+            prev_lower_bound = lower_bound
+            {weights, means, _, precisions_cholesky} = params
+            {log_prob_norm, log_resp} = e_step(x, weights, means, precisions_cholesky)
+            params = m_step(x, log_resp, reg_covar)
+            lower_bound = log_prob_norm
+            change = lower_bound - prev_lower_bound
+            converged = Nx.abs(change) < tol
```
polvalente, line 171:
```suggestion
            converged = iter > 1 and Nx.abs(change) < tol
```

polvalente (commented):
Lots of really minor comments. Overally this is looking great!

lib/scholar/cluster/gmm.ex
krstopro: Yes, it is a multinomial random distribution. Basically GMM can be viewed as a generative model in the following way:
- First sample one out of K Gaussian distributions from multinomial random distribution with parameters w_1, ..., w_K (w_k is the probability of sampling k-th Gaussian; this is the `weights` field of the model).
- Then sample from that Gaussian distribution.

lib/scholar/cluster/gmm.ex
krstopro: The algorithm itself repeats the procedure `num_runs` amount of times, each time with different initial parameters. Similarly as it was already done in k-means implementation.

lib/scholar/cluster/gmm.ex
krstopro: Yes, running the full procedure but with different initial parameters. Indeed, word initializations might be misleading here.

lib/scholar/cluster/gmm.ex
krstopro: The Gaussian here refers to the Gaussian distribution. Perhaps writing "The number of Gaussian distributions in the mixture" would be better.

lib/scholar/cluster/gmm.ex
krstopro: `resp` is short for responsibility or responsibilities. `resp[n][k]` measure how much is the k-th Gaussian responsible for generating the n-th observation. Perhaps using `responsibilities` instead of `resp` might be a better idea.

lib/scholar/cluster/gmm.ex
krstopro: `resp` are probabilities, so `Nx.transpose` will suffice. Should it be `Nx.sqrt(nk)` in the second expression? Is this really more numerically stable?

polvalente: If resp are probabilities then it won't make much difference, but you can just divide the other term by nk instead.

My concern was that for 2 unbounded tensors, the dot product could blow up a bit, and the division tames that to some degree.

Since one of those is bounded between 0 and 1 in each element, this isn't that much of a concern

lib/scholar/cluster/gmm.ex
krstopro: Indeed, this might be vectorizable. I would need to think about lines 255 and 256 though as it's not that straightforward.
I would say let's merge it like this and I can continue working on it (I intend to continue contributing to `Nx` anyway).

polvalente: Sure! To be clear, vectorization is not a blocker at all for this PR

lib/scholar/cluster/gmm.ex
krstopro: Yeah, this one might be easier as both `Nx.LinAlg.cholesky` and `Nx.LinAlg.triangular_solve` work with batches of matrices.

polvalente: You don't have to think much about this. Nx main has Nx.vectorize/devectorize/revectorize which will handle batching for you

lib/scholar/cluster/gmm.ex
krstopro: Yeah, also doesn't seem hard to vectorize.

lib/scholar/cluster/gmm.ex
krstopro: One thing was strange to me. Variable `type` is not declared in the `while` loop and it still compiles. Is this intended?

polvalente: In fact, I'm not sure it _would_ compile if you passed type as an argument :)

type is a tuple of {atom, number} used only at compile time. Passing as argument would force its traversal as an Nx.Container and that would break on the Atom.

It's ok anyway because the type is just used at compile-time

lib/scholar/cluster/gmm.ex
krstopro: Hmmm why exactly? `mix test` also checks the docstrings and runs the code inside.
It's similar to the k-means implementation:
https://github.com/elixir-nx/scholar/blob/main/lib/scholar/cluster/k_means.ex#L297-L299

josevalim: Yup, this is a doctest feature :)

lib/scholar/cluster/gmm.ex
krstopro: Why is `iter > 1` necessary here?

lib/scholar/cluster/gmm.ex
polvalente: TIL :)

lib/scholar/cluster/gmm.ex
polvalente: My previous version of the comment had an explanation haha

I think it would be nice to avoid relying on comparison with neg infinity (the initial value) in the first iteration

krstopro: Oh, good catch! But it's probably still not needed.
`Nx.subtract(12, Nx.Constants.neg_infinity) |> Nx.abs`
gives
```
#Nx.Tensor<
  f32
  Inf
>
```
which is larger than any `tol`. :)

polvalente: Yeah, I just don't _like_ comparisons and operations with inf. But even in the worst case where we get `lower_bound = neg_infinity`, `neg_infinity - neg_infinity = nan`, which will also fail your check because it breaks on `true`

lib/scholar/cluster/gmm.ex
krstopro: Indeed, dot product can explode, but here it should not be a problem. Rewriting it as
` means = Nx.dot(Nx.transpose(resp), x / nk)` might be a good idea though.

lib/scholar/cluster/gmm.ex
krstopro: The only way `lower_bound` ends up being `neg_infinity` at that point is due to numerical error as it's updated `lower_bound = log_prob_norm` at line 169. But yeah, let's avoid operations with infinity.

lib/scholar/cluster/gmm.ex
msluszniak: I think that in the case of k-means, vectorization is more beneficial. You can simply reduce `num_runs` if the memory is the problem. But you cannot apply vectorization programmatically if you use two nested whiles. Nowadays, computation time is much more important than space limitation. You might benchmark two versions on the example that can be loaded into memory and test how much faster the vectorized version is. If the change is insignificant for GMM, we should stay with a memory-efficient solution.

msluszniak: And it's always easier to change vectorized version into 'while', so if people had a memory problem on k-means, then we would use a memory-efficient solution.

msluszniak: And in Sci-kit they cannot run k-means on GPU, so they have no clear advantage of implementing vectorization which is not obvious to understand at first glance, harder to implement (because of broadcasting), memory consuming etc.

lib/scholar/cluster/gmm.ex
krstopro: I guess "The number of times to initialize parameters and run the entire EM algorithm." might be a better description here.

lib/scholar/cluster/gmm.ex
krstopro: I am not sure if I should change anything here. Thoughts?

lib/scholar/cluster/gmm.ex
krstopro: Agreed.

lib/scholar/cluster/gmm.ex
krstopro: Not sure about this as well. Thoughts?

lib/scholar/cluster/gmm.ex
krstopro: Agreed about the suggestion though.

krstopro:
@msluszniak @polvalente Thanks for the support and given feedback! I've addressed most of the suggestions provided; see the comments above (except for vectorization that I would leave for now). There are few more remarks that I would like to make; see the comments below.

lib/scholar/cluster/gmm.ex
josevalim: Nx only checks the arguments between def <-> defn boundaries.

lib/scholar/cluster/gmm.ex
msluszniak: For me, the description is clear. @polvalente do you have any suggestions?

lib/scholar/cluster/gmm.ex
msluszniak: Agreed. This description will be maximally informative.

lib/scholar/cluster/gmm.ex
msluszniak: If it wasn't intuitive for Paulo, then I think that it is worth changing it to `responsibilities`.

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,418 @@
...
+    {weights, means, covariances, precisions_cholesky} =
+      estimate_gaussian_parameters(x, resp, covariance_regularization_eps)
+
+    weights = weights / num_samples
+    {weights, means, covariances, precisions_cholesky}
+  end
+
+  defnp e_step(x, weights, means, precisions_cholesky) do
+    {log_prob_norm, log_resp} = estimate_log_prob_resp(x, weights, means, precisions_cholesky)
+    {Nx.mean(log_prob_norm), log_resp}
```
msluszniak, line 229:
```suggestion
    {log_prob_norm, log_responsibilities} = estimate_log_prob_responsibilities(x, weights, means, precisions_cholesky)
    {Nx.mean(log_prob_norm), log_responsibilities}
```

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,418 @@
...
+
+    precisions_cholesky
+  end
+
+  defnp estimate_log_prob_resp(x, weights, means, precisions_cholesky) do
+    weighted_log_prob = estimate_weighted_log_prob(x, weights, means, precisions_cholesky)
+    log_prob_norm = logsumexp(weighted_log_prob)
+    log_resp = weighted_log_prob - Nx.new_axis(log_prob_norm, 1)
+    {log_prob_norm, log_resp}
+  end
```
msluszniak, line 311:
```suggestion
  defnp estimate_log_prob_responsibilities(x, weights, means, precisions_cholesky) do
    weighted_log_prob = estimate_weighted_log_prob(x, weights, means, precisions_cholesky)
    log_prob_norm = logsumexp(weighted_log_prob)
    log_responsibilities = weighted_log_prob - Nx.new_axis(log_prob_norm, 1)
    {log_prob_norm, log_responsibilities}
  end
```

msluszniak:
Are there any blockers for this PR?

krstopro:
@msluszniak Don't think so. I just need to address few suggestions that you made. Could you have a look at the two reviews I opened? The `logsumexp` one might be important.

krstopro:
> Sure, but tbh I cannot see them 😅

@msluszniak They are probably collapsed among many others :)
Tagged you in the comment.

msluszniak:
Are you sure that your comments are not in `pending` mode or in the batch and not pushed or sth? Because I really cannot find anything

krstopro:
> Are you sure that your comments are not in `pending` mode or in the batch and not pushed or sth? Because I really cannot find anything

Sorry, they were in pending mode and I don't think I am able to submit them. Hence, I'll write them here.
Line 357-366 includes implementation to `logsumexp` which basically calculates the logarithm of the sum of the exponentials of the tensor in a numerically stable way. This is a well known trick that is implemented in numerical libraries, e.g. see SciPy
https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.logsumexp.html.
However, I am not aware that such a function exists in `Nx`. Is there one or perhaps I should raise an issue about implementing it?

Line 292: doing {num_gaussians, num_features, num_features} = Nx.shape(covariances) inside `defn` won't work, even though it works outside. I suppose this is intended?

msluszniak:
I'm almost sure that there is no such function implemented in `Nx` so feel free to send a PR with it. I'm not sure if we eventually want to have this type of functions in `Nx` or in `Scholar`, but it will be helpful.

This line won't work in `defn` because you're doing pattern matching. If you want to assert that the shape is as follows you need to check for it separately.

polvalente:
You can use `case` for checking the validity of the shape inside defn

krstopro:
> You can use `case` for checking the validity of the shape inside defn

I see, thanks. Here it is not an issue, just wanted to check that this is intended.
Anyway, if everything else is fine can you approve the PR?

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,433 @@
...
+  Gaussian Mixture Model.
+
+  Gaussian Mixture Model is a probabilistic model that assumes every data point is generated
+  by choosing one of several fixed Gaussian distributions and then sampling from it.
+  Its parameters are estimated using the Expectation-Maximization (EM) algorithm, which is an
+  iterative algorithm alternating between the two steps: the E-step which computes the
+  expectation of the Gaussian assignment for each data point x and the M-step which updates the
+  parameters to maximize the expectations found in E-step. While every iteration of the algorithm
+  is guaranteed to improve the log-likelihood, the final result depends on the initial values of
+  the parameters and the entire procedure should be repeated several times.
```
polvalente, line 12:
```suggestion
  the parameters. Thus, the algorithm consists of repeating this procedure several times to find the best result. 
```

WDYT? @msluszniak @krstopro

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,433 @@
...
+    {weights, means, covariances, precisions_cholesky} =
+      estimate_gaussian_parameters(x, Nx.exp(log_responsibilities), covariance_regularization_eps)
+
+    weights = weights / Nx.sum(weights)
+    {weights, means, covariances, precisions_cholesky}
+  end
+
+  defnp estimate_gaussian_parameters(x, responsibilities, covariance_regularization_eps) do
+    nk = Nx.sum(responsibilities, axes: [0])
+    means = Nx.dot(Nx.transpose(responsibilities), x / nk)
```
polvalente, line 245:
```suggestion
    means = Nx.dot(responsibilities, [0], x / nk, [0])
```

I believe this will achieve the correct result too, but avoiding the transpose.

```elixir
iex(37)> k = Nx.Random.key(System.system_time())
#Nx.Tensor<
  u32[2]
  EXLA.Backend<host:0, 0.3654900815.2985164818.202632>
  [393393791, 2867920131]
>
iex(38)> {t1, k} = Nx.Random.uniform(k, shape: {3, 4})
{#Nx.Tensor<
   f32[3][4]
   EXLA.Backend<host:0, 0.3654900815.2985164818.203175>
   [
     [0.13576042652130127, 0.8412826061248779, 0.4484595060348511, 0.5624940395355225],
     [0.3512943983078003, 0.004010677337646484, 0.32932519912719727, 0.3748883008956909],
     [0.5318529605865479, 0.8757333755493164, 0.39265263080596924, 0.1744011640548706]
   ]
 >,
 #Nx.Tensor<
   u32[2]
   EXLA.Backend<host:0, 0.3654900815.2985164818.203177>
   [1882076740, 466335474]
 >}
iex(39)> {t2, k} = Nx.Random.uniform(k, shape: {3, 5})
{#Nx.Tensor<
   f32[3][5]
   EXLA.Backend<host:0, 0.3654900815.2985164818.203698>
   [
     [0.36859428882598877, 0.4500901699066162, 0.6306818723678589, 0.5150032043457031, 0.25934934616088867],
     [0.5741105079650879, 0.45720410346984863, 0.05595088005065918, 0.6362648010253906, 0.32299208641052246],
     [0.1018531322479248, 0.21718335151672363, 0.5860718488693237, 0.7845090627670288, 0.2448132038116455]
   ]
 >,
 #Nx.Tensor<
   u32[2]
   EXLA.Backend<host:0, 0.3654900815.2985164818.203700>
   [1307000727, 4256395817]
 >}
iex(40)> Nx.equal(Nx.dot(Nx.transpose(t1), t2), Nx.dot(t1, [-2], t2, [-2]))       #Nx.Tensor<
  u8[4][5]
  EXLA.Backend<host:0, 0.3654900815.2985164818.203708>
  [
    [1, 1, 1, 1, 1],
    [1, 1, 1, 1, 1],
    [1, 1, 1, 1, 1],
    [1, 1, 1, 1, 1]
  ]
>
```

lib/scholar/cluster/gmm.ex
```diff
@@ -0,0 +1,433 @@
...
+    {covariances, _} =
+      while {
+              covariances =
+                Nx.tensor(0.0, type: type)
+                |> Nx.broadcast({num_gaussians, num_features, num_features}),
+              {k = 0, x, responsibilities, nk, means, covariance_regularization_eps}
+            },
+            k < num_gaussians do
+        diff = x - means[k]
+        covariance = Nx.dot(responsibilities[[.., k]] * Nx.transpose(diff), diff) / nk[k]
```
polvalente, line 273:
```suggestion
        covariance = Nx.dot(responsibilities[[.., k]] * diff, [-2], diff / nk[k], [-2])
```

msluszniak (approved):

josevalim:
@krstopro please verify @polvalente suggestions above and we can ship it :)

lib/scholar/cluster/gmm.ex
krstopro: How about `means = Nx.dot(responsibilities, [0], x / nk, [0])`?
Using `[0]` instead of `[-2]` seems more understandable to me.

lib/scholar/cluster/gmm.ex
krstopro: This won't work as `Nx.transpose` is only applied to `diff`, not the entire product. Adding another axis to `responsibilities[[.., k]]` solves the problem.
`covariance = Nx.dot(diff * Nx.new_axis(responsibilities[[.., k]], 1), [0], diff / nk[k], [0])`

lib/scholar/cluster/gmm.ex
polvalente: I went with -2 due to force of habit of supporting batched versions of algorithms haha

Both are fine to me if you're enforcing 2D tensors here

krstopro: Yeah, I had a similar habit with Tensorflow / PyTorch. I would use `[0]` here as both are matrices.

lib/scholar/cluster/gmm.ex
krstopro: I would go for "Thus the procedure consists of repeating the algorithm several times and taking the best obtained result" as "algorithm" refers to EM algorithm mentioned before.

josevalim:
```suggestion
  the parameters. Thus the procedure consists of repeating the algorithm several times
  and taking the best obtained result.
```

josevalim:

Issue #122, opened 2023-06-26
Add classification and regression metrics


msluszniak:
I want to add all the existing metrics which are present in Sci-kit Learn. 

**Classification Metrics**
* [x] Accuracy
* [x] AUC
* [x] Average Precision (AP)
* [x] Balanced Accuracy
* [x] Brier Score Loss
* [x] Cohen Kappa Score
* [x] Confusion Matrix
* [x] Discounted Cumulative Gain (DCG)
* [x] Detection Error Tradeoff Curve (DET curve)
* [x] F1 Score 
* [x] F-beta Score
* [x] Hamming Loss
* [x] Log Loss
* [x] Matthews correlation coefficient (MCC)
* [x] Normalized Discounted Cumulative Gain (Normalized DCG)
* [x] Precision Recall Curve
* [x] Precision Recall Fscore Support
* [x] Precision
* [x] Recall
* [x] ROC-AUC 
* [x] ROC Curve
* [x] Top-k Accuracy Score
* [x] Zero-one Classification Loss

**Regression Metrics**
* [x] Explained Variance Score
* [x] Maximum Residual Error
* [x] Mean Absolute Error (MAE)
* [x] Mean Squared Error (MSE)
* [x] Mean Squared Log Error
* [x] Mean Absolute Percentage Error (MAPE)
* [x] $R^2$ Score
* [x] Mean Poisson deviance
* [x] Mean Gamma deviance
* [x] Mean Tweedie deviance
* [x] $D^2$ - Fraction of Tweedie Deviance Explained
* [x] Mean Pinball Loss
* [x] $D^2$ - Fraction of Pinball Loss Explained
* [x] $D^2$ - Fraction of Absolute Error Explained

We can do the same for distances after (look #149).

paulsullivanjr:
I'd like to contribute to this project. I will start working on a PR for "Discounted Cumulative Gain (DCG)"

msluszniak:
> I'd like to contribute to this project. I will start working on a PR for "Discounted Cumulative Gain (DCG)"

Really cool! Feel free to go :)

paulsullivanjr:
> > I'd like to contribute to this project. I will start working on a PR for "Discounted Cumulative Gain (DCG)"
> 
> Really cool! Feel free to go :)

Going to try to have a PR ready by the end of the week so I can get some feedback.

paulsullivanjr:
I have Normalized DCG in progress.

[...]

PR #139, opened 2023-07-15
Add MSLE and MAPE regression metrics (see #122)

santiago-imelio:


lib/scholar/metrics.ex
```diff
@@ -646,6 +646,65 @@ defmodule Scholar.Metrics do
...
+  @doc ~S"""
+  Calculates the mean square logarithmic error of predictions
+  with respect to targets.
+
+  $$MSLE = \frac{\sum_{i=1}^{n} (\log(\hat{y_i} + 1) - \log(y_i + 1))^2}{n}$$
+
+  ## Examples
+
+      iex> y_true = Nx.tensor([[0.0, 1.0], [0.0, 0.0]], type: {:f, 32})
+      iex> y_pred = Nx.tensor([[1.0, 1.0], [1.0, 0.0]], type: {:f, 32})
```
josevalim, line 658:
```suggestion
      iex> y_true = Nx.tensor([[0.0, 1.0], [0.0, 0.0]])
      iex> y_pred = Nx.tensor([[1.0, 1.0], [1.0, 0.0]])
```

lib/scholar/metrics.ex
```diff
@@ -646,6 +646,65 @@ defmodule Scholar.Metrics do
...
+        f32
+        0.24022650718688965
+      >
+  """
+  defn mean_square_log_error(y_true, y_pred) do
+    assert_same_shape!(y_true, y_pred)
+
+    (Nx.log(y_true + 1) - Nx.log(y_pred + 1))
+    |> Nx.pow(2)
+    |> Nx.mean()
```
msluszniak, line 670:
```suggestion
    mean_square_error(Nx.log(y_true + 1), Nx.log(y_pred + 1))
```

lib/scholar/metrics.ex
```diff
@@ -646,6 +646,65 @@ defmodule Scholar.Metrics do
...
+      iex> y_pred = Nx.tensor([1.2, 0.1, 2.4, 8.0])
+      iex> Scholar.Metrics.mean_absolute_percentage_error(y_true, y_pred)
+      #Nx.Tensor<
+        f32
+        2.5e8
+      >
+  """
+  defn mean_absolute_percentage_error(y_true, y_pred) do
+    assert_same_shape!(y_true, y_pred)
+    eps = Nx.tensor(1.0e-10)
```
msluszniak, line 702:
```suggestion
    eps = Nx.Constants.epsilon({:f, 64})
```

msluszniak (changes requested):
I dropped some suggestions. After applying them, it's good to go.

lib/scholar/metrics.ex
josevalim: Should the epsilon be based on the input type?

msluszniak: I don't think so, because eps(:f32) will be almost always too big (~1.0e-7)

msluszniak: If the role of an eps is to replace 0s in division then we almost always will use `:f64`

msluszniak: On the other hand, if we generate noise in the algorithm and the standard deviation depends on eps, then eps should be dependent on the type

msluszniak: The problem is that `Nx.Constants.epsilon(:f64)` forces output to be `:f64` but ideally we would preserve type if we could (`:f32`), but use the value of `Nx.Constants.epsilon(:f64)`

josevalim: Doesn’t it mean we should have a lower value for epsilon on f32 then?

polvalente: The epsilon value we have is well-defined and is the same that numpy provides.

I believe it's the smallest value where 1+eps returns an accurate result.

msluszniak: Well, to be honest, it's not important which convention we eventually choose.  The result if `eps` is max in the denominator, then we end up with a huge result which can only be interpreted as a nearly zero value somewhere in `y_true` (See second doctest). So I'm ok with either version. Code for type dependant version would be `eps = to_float_type(Nx.Type.merge(Nx.type(x), Nx.type(y))`

josevalim:

PR #140, opened 2023-07-16
Add K-fold

### LINKED ISSUE #133: Add KFold

msluszniak:
Closes #133 

Currently, we return a struct that contains preprocessed data (cropped so the size of axis 0 is divisible by `k` and batched into k parts) and indices for training and evaluation. I believe there are two main ways how to use k-fold in Scholar.

1. Pass the whole partitioned dataset into an algorithm which deals with batched data
2. Access one chunk of K-fold in every loop of `while`

Now if we access data using `training_indices` we will get data that have a non-intuitive shape, which can be fixed by a simple function that I'll implement. However, we could also reshape the data earlier and then return `train` and `val` data instead of indices. This approach has a drawback, cause `train_data` will be roughly `k` times bigger than the original data. So if we use a second workflow with `while` then we potentially allocate a big tensor unnecessarily.


josevalim:
We only need to do the padding upfront but not the batching. If we have the data, we only need to store the batch index to access the relevant parts out of it.

On the other hand, now that I look at the code, I don't think k-fold needs to own the `while` loop. For example, we could do the traversal in Elixir itself and then you pass each batch to `defn`? How would that be different from `to_batched` for example? https://hexdocs.pm/nx/Nx.html#to_batched/3

msluszniak:
Ok, ofc you're absolutely right. I change the code.

josevalim:
And now I am wondering: do we really need kfold? Nx.split can train-test split the data, to_batched can generate the k-fold with batching. The only non-obvious operation is shuffling. Or am I missing something?

msluszniak:
Honestly, I don't see how to perform K-fold simply using `Nx.split` and `Nx.to_batched`

josevalim:
Ok, so I am most likely missing something. Let me think about this during the day and I will follow up.

msluszniak:
I mean, it will be very simple function like this

```elixir
  defnp k_fold_n(data, key, opts \\ []) do
    k = opts[:k]
    data = data[[0..(div(Nx.axis_size(data, 0), k) * k - 1)]]

    mask =
      Nx.transpose(Nx.tri(k - 1, k))
      |> Nx.new_axis(-1)
      |> Nx.broadcast({k, k - 1, div(Nx.axis_size(data, 0), k)})
      |> Nx.reshape({k, :auto})

    train_indices = Nx.iota({(k - 1) * div(Nx.axis_size(data, 0), k)}) |> Nx.tile([k, 1])
    train_indices = Nx.select(mask, train_indices + div(Nx.axis_size(data, 0), k), train_indices)
    validation_indices = Nx.iota({k, div(Nx.axis_size(data, 0), k)})
    {train_indices, validation_indices}
  end
```

josevalim:
So I have been thinking more about this and I wonder if we should implement this as a stream instead. The downside is that it doesn't work inside `defn` but we could eventually support it (although I am not sure how worth it would be). Here is an implementation of it as a stream:

```elixir
Mix.install [:nx]

defmodule KFold do
  def stream(nx, k) do
    Stream.resource(
      fn ->
        fold_size = ceil(Nx.axis_size(nx, 0) / k)
        {nx |> Nx.to_batched(fold_size, leftover: :repeat) |> Enum.to_list(), 0, k}
      end,
      fn
        {list, k, k} ->
          {:halt, {list, k, k}}

        {list, current, k} ->
          {left, [test | right]} = Enum.split(list, current)
          {[{left ++ right, test}], {list, current + 1, k}}
      end,
      fn _ -> :ok end
    )
  end
end

nx = Nx.iota({9, 4})
KFold.stream(nx, 2) |> Enum.to_list() |> IO.inspect
KFold.stream(nx, 5) |> Enum.to_list() |> IO.inspect
```

Each iteration emits a `{train_list, test}` entry. Now you can do whatever you want with it. Thoughts?

msluszniak:
The stream will be very convenient. I think it would be sufficient if it were available in `defn`.  So, maybe until we add support for `streams` in `Nx`, we add two functions, one that generates tensor for those who need `k-fold` inside `defn` and stream for the rest of them?

polvalente:
I'm with what @seanmor5 mentioned in Slack. The full name for k-fold is "K-Fold Cross Validation", and this name implies it's a training algorithm for supervised (supervised-like) learning where you have {sample, expected output} pairs on which you want to train and validade your data, generally stemming from the training set.

This means that the core part of the Nx function it would need to provide is just this:

```
deftransform k_splits(tensor_with_batch_axis, k) do
  split_size = div(Nx.axis_size(tensor_with_batch_axis, 0), k)
  
  1..k
  |> Enum.map(&Nx.slice_along_axis(tensor_with_batch_axis, (&1 - 1) * k, k, axis: 0)
  |> List.to_tuple()
end
```

And the whole rest of the k-folds algorithm pertains to training some ML algorithm, which in general is something that we have either already JIT-compiled, or can treat as the most expensive part to run. I see this living on the same level of abstraction as `Axon.Loop`.

msluszniak:
Hmm, I think that we should have a method that only splits data and then you can have your very own validation loop with some training data augmentation, and some other operations other than just fit and predict calls. But of course, the `CV` function also is a must. However, I'm wondering what interface would be the best for `CV`. We can for example accept a module name that implements functions `fit` and `predict` and call them inside the loop. Then we can also use my implementation of splitting without streams.

josevalim:
@msluszniak you can always easily convert the stream into a list and do any processing that you want, no? The stream is then just a convenience API. scikit-learn also works as Python generators, right?

msluszniak:
Yes, that's right. So in fact, we'll stick to stream implementation. But my main point is that we need both the `cross_validation` and splitting function in our module even though `CV` will use this splitting function.

josevalim:
Ah, it is fine to me to add new functions :)

msluszniak:
> 

`Nx.to_batched` can only emit batches of the same size, right?

msluszniak:
Ok, so we must change the function to cut the rest of the data.

```elixir
  def k_fold_split(container, k) do
    Stream.resource(
      fn ->
        fold_size = floor(Nx.axis_size(container, 0) / k)
        {container |> Nx.to_batched(fold_size, leftover: :discard) |> Enum.to_list(), 0, k}
      end,
      fn
        {list, k, k} ->
          {:halt, {list, k, k}}

        {list, current, k} ->
          {left, [test | right]} = Enum.split(list, current)
          {[{left ++ right, test}], {list, current + 1, k}}
      end,
      fn _ -> :ok end
    )
  end
```

The reason why it's a must is that when we repeat the data we may "contaminate" the validation data with training samples, which will introduce bias to our validation. So we rather drop some data, than accept this data leakage.

msluszniak:
And there is also my draft of `cv` and `cv_score`, but I'm wondering how we deal with options in `fit` and `predict` functions and the fact that some `predict` functions don't accepts `opts`.

```elixir
  def cross_validation(x, y, k, fun_module, scoring_fun) do
    Stream.zip([k_fold_split(x, k), k_fold_split(y, k)])
    |> Enum.map(fn {{train_x, test_x}, {train_y, test_y}} ->
      model = fun_module.fit(Nx.concatenate(train_x), Nx.concatenate(train_y))
      pred = fun_module.predict(model, test_x)
      scoring_fun.(pred, test_y)
    end)
  end

  def cross_validation_score(x, y, k, fun_module, scoring_fun) do
    cross_validation(x, y, k, fun_module, scoring_fun)
    |> Enum.map(fn x -> Nx.new_axis(x, 0) end)
    |> Nx.concatenate(axis: 0)
    |> Nx.mean(axes: [0])
  end
```

josevalim:
I would make it so cross_validation is only the stream too:

```
  def cross_validation(x, y, k) do
    Stream.zip([k_fold_split(x, k), k_fold_split(y, k)])
  end
```

And for score:

```
  def cross_validation_score(x, y, k, fun) do
    cross_validation(x, y, k)
    |> Enum.map(fun)
    |> Nx.stack()
    |> Nx.mean(axes: [0])
  end
```

Where fun calls the module and computes the score? I think we can rely on passing lambdas around much more in Elixir. In Python it is not as common because they are expected to be single line.

josevalim:
> Ok, so we must change the function to cut the rest of the data.

Should we make leftover an option with default of discard?

msluszniak:
> I would make it so cross_validation is only the stream too:
> 
> ```
>   def cross_validation(x, y, k) do
>     Stream.zip([k_fold_split(x, k), k_fold_split(y, k)])
>   end
> ```
> 
> And for score:
> 
> ```
>   def cross_validation_score(x, y, k, fun) do
>     cross_validation(x, y, k)
>     |> Enum.map(fun)
>     |> Nx.stack()
>     |> Nx.mean(axes: [0])
>   end
> ```
> 
> Where fun calls the module and computes the score? I think we can rely on passing lambdas around much more in Elixir. In Python it is not as common because they are expected to be single line.

I really like this approach. It is elegant, user can compute as many metrics as he or she wants. The only thing I'm thinking of is the name of the first function. It'll suggest that it computes the `cv`, but it only does the splitting.

josevalim:
Sure, please go ahead with your preferences and then we can discuss them in a PR :)

josevalim:
For the name, cross_split or cross_partition or cross_fold would work? And the second one can be cross_validation itself?

lib/scholar/model_selection/k_fold.ex
```diff
@@ -0,0 +1,123 @@
...
+          [1.5700000524520874, 1.2149654626846313, 0.004999990575015545],
+          [1.100000023841858, 1.0735294818878174, 0.04999995231628418]
+        ]
+      >
+  """
+  def cross_validation(x, y, k, fun) do
+    cross_fold(x, y, k)
+    |> Enum.map(fun)
+    |> Enum.map(&Nx.stack/1)
+    |> Nx.stack(axis: 1)
```
josevalim, line 121: Should this be stack or mean?

msluszniak: I decided to return all values for each fold. This is more detailed, sklearn also implements it that way, and it's easy to call Nx.mean(cv, axes: [0]) if we want means of a specific metric.

josevalim:
I like everything here. I just have some questions about how the API should be exposed. I assume cross validation is a general technique and we have several cross validation techniques. In other words, if we have this function:

```
  def cross_validation(x, y, k) do
    Stream.zip([k_fold_split(x, k), k_fold_split(y, k)])
  end
```

Which we could pass the function as argument:

```
  def cross_validation(x, y, fun) do
    Stream.zip([fun.(x), fun.(y)])
  end
```

Which means we need a single cross validation function:

Which we could pass the function as argument:

```
  def cross_validation(x, y, folding_fun, scoring_fun) do
    Stream.zip([folding_fun.(x), folding_fun.(y)])
    |> Enum.map(scoring_fun)
    |> Enum.map(&Nx.stack/1)
    |> Nx.stack(axis: 1)
  end
```

I think we should go in this direction but I am not the biggest fun of passing two anonymous functions around. One approach is two build the whole stack as function compositions:

```
fun =
  ModelSelection.k_fold(k: 5)
  |> ModelSelection.cross_validator(fn {x_train, x_test}, {y_train, y_test} ->
    x_train = Nx.concatenate(x_train)
    y_train = Nx.concatenate(y_train)
    model = Scholar.Linear.LinearRegression.fit(x_train, y_train)
    y_pred = Scholar.Linear.LinearRegression.predict(model, x_test)
    mse = Scholar.Metrics.mean_square_error(y_test, y_pred)
    mae = Scholar.Metrics.mean_absolute_error(y_test, y_pred)
    [mse, mae]
  end)

fun.(x, y)
```

I believe this approach is closer to how Axon composes functions. The big question: are there other types of validators we could use that we could also compose on top of `k_fold` and similar?

josevalim:
Special bonus to the above if we can `Nx.Defn.jit(fun)` before calling it.

josevalim:
> Special bonus to the above if we can Nx.Defn.jit(fun) before calling it.

I think we can jit if we use our own version of to_batched that works with `Nx.Defn` expressions. Basically, we build the indexes ourselves, it should be trivial.

msluszniak:
> The big question: are there other types of validators we could use that we could also compose on top of k_fold and similar?

Well, there are a lot of splitting methods different than `k_fold`, but `cv` (as far as I know) always looks the same. There are also methods that use `cv` to search for the best values of hyper-params (like grid-search, simple iterate through the cartesian product of params).

msluszniak:
> > Special bonus to the above if we can Nx.Defn.jit(fun) before calling it.
> 
> I think we can jit if we use our own version of to_batched that works with `Nx.Defn` expressions. Basically, we build the indexes ourselves, it should be trivial.

Wasn't it what I implemented earlier?

josevalim:
> Wasn't it what I implemented earlier?

I think so, but it doesn't need to be defn related. Basically a Enum.map which then does `tensor[x..y]`. :)

> Well, there are a lot of splitting methods different than k_fold, but cv (as far as I know) always looks the same. There are also methods that use cv to search for the best values of hyper-params (like grid-search, simple iterate through the cartesian product of params).

Can you please provide some examples of how we would implement that? It can be a reference in Python or even a sample implementation in Elixir you can push to your PR. :)

msluszniak:
Ok, here would be a pseudo code of grid search, which realized me some flaws of how we are dealing with `cv`

```python
import numpy as np
import itertools
def grid_search(x, y, k, list_of_params):
  best_score = np.inf
  best_params = None
  best_model = None
  for tuple_params in itertools.product(*list_of_params):
    clf_or_rgs = Classifier_of_regressor.new(tuple_params)
    score = Cross_validation(clf_or_rgs, x, y, k, tuple_params)
    score = np.mean(scores)
    if score < best_score:
      best_score = score
      best_params = tuple_params
      best_model = clf_or_rgs 
  return best_model, best_score, best_params
```

our cross-validation needs to be easily modified based on hyperparams values. So we need to rather create a general interface for the train and the validate loop.

josevalim:
I don't think we have issues. We can just build the whole cross validation in the loop no? We replace:

    score = Cross_validation(clf_or_rgs, x, y, k, eval_metric)

by:

```elixir
cls_or_rgs = ...
scope =
  ModelSelection.k_fold(k: 5)
  |> ModelSelection.cross_validate(fn {x_train, x_test}, {y_train, y_test} ->
    x_train = Nx.concatenate(x_train)
    y_train = Nx.concatenate(y_train)
    model = cls_or_rgs.fit(x_train, y_train)
    y_pred = cls_or_rgs.predict(model, x_test)
    mse = Scholar.Metrics.mean_square_error(y_test, y_pred)
    mae = Scholar.Metrics.mean_absolute_error(y_test, y_pred)
    [mse, mae]
  end)
```

no?

msluszniak:
But in grid search we need to have ultimately one metric and we need to pass options to our `sth.fit` function.

msluszniak:
And I believe that sometimes those param search utilities would have their own opts

josevalim:
I may be missing something but I think that would still work. You return a simple score from our cross validation and it is fine? Anyway, if we want to make the `fit` and `predict` replaceable, we need to make sure that all of the `fit` and `receive` functions receive opts as arguments, even if some of them no options will ever be allowed. Shall we do that change in a separate PR?

msluszniak:
> Anyway, if we want to make the `fit` and `predict` replaceable, we need to make sure that all of the `fit` and `receive` functions receive opts as arguments, even if some of them no options will ever be allowed. Shall we do that change in a separate PR?

Yes, that probably will be needed

msluszniak:
> I may be missing something but I think that would still work. You return a simple score from our cross validation and it is fine? 

Actually, it might be easier than I thought 😅. But I would ask: how to generate a Cartesian product of an arbitrary length list of lists?

josevalim:
Perhaps?

```elixir
for {x, i} <- Enum.with_index(list),
    {y, j} <- Enum.with_index(list),
    i != j,
    do: {x, y}
```

msluszniak:
> Perhaps?
> 
> ```elixir
> for {x, i} <- Enum.with_index(list),
>     {y, j} <- Enum.with_index(list),
>     i != j,
>     do: {x, y}
> ```

Nvm, I found an "implementation" of itertools product

```python
def product(*args, repeat=1):
    # product('ABCD', 'xy') --> Ax Ay Bx By Cx Cy Dx Dy
    # product(range(2), repeat=3) --> 000 001 010 011 100 101 110 111
    pools = [tuple(pool) for pool in args] * repeat
    result = [[]]
    for pool in pools:
        result = [x+[y] for x in result for y in pool]
    for prod in result:
        yield tuple(prod)
```

msluszniak:
Ok, so to sum up, what I should do in this PR before I will move to implement grid search?

josevalim:
here is my proposal for an API:

```elixir
defmodule Scholar.ModelSelection do
  def k_fold(x, k) do
  
  end
  
  def cross_validate(x, y, folding_fun, scoring_fun) do
  
  end
end
```

* `k_fold` will emit the ranges instead of using `to_batched` and we should write a test that it works with `defn`
* to avoid confusion, let's make scoring_fun receives two args: `Enum.map(fn {x, y} -> scoring_fun.(x, y) |> Nx.stack() end)`
* in another PR, let's unify the fit+predict APIs

WDYT?

msluszniak:
Ok, looks good!

msluszniak:
About the first point, with ranges, there is a problem because they would be like

for the `k = 3`, `x = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]`
fold 1 -> train: `3..8`, test: `0..2`
fold 2 -> train:  UNION(`0..2`, `6..8`), test: `3..5`
fold 3 -> train: `0..5`, test: `6..8`

The more folds we want, the more unions of disjunctive ranges we have. So we simply cannot call `tensor[start..end]`. More intuitive for me would be creating tensors that represent indices and then call `Nx.take`, but it's tensors dependent.

lib/scholar/model_selection/model_selection.ex
```diff
@@ -0,0 +1,110 @@
...
+      iex> Scholar.ModelSelection.cross_validation_general(x, y, folding_fun, scoring_fun)
+      #Nx.Tensor<
+        f32[2][3]
+        [
+          [1.5700000524520874, 1.2149654626846313, 0.004999990575015545],
+          [1.100000023841858, 1.0735294818878174, 0.04999995231628418]
+        ]
+      >
+  """
+  def cross_validation_general(x, y, folding_fun, scoring_fun) do
```
josevalim, line 105:
```suggestion
  def cross_validate(x, y, folding_fun, scoring_fun) do
```

WDYT?

josevalim:

PR #141, opened 2023-07-20
Add r2_score based on requested changes.

alexbispo:
This PR is a continuation of the PR #72. 

I applied the requested changes.

### DIFF
--- a/lib/scholar/metrics.ex
+++ b/lib/scholar/metrics.ex
@@ -981,13 +981,13 @@ defmodule Scholar.Metrics do
         -Inf
       >
 
-      iex> Scholar.Metrics.r2_score(Nx.tensor([-2.0, -2.0, -2.0]), Nx.tensor([-2.0, -2.0, -2.0 + 1.0e-8]), force_finite: false)
+      iex> Scholar.Metrics.r2_score(Nx.tensor([-2.0, -2.0, -2.0]), Nx.tensor([-2.0, -2.0, -2.0]), force_finite: false)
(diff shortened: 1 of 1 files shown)


lib/scholar/metrics.ex
```diff
@@ -941,6 +942,31 @@ defmodule Scholar.Metrics do
     auc(fpr, tpr)
   end
 
+   @doc ~S"""
```
josevalim, line 945:
```suggestion
  @doc ~S"""
```

lib/scholar/metrics.ex
```diff
@@ -941,6 +942,31 @@ defmodule Scholar.Metrics do
...
+      >
+  """
+  defn r2_score(y_true, y_pred) do
+    check_shape(y_true, y_pred)
+    ssr = squared_euclidean(y_true, y_pred)
+
+    y_mean = Nx.broadcast(Nx.mean(y_true), Nx.shape(y_true))
+    sst = squared_euclidean(y_true, y_mean)
+
+    1 - ssr/sst
```
josevalim, line 967:
```suggestion
    1 - ssr / sst
```

msluszniak:
Thanks for a PR. I think that for now, we don't need to implement `sample_weight` option, because I will make a PR which adds them to all metrics when they will be implemented. However, I think that it is worth to add similar option to sci-kit r2: `force_finite`. What do you think about it?

msluszniak (approved):

josevalim:

PR #143, opened 2023-07-24
Add zero-one loss

msluszniak:


lib/scholar/metrics.ex
```diff
@@ -717,6 +749,41 @@ defmodule Scholar.Metrics do
...
 
+  @doc """
+  Zero-one classification loss.
+
+  # Examples
+
+      iex> y_pred = Nx.tensor([1, 2, 3, 4])
+      iex> y_true = Nx.tensor([2, 2, 3, 4])
+      iex> Scholar.Metrics.zero_one_loss(y_true, y_pred)
+      #Nx.Tensor<
```
polvalente, line 760: Shouldn't this and accuracy support batching? Or are we only doing that for distances still?

msluszniak: It should, but I will add batching to all metrics. For now, we do not support them

polvalente (approved):

lib/scholar/metrics.ex
msluszniak: I'll make a separate issue now, to not forget about it.

PR #151, opened 2023-07-25
Change the structure of metrics.ex

### LINKED ISSUE #150: Change the Structure of Files with Metrics
My idea is to split `lib/scholar/metrics.ex` into `lib/scholar/metrics/classification.ex` and `lib/scholar/metric/regression.ex`. This is because I think that the current structure:

```
/lib/scholar 
| - ...
| - metrics.ex
| - /metrics
    | - distance.ex
    | - clustering.ex
    | - similarity.ex
```                
might be ambiguous because users might not know what is inside metrics.ex when there is a directory called metrics. If we break apart `metrics.ex` into `classification.ex` and `regression.ex`, we will get rid of this ambiguity. The `metrics.ex` is also bulky, so it might be helpful to improve the readability.

```
/lib/scholar 
| - ...
| - /metrics
    | - distance.ex
    | - clustering.ex
    | - similarity.ex
    | - regression.ex
    | - classification.ex
```

msluszniak:
Closes #150

### DIFF
--- a/lib/scholar/metrics/classification.ex
+++ b/lib/scholar/metrics/classification.ex
@@ -1,6 +1,6 @@
-defmodule Scholar.Metrics do
+defmodule Scholar.Metrics.Classification do
   @moduledoc """
-  Metric functions.
+  Classification Metric functions.
 
   Metrics are used to measure the performance and compare
   the performance of any kind of classifier in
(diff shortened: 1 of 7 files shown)


josevalim (approved):

josevalim:
Feel free to also change the guides in the same PR (or a separate one) and merge!

PR #153, opened 2023-08-02
Add Dice coefficient and upgrade Jaccard similarity

### LINKED ISSUE #149: Add Distances
As mentioned in #122 we also want to add all distance metrics. Here is a list that we should probably implement:

* [ ] Sorensen-Dice

It's very little, so adding some proposals in comments to this issue is recommended. For example, we may add kernels like linear kernel or RBF kernel and so on.

msluszniak:
Closes #149

I added the Dice coefficient and support for the option `axis` in Jaccard Similarity.

### DIFF
--- a/lib/scholar/metrics/similarity.ex
+++ b/lib/scholar/metrics/similarity.ex
@@ -6,15 +6,34 @@ defmodule Scholar.Metrics.Similarity do
   import Nx.Defn
   import Scholar.Shared
 
-  @doc ~S"""
+  general_schema = [
+    axis: [
+      type: {:custom, Scholar.Options, :axis, []},
+      doc: """
+      Axis to calculate the distance over. By default the distance
+      is calculated between the whole tensors.
+      """
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(general_schema)
+
+  @doc """
   Calculates Jaccard similarity (also known as Jaccard similarity coefficient, or Jaccard index).
 
   Jaccard similarity is a statistic used to measure similarities between two sets. Mathematically, the calculation
   of Jaccard similarity is the ratio of set intersection over set union.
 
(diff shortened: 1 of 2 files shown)


lib/scholar/metrics/similarity.ex
```diff
@@ -6,15 +6,34 @@ defmodule Scholar.Metrics.Similarity do
   import Nx.Defn
   import Scholar.Shared
 
-  @doc ~S"""
+  general_schema = [
+    axis: [
+      type: {:custom, Scholar.Options, :axis, []},
+      doc: """
+      Axis to calculate the distance over. By default the distance
```
polvalente, line 13:
```suggestion
      Axis over which the distance will be calculated. By default the distance
```

lib/scholar/metrics/similarity.ex
```diff
@@ -34,29 +53,59 @@ defmodule Scholar.Metrics.Similarity do
         0.25
       >
 
+      iex> x = Nx.iota({2,3})
```
polvalente, line 56: I think if you write the tensor explicitly it's easier to visualize the calculation

msluszniak: Ok

msluszniak:
```suggestion
      iex> x = Nx.tensor([[0, 1, 2], [3, 4, 5]])
```

josevalim (approved):

msluszniak:
@josevalim Are branch protection rules configured correctly? Both workflows are waiting idle.

polvalente:
@msluszniak I don't really know what happened there. Branch protection rules were fine; your actions were running correctly.
I checked out on your branch locally and did a `commit --amend + push -f` to resolve this.

PR #158, opened 2023-08-17
Incorporate newer API for plots into guides

msluszniak:


notebooks/k_means.livemd
```diff
@@ -1237,42 +1233,35 @@ silhouette_scores =
...
+null
```
josevalim, line 1264: Should this be null?!

josevalim (approved):

notebooks/k_means.livemd
```diff
@@ -5,13 +5,14 @@
 ```elixir
 Mix.install([
   {:scholar, github: "elixir-nx/scholar", override: true},
-  {:exla, "~> 0.5.3"},
-  {:nx, github: "elixir-nx/nx", sparse: "nx", override: true},
+  {:exla, "~> 0.6.0"},
+  {:nx, "~> 0.6.0", override: true},
```
polvalente, line 9: what needs this override?

notebooks/k_means.livemd
```diff
@@ -28,6 +29,7 @@ The main purpose of this livebook is to introduce the **KMeans** clustering algo
 
 ```elixir
 alias VegaLite, as: Vl
+alias Vl.Data
```
polvalente, line 32:
```suggestion
alias VegaLite.Data
```

for aliasing, unless I'm aliasing a really long module, I'd rather avoid using the alias instead of the fully qualified name for module prefixes

notebooks/k_means.livemd
msluszniak:
```
Because your app depends on explorer ~> 0.6.1 which depends on nx ~> 0.4.0 or ~> 0.5.0, nx ~> 0.4.0 or ~> 0.5.0 is required.
So, because your app depends on nx ~> 0.6.0, version solving failed.
```

notebooks/k_means.livemd
msluszniak: Ok, that makes sense. I'll push changes.

notebooks/k_means.livemd
polvalente: Makes sense. Though maybe if explorer is ready, allowing for more recent Nx could be nice

josevalim: We will
Ship new
Version soon

msluszniak: I'll change this as soon as the new version of Explorer will be released

PR #161, opened 2023-08-24
Add initial version of svm

### LINKED ISSUE #134: Add SVM

msluszniak:
Closes #134


lib/scholar/linear/svm.ex
```diff
@@ -0,0 +1,239 @@
+defmodule Scholar.Linear.SVM do
+  @moduledoc """
+  SVM in both binary and multinomial variants for classification and regression.
```
josevalim, line 3: We should add a note for now about our implementation and how it is limited. Any references are also welcome!

test/scholar/linear/svm_test.exs
```diff
@@ -0,0 +1,23 @@
...
+  use Scholar.Case, async: true
+  alias Scholar.Linear.SVM
+  doctest SVM
+
+  test "Iris Data Set - multinomial classification svm test" do
+    Nx.Defn.default_options(compiler: EXLA)
+
+    {x_train, x_test, y_train, y_test} = Datasets.get(:iris) |> Nx.backend_transfer(EXLA.Backend)
+    y_train = Nx.dot(y_train, Nx.iota({3}, backend: EXLA.Backend))
+    y_test = Nx.dot(y_test, Nx.iota({3}, backend: EXLA.Backend))
```
msluszniak, line 11: This module doesn't exist anymore and should be replaced with the function that was previously in this PR

lib/scholar/linear/svm.ex
msluszniak: Is the note that I pushed okay?

josevalim: Yes!

msluszniak:
This score of 96.7% accuracy that we have now is rather cherry-picked and does not represent the real accuracy of the model. Currently, the model is very dependent on the value of the `:margin` parameter which is hard to tune in general. I don't know how to deal with this.

polvalente:
> This score of 96.7% accuracy that we have now is rather cherry-picked and does not represent the real accuracy of the model. Currently, the model is very dependent on the value of the `:margin` parameter which is hard to tune in general. I don't know how to deal with this.

I don't think this is an issue. You're using too little data to actually get any conclusions on how hard the model is to train.

Besides that, in ML having a model being highly dependent in one or 2 hyperparams is not _that_ unusual.

lib/scholar/linear/svm.ex
```diff
@@ -0,0 +1,250 @@
...
+
+            bias = Polaris.Updates.apply_updates(bias, bias_updates)
+
+            has_converged_j = Nx.sum(Nx.abs(loss)) < Nx.axis_size(x, 0) * eps
+
+            has_converged =
+              Nx.indexed_put(
+                has_converged,
+                Nx.new_axis(Nx.new_axis(j, -1), -1),
+                Nx.new_axis(has_converged_j, -1)
```
polvalente, line 206: I think we can use this now:
```suggestion
                Nx.new_axis(j, -1),
                has_converged_j
```

lib/scholar/linear/svm.ex
```diff
@@ -0,0 +1,250 @@
...
+    %__MODULE__{
+      coefficients: final_coef,
+      bias: final_bias
+    }
+  end
+
+  defnp loss_and_grad(coeff, bias, xs, ys, c, margin) do
+    value_and_grad({coeff, bias}, fn {coeff, bias} ->
+      0.5 * Nx.sum(coeff ** 2, axes: [-1]) +
+        c * Nx.sum(Nx.max(0, margin - (Nx.dot(xs, Nx.transpose(coeff)) + bias) * ys), axes: [-1])
```
polvalente, line 229: One concern I have with this loss being hardcoded is that it already enforces regularization, which should be delegated to the optimizer.

Maybe we could have the loss being an optional arbitrary arity-2 function instead, receiving y_true and y_pred as arguments.

test/scholar/linear/svm_test.exs
```diff
@@ -0,0 +1,186 @@
+defmodule Scholar.Linear.SVMTest do
+  use Scholar.Case, async: true
+  alias Scholar.Linear.SVM
+  doctest SVM
+
+  defp iris_data do
```
polvalente, line 6: I'm not sure this data should be kept here. Isn't this dataset used elsewhere too?

test/scholar/linear/svm_test.exs
```diff
@@ -0,0 +1,186 @@
...
+    y = Nx.take(y, shuffle)
+    {x_train, x_test} = Nx.split(x, 120)
+    {y_train, y_test} = Nx.split(y, 120)
+    {x_train, x_test, y_train, y_test}
+  end
+
+  test "Iris Data Set - multinomial classification svm test" do
+    {x_train, x_test, y_train, y_test} = iris_data()
+
+    model = SVM.fit(x_train, y_train, num_classes: 3, margin: 150)
```
polvalente, line 178: keeping in with the parameterized loss: `:margin` and `:c` should be loss-specific opts instead, generally enclosed in a function capture: `SVM.hinge_loss(&1, &2, c: c, margin: margin)`

test/scholar/linear/svm_test.exs
```diff
@@ -0,0 +1,186 @@
...
+    res = SVM.predict(model, x_test)
+
+    accuracy = Scholar.Metrics.Classification.accuracy(res, y_test)
+    IO.inspect(accuracy)
```
polvalente, line 182:
```suggestion
```

lib/scholar/linear/svm.ex
```diff
@@ -0,0 +1,250 @@
...
+      required: true,
+      type: :pos_integer,
+      doc: "number of classes contained in the input tensors."
+    ],
+    iterations: [
+      type: :pos_integer,
+      default: 1000,
+      doc: """
+      number of iterations of gradient descent performed inside logistic
+      regression.
```
polvalente, line 33: copy-paste here

lib/scholar/linear/svm.ex
msluszniak: I'm okay with that but one question then: What would be a default function? @josevalim

test/scholar/linear/svm_test.exs
msluszniak: Yes, it is also used in logistic regression. I may move both data and splitting function into `test_helper.exs`, WDYT?

msluszniak: Actually, there is a `test/support` directory, so I'll add that there ;)

test/support/scholar_case.ex
```diff
@@ -25,4 +25,173 @@ defmodule Scholar.Case do
       """)
     end
   end
+
+  def iris_data do
+    key = Nx.Random.key(42)
+
+    x =
+      Nx.tensor([
```
josevalim, line 33: Btw, the `~M` sigil can help a lot in writing large matrices like this. :)

josevalim (approved):

lib/scholar/linear/svm.ex
```diff
@@ -0,0 +1,257 @@
+defmodule Scholar.Linear.SVM do
+  @moduledoc """
+  SVM classifier
+
+  It uses the OvR strategy to handle both binary and multinomial classification.
+  This implementation uses stochastic gradient descent from default or any other optimizer
+  available in `Polaris`. This makes it similar to a sklearn SGDClassifier [1].
+  It means that on average it will work slower than algorithms that use QP and kernel trick (LIBSVM [2]) or
+  Coordinate Descent Algorithm (LIBLINEAR [3]). It also cannot use different kernels like in LIBSVM,
+  but you can use any type of optimizer and regularizer available in `Polaris`.
```
polvalente, line 10: Polaris doesn't have a separate concept of regularizers. Perhaps this should be removed.
```suggestion
  but you can use any type of optimizer available in `Polaris`.
```

lib/scholar/linear/svm.ex
```diff
@@ -0,0 +1,257 @@
...
+  defnp loss_and_grad(coeff, bias, xs, ys, loss_fn) do
+    value_and_grad({coeff, bias}, fn {coeff, bias} ->
+      y_pred = predict(coeff, bias, xs)
+      loss_fn.(y_pred, ys)
+    end)
+  end
+
+  defnp predict(coeff, bias, xs) do
+    # Nx.dot(xs, [1], coeff, [1]) + bias
+    Nx.dot(xs, Nx.transpose(coeff)) + bias
```
polvalente, line 224:
```suggestion
    Nx.dot(xs, [-1], coeff, [-1]) + bias
```

Why not use the dot/4 trick here?

lib/scholar/linear/svm.ex
```diff
@@ -0,0 +1,257 @@
...
+    margin = opts[:margin]
+    c * Nx.sum(Nx.max(0, margin - y_pred) * ys, axes: [-1])
+  end
+
+  # defnp loss_and_grad(coeff, bias, xs, ys, c, margin) do
+  #   value_and_grad({coeff, bias}, fn {coeff, bias} ->
+  #     0.5 * Nx.sum(coeff ** 2, axes: [-1]) +
+  #       c * Nx.sum(Nx.max(0, margin - (Nx.dot(xs, Nx.transpose(coeff)) + bias) * ys), axes: [-1])
+  #   end)
+  # end
```
polvalente, line 238:
```suggestion
```

lib/scholar/linear/svm.ex
```diff
@@ -0,0 +1,257 @@
...
+      iex> y = Nx.tensor([1, 0, 1])
+      iex> model = Scholar.Linear.SVM.fit(x, y, num_classes: 2)
+      iex> Scholar.Linear.SVM.predict(model, Nx.tensor([[-3.0, 5.0]]))
+      #Nx.Tensor<
+        s64[1]
+        [1]
+      >
+  """
+  defn predict(%__MODULE__{coefficients: coeff, bias: bias}, x) do
+    score = Nx.dot(x, [1], coeff, [1]) + bias
```
polvalente, line 254:
```suggestion
    score = predict(coeff, bias, x)
```

lib/scholar/options.ex
```diff
@@ -27,6 +27,22 @@ defmodule Scholar.Options do
     end
   end
 
+  def loss_function(function) do
+    case function do
+      function when is_function(function, 2) ->
+        {:ok, function}
+
+      nil ->
+        loss_fn = &Scholar.Linear.SVM.hinge_loss(&1, &2, c: 1.0, margin: 10)
```
polvalente, line 36: This shouldn't be in a general Scholar.Options module, because it's specific to SVM

polvalente (commented):
minor comments, but this overall looks good!

josevalim (approved):
Paulo told me to approve it :D

PR #164, opened 2023-08-27
Draft: Adding CountVectorizer & TfidfVectorizer

acalejos:
Adding CountVectorizer and TFIDF Vectorizer very similar to SKLearn's implementation. Based on some work currently being done by @polvalente on Nx to potentially add sparse tensor support (see https://github.com/elixir-nx/nx/tree/pv-proof-of-concept-sparse-torchx) , this would definitely benefit from sparse tensors, but for now I plan on proceeding with dense tensors.

Considering these vectorizers operate on text, perhaps they would best be fit for a standalone NLP library (I have begun scaffolding out a bit of one) but they could also very well fit in here in Scholar.Preprocessing. I'm open to suggestions as to where they best belong. 

As of now, I only have finished setting up the options, and building a vocab given the options. I will try to wrap up at least the CountVectorizer by tomorrow night.

### DIFF
--- a/lib/scholar/preprocessing/count_vectorizer.ex
+++ b/lib/scholar/preprocessing/count_vectorizer.ex
@@ -1,0 +1,60 @@
+defmodule Scholar.Preprocessing.CountVectorizer do
+  defstruct vocabulary: nil,
+            fixed_vocabulary: false,
+            ngram_range: {1, 1},
+            max_features: nil,
+            min_df: 1.0,
+            max_df: 1.0,
+            stop_words: nil,
+            binary: false
+
+  defp make_ngrams(tokens, ngram_range) do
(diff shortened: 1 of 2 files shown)


josevalim:
@acalejos I would propose to keep this in the NLP library for now. For Scholar, we have been focusing on keeping everything implemented with `defn` and this is plain not possible here. Also, I can see future interest in implementing this in native code for performance, so I am not even sure if this will survive after some iterations :)

acalejos:
> @acalejos I would propose to keep this in the NLP library for now. For Scholar, we have been focusing on keeping everything implemented with `defn` and this is plain not possible here. Also, I can see future interest in implementing this in native code for performance, so I am not even sure if this will survive after some iterations :)

Makes sense. Appreciate the feedback!

PR #170, opened 2023-09-01
Isotonic regression

### LINKED ISSUE #162: Add Isotonic Regression

msluszniak:
Closes #162 

I need to add tests


mix.exs
```diff
@@ -30,7 +30,8 @@ defmodule Scholar.MixProject do
   defp deps do
     [
       {:ex_doc, "~> 0.30", only: :docs},
-      {:nx, "~> 0.6", override: true},
+      # {:nx, "~> 0.6", override: true},
+      {:nx, github: "elixir-nx/nx", sparse: "nx", override: true},
```
josevalim, line 34: We can continue using 0.6, no?

lib/scholar/linear/isotonic_regression.ex
```diff
@@ -0,0 +1,419 @@
...
+
+    {y, _} =
+      while {y, {target, i = 0, _k = 0, max_size}}, i < max_size + 1 do
+        k = target[i] + 1
+        indices = Nx.iota({Nx.axis_size(y, 0)})
+        in_range? = Nx.logical_and(i + 1 <= indices, indices < k)
+        y = Nx.select(in_range?, y[i], y)
+        i = k
+        {y, {target, i, k, max_size}}
+      end
```
josevalim, line 415: The big question is if any of these loops can be vectorized. Thoughts? :)

lib/scholar/linear/isotonic_regression.ex
```diff
@@ -0,0 +1,419 @@
...
+            [1.0, -1.0],
+            [1.0, -1.0]
+          ]
+        ),
+        x: Nx.tensor(
+          [1.0, 4.0, 7.0, 9.0, 10.0]
+        )
+      }
+  """
+  def build_function(model) do
```
josevalim, line 210: What would this function be used for? The docs says what it does, but not when/why/how we use it. :)

mix.exs
msluszniak: Not really, since in the main there is a fix for `Nx.clip` bug with infinities. In Isotonic regression, this is a default case, unfortunately.

lib/scholar/linear/isotonic_regression.ex
msluszniak: Actually, I thought about it. But the main problem is that in each iteration, we use data from the previous iteration which is problematic to vectorize. @polvalente @jonatanklosko, maybe you would have some thoughts?

lib/scholar/linear/isotonic_regression.ex
msluszniak: This function is essential because based on thresholds that are calculated in isotonic regression it creates a linear interpolation model that generates the results of actual predictions. It is called after fit function when we have all results from isotonic regression.

mix.exs
josevalim: I see!

lib/scholar/linear/isotonic_regression.ex
josevalim: Ah, and if I understand this right, this function cannot be used as `defn` right? We had other models that had some code that could not be executed inside `defn` right? Do you remember which model it was? Maybe we should copy the same structure that we have there?

msluszniak: I write it as a `def` because through the whole algorithm, I propagate tensors that normally would have been reduced to smaller tensors (in `make_unique`). So at the very end, I cut them and pass them to `Scholar.Interpolation.Linear`. I think that we have some functions in `Scholar.Metrics.Classification` like `roc_curve` which use `def`.

josevalim: Let me rephrase the question: why don't we do this inside `fit` then?

josevalim: Oh, I found it. it was AffinityPropagation, where we added an optional `prune` function. Could we do something similar here? You can optionally prune but, if you don't, we run with the whole tensor?

msluszniak: Hmm, If we pass unpruned tensor to interpolation, then the results would be wrong. The problem with fit and this functionality is that the cutoff index is calculated based on the content of the tensor. So we simply would need a dynamic shape to support this. I would think if we could somehow force users to preprocess data before so we don't need to prune it

msluszniak: Okay, I think that we can preprocess the data so we can put this interpolation inside defn, but to be honest this solution is worse than what we have now. Because it will introduce another `def` function that will change the shapes of tensors. 

The solution will be the following: For each data point from x: if there are some indices for which values in `X` are the same, take a weighted (by sample weights) mean of those values, remove all repetitions from X and the according values from `y` and replace them with the calculated mean.

lib/scholar/linear/isotonic_regression.ex
```diff
@@ -0,0 +1,419 @@
...
+    cutoff = Nx.to_number(model.cutoff_index)
+    x = model.x_thresholds[0..cutoff]
+    y = model.y_thresholds[0..cutoff]
+
+    if Nx.size(y) == 1 do
+      Nx.broadcast(y, Nx.shape(x))
```
josevalim, line 216: Will this clause work? Can we pass a tensor as a function to `Scholar.Interpolation.Linear.predict`?

lib/scholar/linear/isotonic_regression.ex
josevalim: Hrm, what do you think about this approach.

We add a new field called `preprocess: nil | Scholar.Interpolation.Linear.t`. You must call `preprocess` before you call `predict` to fill in the preprocess field. In `predict`, if `preprocess` is nil, we raise.

The goal is to keep everything in a single struct. Thoughts?

lib/scholar/linear/isotonic_regression.ex
msluszniak: Good catch, now it won't work. It should be called instead of interpolation `predict` in `Scholar.Linear.Isotonic.predict`

lib/scholar/linear/isotonic_regression.ex
msluszniak: Sure, it will be convenient to have it in one place :)

lib/scholar/linear/isotonic_regression.ex
jonatanklosko: Nothing stands out to me, the iterations seem to be sequentially dependent because of the chain lookup in `target`.

polvalente: But target isn't updated, so maybe it's possible

msluszniak: Target is updated in every loop of inner while which implies that is updated on every loop of outer while as well

jonatanklosko: It's not updated in this specific loop though :) @polvalente The thing is that for second iteration we have `i = target[0] + 1`, for third one we have `target[target[0] + 1] + 1`, for forth `target[target[target[0] + 1] + 1] + 1` and so on. I don't see how we could compute `i` vector in a single step.

polvalente: Ah, that makes more sense. I also don't see how it could be vectorized

msluszniak: Oh, ok, I forgot that we have 3 while loops :sweat_smile:

lib/scholar/linear/isotonic_regression.ex
msluszniak: @josevalim I pushed changes. I changed the `preprocess` default value from `nil` to `{}` cause the protocol container is not implemented for `nil`. I added tests. The only thing I need to add is support for decreasing sequences.

lib/scholar/linear/isotonic_regression.ex
```diff
@@ -0,0 +1,492 @@
...
+            model.x_thresholds,
+            model.y_thresholds
+          )
+    }
+  end
+
+  deftransform check_preprocess(model) do
+    if model.preprocess == {} do
+      raise ArgumentError, "Model has not been preprocessed. Please call
+      `preprocess` on the model before calling `predict`."
```
josevalim, line 293:
```suggestion
      raise ArgumentError,
            "model has not been preprocessed. " <>
              "Please call preprocess/1 on the model before calling predict/2"
```

josevalim (approved):

msluszniak:
Okay, I just need to add one more feature to this PR. In `MDS` it would be necessary to support option `:auto` in `:increasing?`. In `Sklearn` they function called `Spearman correlation` that depends on `t_test` that depends on a special integral called `Student t distribution cumulative distribution function`, that depends on `Incomplete Beta Integral` and the chain is much much longer. So instead of implementing all of those, we would use `Linear Regression` to determine the monotonicity of the data from `Isotonic Regression`

lib/scholar/linear/isotonic_regression.ex
```diff
@@ -495,4 +505,11 @@ defmodule Scholar.Linear.IsotonicRegression do
 
     if opts[:increasing?], do: y, else: Nx.reverse(y)
   end
+
+  deftransform check_increasing(x, y) do
+    x = Nx.new_axis(x, -1)
+    y = Nx.new_axis(y, -1)
+    model = Scholar.Linear.LinearRegression.fit(x, y)
+    Nx.greater_equal(model.coefficients[0][0], 0) == Nx.u8(1)
```
josevalim, line 513: This is not going to work if you call isotonic_regression from within a defn, because all you have is expressions, never values. You can try by doing this:

```elixir
defn foo do
  call_isotonic_stuff(...)
end
```

It will always be false. I think what you will have to do is to keep this as `defnp`. Then increasing? should either be 0, 1, or the result of this expression, and you pass increasing as a tensor to fit_n, rather than an option.

josevalim: We need to remove the comparison from here:

```suggestion
    Nx.greater_equal(model.coefficients[0][0], 0)
```

lib/scholar/linear/isotonic_regression.ex
```diff
@@ -0,0 +1,523 @@
...
+
+    sample_weights =
+      if Nx.is_tensor(sample_weights),
+        do: Nx.as_type(sample_weights, x_type),
+        else: Nx.tensor(sample_weights, type: x_type)
+
+    sample_weights = Nx.broadcast(sample_weights, {Nx.axis_size(y, 0)})
+
+    increasing? =
+      case opts[:increasing?] do
```
josevalim, line 165: Given it is no longer a boolean only, let's rename it to simply `:increasing`.

josevalim (approved):

lib/scholar/linear/isotonic_regression.ex
```diff
@@ -0,0 +1,523 @@
...
+        x_max: Nx.tensor(
+          11.0
+        ),
+        x_thresholds: Nx.tensor(
+          [1.0, 4.0, 7.0, 9.0, 10.0, 11.0]
+        ),
+        y_thresholds: Nx.tensor(
+          [1.0, 3.0, 6.0, 8.0, 9.0, 10.0]
+        ),
+        increasing?: Nx.u8(1),
```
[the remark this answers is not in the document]
msluszniak: Yes, we should store this information since for `:auto`, we might not know if is it either decreasing or increasing.

PR #172, opened 2023-09-04
Fix memory usage in Affinity Propagation

### LINKED ISSUE #171: Memory use for Affinity Propagation seems high, OOMing
Hey! I'm running into an issue where I'm OOMing with Affinity Propagation w/ 5868 samples. I understand that memory use will be n^2 for the affinity matrix, but w/ f32 that should be 5868^2*4=137_733_696 bytes, no? Certainly shouldn't be blowing up the show. I snapped this before it crashed:
Screenshot 2023-09-04 at 08 32 25

Here's a Livebook to replicate:
https://gist.github.com/cigrainger/4f2531aa84cd8e1b3593d6c718a4d1ae

For reference, with sklearn:
image

I assume there are plenty of optimisations in sklearn and comparisons may be a bit futile. Just not sure if there's a memory leak here or something. Thanks for the hard work on this lib and for any help!

Oh and if it matters: this is on an M1 Mac w/ 16gb ram.

msluszniak:
Closes #171


lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -307,14 +308,12 @@ defmodule Scholar.Cluster.AffinityPropagation do
...
+    n = Nx.axis_size(data, 0)
     self_preference = opts[:self_preference]
-    t1 = Nx.reshape(data, {1, n, dims}) |> Nx.broadcast({n, n, dims})
-    t2 = Nx.reshape(data, {n, 1, dims}) |> Nx.broadcast({n, n, dims})
 
-    dist =
-      (-1 * Scholar.Metrics.Distance.squared_euclidean(t1, t2, axes: [-1]))
-      |> Nx.as_type(to_float_type(data))
+    norm1 = Nx.sum(data ** 2, axes: [1], keep_axes: true)
+    norm2 = Nx.sum(data ** 2, axes: [1], keep_axes: true) |> Nx.transpose()
```
jonatanklosko, line 315:
```suggestion
    norm2 = Nx.transpose(norm1)
```

jonatanklosko (approved):
Nice!

lib/scholar/cluster/affinity_propagation.ex
```diff
@@ -307,14 +308,12 @@ defmodule Scholar.Cluster.AffinityPropagation do
...
+    n = Nx.axis_size(data, 0)
     self_preference = opts[:self_preference]
-    t1 = Nx.reshape(data, {1, n, dims}) |> Nx.broadcast({n, n, dims})
-    t2 = Nx.reshape(data, {n, 1, dims}) |> Nx.broadcast({n, n, dims})
 
-    dist =
-      (-1 * Scholar.Metrics.Distance.squared_euclidean(t1, t2, axes: [-1]))
-      |> Nx.as_type(to_float_type(data))
+    norm1 = Nx.sum(data ** 2, axes: [1], keep_axes: true)
+    norm2 = Nx.transpose(norm1)
```
josevalim, line 315: Is is squared_euclidean so expensive? Is there something we should do about it?

msluszniak: In this specific case, it is very expensive because we need to create first a tensor of dim `{num_samples, num_sample, num_features}`. So I wrote a formula that calculates it without that expensive broadcasting. Is it something we should do about it? Hmm, rather no. Our `squared_euclidean` is more general than calculating only pairwise distance. What we can do is actually support pairwise distances in general. They can be calculated using a general formula:
image
(This one is actually for squared distance, but it can be simply wrapped into sqrt). This `|| x ||` is literally any norm and because all norms induces metrics (all normed vector spaces are metric spaces) we can implement any pairwise distance this way :)

josevalim: Excellent, thank you!

msluszniak: @josevalim do we plan to add pairwise distances? If yes, then I'll create an issue.

josevalim: Your call!

josevalim (approved):

PR #176, opened 2023-09-09
Fix notebook markup and improvements

santiago-imelio:
- Use correct markup on hyperlinks
- Pass only training data to the grid search
- Add final evaluation on test set


josevalim:

PR #177, opened 2023-09-12
Pairwise distances

msluszniak:


lib/scholar/manifold/tsne.ex
```diff
@@ -167,18 +167,21 @@ defmodule Scholar.Manifold.TSNE do
...
       :euclidean ->
         Distance.euclidean(t1, t2, axes: [2])
+        # Distance.pairwise_euclidean(x)
 
       :manhattan ->
         Distance.manhattan(t1, t2, axes: [2])
 
       :cosine ->
         Distance.cosine(t1, t2, axes: [2])
+        # Distance.pairwise_cosine(x)
```
msluszniak, line 184: I have no idea why using pairwise distances causes the results to be completely different than those from broadcasting and calculating distances by hand. I tested this function, and it returns the same results for the same input value.

josevalim (approved):

lib/scholar/manifold/tsne.ex
josevalim: Try doing `print_expr` and see if the expressions are the same or different?

msluszniak: Okay, I discovered that after 10 iterations of tsne algorithms, both functions gives the same results. After 50, there is a small discrepancy in distances, and after 500 iterations (default) there are completely different results.

msluszniak:
I tested TSNE with pairwise distances and despite results from pairwise and classical distances generated results with different orders of magnitude, it seems to work. I plotted the results for Iris, and they are reasonable. For now, I changed the test so they are passing, but we may think of more sophisticated tests that will pass if embedded space is "good enough".

josevalim (approved):

PR #178, opened 2023-09-13
Deprecate covariance module

msluszniak:
I'm wondering if we should support the argument `mean` as in `Nx.covariance`. We couldn't handle it within one function like in `Nx.covariance` using guards. So, I think that introducing another function which deals with mean would not be elegant. But maybe it's worth it I'm open to suggestions :)


lib/scholar/covariance.ex
```diff
@@ -27,6 +27,7 @@ defmodule Scholar.Covariance do
   @opts_schema NimbleOptions.new!(opts)
 
   @doc """
+  This function is deprecated and will be removed in the future. Use `Nx.covariance/2` instead.
```
josevalim, line 30: The correct way of deprecating is:

```
@deprecated "Use Nx.convariance/2 instead"
```

The compiler will also warn when people use the other ones.

josevalim (approved):
> We couldn't handle it within one function like in Nx.covariance using guards.

That is not possible even when using transform? Anyway, once the correct deprecation syntax is used, it looks good to me. :)

msluszniak:
I added the correct deprecate messages. I also remove opts with `ddof` options since it doesn't have any sense in correlation. Result for both unbiased covariance estimator and unbiased variances in denominator results with exactly the same output.

PR #179, opened 2023-09-14
Standard Scaler fit-transform interface

santiago-imelio:
There is already `Preprocessing.standard_scale` that performs this transformation. However, with this function we cannot standardize a test set using the train set distribution data (mean and std deviation).

`Scaler.StandardScaler.fit/2` will return a struct containing the mean and standard deviation of the distribution of learned observations.

### DIFF
--- a/lib/scholar/preprocessing/standard_scaler.ex
+++ b/lib/scholar/preprocessing/standard_scaler.ex
@@ -1,7 +1,26 @@
 defmodule Scholar.Preprocessing.StandardScaler do
+  @moduledoc """
+  Standardizes the tensor by removing the mean and scaling to unit variance.
+
+  #{~S'''
+  Formula for input tensor $x$:
+  $$
+  z = \frac{x - \mu}{\sigma}
+  $$
+  Where $\mu$ is the mean of the samples, and $\sigma$ is the standard deviation.
+  Standardization can be helpful in cases where the data follows
+  a Gaussian distribution (or Normal distribution) without outliers.
+  '''}
+
+  Centering and scaling happen independently on each feature by computing the relevant
+  statistics on the samples in the training set. Mean and standard deviation are then
+  stored to be used on new samples.
+  """
+
   import Nx.Defn
 
-  defstruct [:deviation, :mean]
+  @derive {Nx.Container, containers: [:standard_deviation, :mean]}
+  defstruct [:standard_deviation, :mean]
 
   opts_schema = [
     axes: [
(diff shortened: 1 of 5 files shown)


lib/scholar/scaler/standard_scaler.ex
```diff
@@ -0,0 +1,39 @@
...
+    %__MODULE__{deviation: std, mean: mean_reduced}
+  end
+
+  deftransform transform(tensor, %__MODULE__{deviation: std, mean: mean}) do
+    scale(tensor, std, mean)
+  end
+
+  deftransform fit_transform(tensor, opts \\ []) do
+    scaler = __MODULE__.fit(tensor, opts)
+    __MODULE__.transform(tensor, scaler)
```
josevalim, line 33:
```suggestion
    scaler = fit(tensor, opts)
    transform(tensor, scaler)
```

lib/scholar/scaler/standard_scaler.ex
```diff
@@ -0,0 +1,39 @@
...
+      Axes to calculate the distance over. By default the distance
+      is calculated between the whole tensors.
+      """
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(opts_schema)
+
+  deftransform fit(tensor, opts \\ []) do
+    NimbleOptions.validate!(opts, @opts_schema)
```
josevalim, line 19: We should delegate the rest of this function to a defnp!

lib/scholar/scaler/standard_scaler.ex
josevalim, line 33:
```suggestion
  defn fit_transform(tensor, opts \\ []) do
    scaler = fit(tensor, opts)
    transform(tensor, scaler)
```

lib/scholar/scaler/standard_scaler.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Scholar.Scaler.StandardScaler do
```
josevalim, line 1: Let's put this in the Preprocessing namespace:

```suggestion
defmodule Scholar.Preprocessing.StandardScaler do
```

josevalim:
@msluszniak should we generally convert our preprocessing steps into modules to mirror scikit learn?

msluszniak:
I think we shouldn't convert preprocessing into modules. For functions that calculate some metrics and then apply them to data, it would be easy to make this interface. However, for functions like one-hot encoding or ordinal encoding, we would need to somehow map the relation `key -> calculated_mapping`. In `Scikit` they use dictionaries. One of the functions - `normalize` has the equivalent in `Nx.Linalg.norm`. If we decide to support `standard_scaler`, `max_abs_scaler`, `min_max_scaler`, and `binarizer` then the interface would be inconsistent. If we change the interface to use struct users wouldn't be able to pipe the results of `standard_scale`. But on the other hand, it would be cool to support both.

msluszniak:
What we can actually do is create separate functions like standard_scale(reference_tensor, tensor_to_apply, opts) and support the application of standardization on a different tensor, WDYT?

msluszniak:
And the standard_scale/2 is a simplified version of this because the reference tensor is the same as tensor_to_apply

josevalim:
@msluszniak the downside of this approach is that we would not be able to fit once and apply the same fitting multiple times. Is that a concern?

msluszniak:
You're right, hmm. In a standard use case, this function won't be called multiple times. But if so, then the benefit from modular implementation might be considerable.

santiago-imelio:
@msluszniak From what I understand, it seems too soon to agree on an interface that fits both the language and the needs of developers, so in my opinion keeping the preprocessing API functional seems to be appropriate for now. 

Closing in favor of continuing the discussion.

josevalim:
I have reopened this PR in case you would like to address the comments in it. :)

lib/scholar/preprocessing/standard_scaler.ex
```diff
@@ -0,0 +1,46 @@
...
+
+  defnp fit_n(tensor, opts) do
+    std = Nx.standard_deviation(tensor, axes: opts[:axes], keep_axes: true)
+    mean_reduced = Nx.mean(tensor, axes: opts[:axes], keep_axes: true)
+    mean_reduced = Nx.select(Nx.equal(std, 0), 0.0, mean_reduced)
+
+    {std, mean_reduced}
+  end
+
+  deftransform transform(%__MODULE__{deviation: std, mean: mean}, tensor) do
```
msluszniak, line 33:
```suggestion
  defn transform(%__MODULE__{standard_deviation: std, mean: mean}, tensor) do
```

lib/scholar/preprocessing/standard_scaler.ex
```diff
@@ -0,0 +1,46 @@
+defmodule Scholar.Preprocessing.StandardScaler do
+  import Nx.Defn
+
+  defstruct [:deviation, :mean]
```
msluszniak, line 4:
```suggestion
  defstruct [:standard_deviation, :mean]
```

lib/scholar/preprocessing/standard_scaler.ex
```diff
@@ -0,0 +1,46 @@
...
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(opts_schema)
+
+  deftransform fit(tensor, opts \\ []) do
+    NimbleOptions.validate!(opts, @opts_schema)
+    {std, mean} = fit_n(tensor, opts)
+
+    %__MODULE__{deviation: std, mean: mean}
```
msluszniak, line 22:
```suggestion
    %__MODULE__{standard_deviation: std, mean: mean}
```

msluszniak (commented):
We also need to add docs for `fit`, `transform`, and `fit_transform`

lib/scholar/preprocessing/standard_scaler.ex
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  defnp fit_n(tensor, opts) do
+    std = Nx.standard_deviation(tensor, axes: opts[:axes], keep_axes: true)
+    mean_reduced = Nx.mean(tensor, axes: opts[:axes], keep_axes: true)
+    mean_reduced = Nx.select(Nx.equal(std, 0), 0.0, mean_reduced)
```
msluszniak, line 28:
```suggestion
    mean_reduced = Nx.select(std == 0, 0.0, mean_reduced)
```

lib/scholar/preprocessing/standard_scaler.ex
```diff
@@ -0,0 +1,145 @@
...
+            [1.0657403469085693]
+          ]
+        >,
+        mean: #Nx.Tensor<
+          f32[1][1]
+          [
+            [0.4444444477558136]
+          ]
+        >
+      }
```
msluszniak, line 105:
```suggestion
      iex> t = Nx.tensor([[1, -1, 2], [2, 0, 0], [0, 1, -1]])
      iex> scaler = Scholar.Preprocessing.StandardScaler.fit(t)
      %Scholar.Preprocessing.StandardScaler{
        standard_deviation:  Nx.tensor(
          [
            [1.0657403469085693]
          ]
        ),
        mean:  Nx.tensor(
          [
            [0.4444444477558136]
          ]
        )
      }
```

msluszniak: Generally, if you want to execute a test with a structure as a result, then you need to pass tenors as you would create them in `defn`. So you should replace all `#Nx.Tensor<...>` with `Nx.tensor(..)`

lib/scholar/preprocessing/standard_scaler.ex
```diff
@@ -0,0 +1,145 @@
...
```
msluszniak, line 68: Same case here

msluszniak (approved):

lib/scholar/preprocessing/standard_scaler.ex
```diff
@@ -0,0 +1,138 @@
...
+
+  #{~S'''
+  Formula for input tensor $x$:
+  $$
+  z = \frac{x - \mu}{\sigma}
+  $$
+  Where $\mu$ is the mean of the samples, and $\sigma$ is the standard deviation.
+  Standardization can be helpful in cases where the data follows
+  a Gaussian distribution (or Normal distribution) without outliers.
+  '''}
```
josevalim, line 13:
```suggestion
  @moduledoc ~S"""
  Standardizes the tensor by removing the mean and scaling to unit variance.

  Formula for input tensor $x$:
  
  $$
  z = \frac{x - \mu}{\sigma}
  $$

  Where $\mu$ is the mean of the samples, and $\sigma$ is the standard deviation.
  Standardization can be helpful in cases where the data follows
  a Gaussian distribution (or Normal distribution) without outliers.
```

lib/scholar/preprocessing/standard_scaler.ex
```diff
@@ -0,0 +1,138 @@
...
+  @doc """
+  Compute the standard deviation and mean of samples to be used for later scaling.
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return values
+
+    Returns a struct with the following parameters:
```
josevalim, line 46:
```suggestion
  Returns a struct with the following parameters:
```

lib/scholar/preprocessing/standard_scaler.ex
```diff
@@ -0,0 +1,138 @@
...
+
+  Formula for input tensor $x$:
+  
+  $$
+  z = \frac{x - \mu}{\sigma}
+  $$
+
+  Where $\mu$ is the mean of the samples, and $\sigma$ is the standard deviation.
+  Standardization can be helpful in cases where the data follows
+  a Gaussian distribution (or Normal distribution) without outliers.
```
josevalim, line 13:
```suggestion
  Standardizes the tensor by removing the mean and scaling to unit variance.

  Formula for input tensor $x$:

  $$
  z = \frac{x - \mu}{\sigma}
  $$

  Where $\mu$ is the mean of the samples, and $\sigma$ is the standard deviation.
  Standardization can be helpful in cases where the data follows
  a Gaussian distribution (or Normal distribution) without outliers.
```

PR #185, opened 2023-10-10
F-beta score

0urobor0s:
Used expression:
$$F_\beta = (1 + \beta^2) \cdot \frac{\mathrm{precision} \cdot \mathrm{recall}}{(\beta^2 \cdot \mathrm{precision}) + \mathrm{recall}}$$

From https://en.wikipedia.org/wiki/F-score

### DIFF
--- a/lib/scholar/metrics/classification.ex
+++ b/lib/scholar/metrics/classification.ex
@@ -653,7 +653,8 @@ defmodule Scholar.Metrics.Classification do
     num_classes = check_num_classes(opts[:num_classes])
     average = opts[:average]
 
-    {_precision, _recall, per_class_fscore} = precision_recall_fscore_n(y_true, y_pred, beta, num_classes, average)
+    {_precision, _recall, per_class_fscore} =
+      precision_recall_fscore_n(y_true, y_pred, beta, num_classes, average)
 
     per_class_fscore
   end
@@ -672,7 +673,7 @@ defmodule Scholar.Metrics.Classification do
         {true_positive, false_positive, false_negative}
 
       _ ->
-      {true_positive, false_positive, false_negative}
+        {true_positive, false_positive, false_negative}
     end
   end
 
@@ -688,8 +689,10 @@ defmodule Scholar.Metrics.Classification do
         # Should only be +Inf
         Nx.is_infinity(beta) ->
           recall
+
         beta == 0 ->
           precision
+
         true ->
           beta2 = Nx.pow(beta, 2)
           safe_division((1 + beta2) * precision * recall, beta2 * precision + recall)
@@ -709,7 +712,7 @@ defmodule Scholar.Metrics.Classification do
         support = (y_true == Nx.iota({num_classes, 1})) |> Nx.sum(axes: [1])
 
         per_class_fscore =
-          per_class_fscore * support
+          (per_class_fscore * support)
           |> safe_division(Nx.sum(support))
           |> Nx.sum()
 
--- a/lib/scholar/metrics/classification.ex
+++ b/lib/scholar/metrics/classification.ex
@@ -638,6 +638,11 @@ defmodule Scholar.Metrics.Classification do
         f32[2]
         [0.0, 0.0]
       >
+      iex> Scholar.Metrics.Classification.fbeta_score(Nx.tensor([1, 0, 1, 0]), Nx.tensor([0, 1, 0, 1]), 0.5, num_classes: 2, average: :none)
+      #Nx.Tensor<
+        f32[2]
+        [0.0, 0.0]
+      >
   """
   deftransform fbeta_score(y_true, y_pred, beta, opts \\ []) do
     fbeta_score_n(y_true, y_pred, beta, NimbleOptions.validate!(opts, @fbeta_score_schema))
@@ -754,7 +759,7 @@ defmodule Scholar.Metrics.Classification do
       >
   """
   deftransform f1_score(y_true, y_pred, opts \\ []) do
-    fbeta_score_n(y_true, y_pred, Nx.tensor(1), NimbleOptions.validate!(opts, @f1_score_schema))
+    fbeta_score_n(y_true, y_pred, 1, NimbleOptions.validate!(opts, @f1_score_schema))
   end
 
   @doc """
--- a/lib/scholar/metrics/classification.ex
+++ b/lib/scholar/metrics/classification.ex
@@ -658,7 +658,7 @@ defmodule Scholar.Metrics.Classification do
     per_class_fscore
   end
 
-  defn fbeta_score_v(confusion_matrix, average) do
+  defnp fbeta_score_v(confusion_matrix, average) do
     true_positive = Nx.take_diagonal(confusion_matrix)
     false_positive = Nx.sum(confusion_matrix, axes: [0]) - true_positive
     false_negative = Nx.sum(confusion_matrix, axes: [1]) - true_positive
@@ -676,7 +676,7 @@ defmodule Scholar.Metrics.Classification do
     end
   end
 
-  defn precision_recall_fscore_n(y_true, y_pred, beta, num_classes, average) do
+  defnp precision_recall_fscore_n(y_true, y_pred, beta, num_classes, average) do
     confusion_matrix = confusion_matrix(y_true, y_pred, num_classes: num_classes)
     {true_positive, false_positive, false_negative} = fbeta_score_v(confusion_matrix, average)
 
--- a/lib/scholar/metrics/classification.ex
+++ b/lib/scholar/metrics/classification.ex
@@ -41,10 +41,12 @@ defmodule Scholar.Metrics.Classification do
           * `:micro` - Calculate metrics globally by counting the total true positives,
           false negatives and false positives.
 
-          * `:none` - The f1 scores for each class are returned.
+          * `:none` - The F-score values for each class are returned.
           """
         ]
       ]
+
+  fbeta_score_schema = f1_score_schema
 
   confusion_matrix_schema =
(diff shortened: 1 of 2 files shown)


lib/scholar/metrics/classification.ex
```diff
@@ -586,86 +589,174 @@ defmodule Scholar.Metrics.Classification do
...
+        0.6000000238418579
+      >
+      iex> Scholar.Metrics.Classification.f1_score(Nx.tensor([1, 0, 1, 0]), Nx.tensor([0, 1, 0, 1]), num_classes: 2, average: :none)
+      #Nx.Tensor<
+        f32[2]
+        [0.0, 0.0]
+      >
+  """
+  deftransform f1_score(y_true, y_pred, opts \\ []) do
+    fbeta_score_n(y_true, y_pred, Nx.tensor(1), NimbleOptions.validate!(opts, @f1_score_schema))
```
0urobor0s, line 757: Not sure if `Nx.tensor(1)` could be marginally improved by changing to `Nx.tensor(1, type: :u8)` or if, since this is being passed to a `defn` context if the tensor creation is necessary (currently not enforced). From testing, it doesn't seems to make a difference in results.

lib/scholar/metrics/classification.ex
```diff
@@ -586,86 +589,174 @@ defmodule Scholar.Metrics.Classification do
...
-      iex> Scholar.Metrics.Classification.f1_score(Nx.tensor([1,0,1,0]), Nx.tensor([0, 1, 0, 1]), num_classes: 2, average: :none)
+      iex> Scholar.Metrics.Classification.fbeta_score(Nx.tensor([1, 0, 1, 0]), Nx.tensor([0, 1, 0, 1]), Nx.tensor(0.5), num_classes: 2, average: :none)
       #Nx.Tensor<
         f32[2]
         [0.0, 0.0]
       >
   """
-  deftransform f1_score(y_true, y_pred, opts \\ []) do
-    f1_score_n(y_true, y_pred, NimbleOptions.validate!(opts, @f1_score_schema))
+  deftransform fbeta_score(y_true, y_pred, beta, opts \\ []) do
```
polvalente, line 642: You shouldn't rename the function because this is a breaking change. Either keep the name, or add a separate function and delegate both to the same private implementation.

I'd rather keep the name because they're both basically the same function.

josevalim (approved):

lib/scholar/metrics/classification.ex
josevalim: It is not being renamed, please check below the diff :D

lib/scholar/metrics/classification.ex
```diff
@@ -586,86 +589,174 @@ defmodule Scholar.Metrics.Classification do
...
+    fbeta_score_n(y_true, y_pred, beta, NimbleOptions.validate!(opts, @fbeta_score_schema))
   end
 
-  defnp f1_score_n(y_true, y_pred, opts) do
+  defnp fbeta_score_n(y_true, y_pred, beta, opts) do
```
0urobor0s, line 646: Not sure if `beta` needs to receive a `Nx.tensor` or if it can be an integer. From my understanding of the `defn` context the value is converted to a tensor when a Nx function is called.

lib/scholar/metrics/classification.ex
polvalente: You could just pass 1 directly and let it be converted. Though given that the output type is floating point, any integer type tensor will work

lib/scholar/metrics/classification.ex
polvalente: It can be a number, that will be converted to a tensor upon call

lib/scholar/metrics/classification.ex
```diff
@@ -586,86 +589,179 @@ defmodule Scholar.Metrics.Classification do
...
 
-  #{NimbleOptions.docs(@f1_score_schema)}
+  #{NimbleOptions.docs(@fbeta_score_schema)}
 
   ## Examples
 
       iex> y_true = Nx.tensor([0, 1, 1, 1, 1, 0, 2, 1, 0, 1], type: :u32)
       iex> y_pred = Nx.tensor([0, 2, 1, 1, 2, 2, 2, 0, 0, 1], type: :u32)
-      iex> Scholar.Metrics.Classification.f1_score(y_true, y_pred, num_classes: 3)
+      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.tensor(1), num_classes: 3)
```
msluszniak, line 606:
```suggestion
      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.u32(1), num_classes: 3)
```

lib/scholar/metrics/classification.ex
```diff
@@ -586,86 +589,179 @@ defmodule Scholar.Metrics.Classification do
...
       #Nx.Tensor<
         f32[3]
         [0.6666666865348816, 0.6666666865348816, 0.4000000059604645]
       >
-      iex> Scholar.Metrics.Classification.f1_score(y_true, y_pred, num_classes: 3, average: :macro)
+      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.tensor(2), num_classes: 3)
```
msluszniak, line 611:
```suggestion
      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.u32(2), num_classes: 3)
```

lib/scholar/metrics/classification.ex
```diff
@@ -586,86 +589,179 @@ defmodule Scholar.Metrics.Classification do
...
+      #Nx.Tensor<
+        f32[3]
+        [0.6666666865348816, 0.5555555820465088, 0.625]
+      >
+      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.tensor(0.5), num_classes: 3)
```
msluszniak, line 616:
```suggestion
      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.f32(0.5), num_classes: 3)
```

lib/scholar/metrics/classification.ex
```diff
@@ -586,86 +589,179 @@ defmodule Scholar.Metrics.Classification do
...
+      #Nx.Tensor<
+        f32[3]
+        [0.6666666865348816, 0.8333333134651184, 0.29411765933036804]
+      >
+      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.tensor(2), num_classes: 3, average: :macro)
```
msluszniak, line 621:
```suggestion
      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.u32(2), num_classes: 3, average: :macro)
```

lib/scholar/metrics/classification.ex
```diff
@@ -586,86 +589,179 @@ defmodule Scholar.Metrics.Classification do
...
       #Nx.Tensor<
         f32
-        0.5777778029441833
+        0.6157407760620117
       >
-      iex> Scholar.Metrics.Classification.f1_score(y_true, y_pred, num_classes: 3, average: :weighted)
+      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.tensor(2), num_classes: 3, average: :weighted)
```
msluszniak, line 626:
```suggestion
      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.u32(2), num_classes: 3, average: :weighted)
```

lib/scholar/metrics/classification.ex
```diff
@@ -586,86 +589,179 @@ defmodule Scholar.Metrics.Classification do
...
       #Nx.Tensor<
         f32
-        0.6399999856948853
+        0.5958333611488342
       >
-      iex> Scholar.Metrics.Classification.f1_score(y_true, y_pred, num_classes: 3, average: :micro)
+      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.tensor(0.5), num_classes: 3, average: :micro)
```
msluszniak, line 631:
```suggestion
      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.f32(0.5), num_classes: 3, average: :micro)
```

msluszniak (approved):
Looks really good! I only drop some suggestions about using `Nx.tensor` in tests. Imo, if it's possible then replacing it with simplified syntax is better because it's shorter and you explicitly type tensors. But I don't insist on changing every example so if you want to change only some of those then I'm totally ok with that :)

lib/scholar/metrics/classification.ex
0urobor0s: I wasn't aware of these "short-hand functions", I think it makes sense. Just a question, any reason to not use the smaller versions, u8 and f16?

josevalim:
@0urobor0s please run the formatter and we should be good :)

lib/scholar/metrics/classification.ex
polvalente: @0urobor0s getting a bit more into the gist of your question, it wouldn't make much of a difference because the inferred type for the metric would be f32 because it comes from integer typed tensors.

I believe that by using y_true and y_pred as f16, then it would make a difference if beta was f32/f64 or not. Those two would upcast everything to that same type. You can explore a bit on this by looking at `Nx.Type.merge`

0urobor0s: Makes sense, I'll take a look at those. Thank you all.

0urobor0s:
It should be all done. Let me know if I should squash, rebase as well.

josevalim:

PR #186, opened 2023-10-10
Add precision_recall_fscore_support function

0urobor0s:
Noticed that in #122 there was also the issue for `precision_recall_fscore_support` which was mainly done in #185.

### DIFF
--- a/lib/scholar/metrics/classification.ex
+++ b/lib/scholar/metrics/classification.ex
@@ -46,36 +46,7 @@ defmodule Scholar.Metrics.Classification do
         ]
       ]
 
-  fbeta_score_schema =
-    general_schema ++
-      [
-        average: [
-          type: {:in, [:micro, :macro, :weighted, :none]},
-          default: :none,
-          doc: """
-          This determines the type of averaging performed on the data.
-
-          * `:macro` - Calculate metrics for each label, and find their unweighted mean.
-          This does not take label imbalance into account.
-
-          * `:weighted` - Calculate metrics for each label, and find their average weighted by
-          support (the number of true instances for each label).
-
-          * `:micro` - Calculate metrics globally by counting the total true positives,
-          false negatives and false positives.
-
-          * `:none` - The F-score values for each class are returned.
-          """
-        ],
-        beta: [
-          type: {:custom, Scholar.Options, :beta, []},
-          required: true,
-          doc: """
-          Determines the weight of recall in the combined score.
-          For values of `beta` > 1 it gives more weight to recall, while `beta` < 1 favors precision.
-          """
-        ]
-      ]
+  fbeta_score_schema = f1_score_schema
 
   precision_recall_fscore_support_schema =
     general_schema ++
@@ -666,51 +637,49 @@ defmodule Scholar.Metrics.Classification do
 
       iex> y_true = Nx.tensor([0, 1, 1, 1, 1, 0, 2, 1, 0, 1], type: :u32)
       iex> y_pred = Nx.tensor([0, 2, 1, 1, 2, 2, 2, 0, 0, 1], type: :u32)
-      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, beta: Nx.u32(1), num_classes: 3)
+      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.u32(1), num_classes: 3)
       #Nx.Tensor<
         f32[3]
         [0.6666666865348816, 0.6666666865348816, 0.4000000059604645]
       >
-      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, beta: Nx.u32(2), num_classes: 3)
+      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.u32(2), num_classes: 3)
       #Nx.Tensor<
         f32[3]
         [0.6666666865348816, 0.5555555820465088, 0.625]
       >
-      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, beta: Nx.f32(0.5), num_classes: 3)
+      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.f32(0.5), num_classes: 3)
       #Nx.Tensor<
         f32[3]
         [0.6666666865348816, 0.8333333134651184, 0.29411765933036804]
       >
-      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, beta: Nx.u32(2), num_classes: 3, average: :macro)
+      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.u32(2), num_classes: 3, average: :macro)
       #Nx.Tensor<
         f32
         0.6157407760620117
       >
-      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, beta: Nx.u32(2), num_classes: 3, average: :weighted)
+      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.u32(2), num_classes: 3, average: :weighted)
       #Nx.Tensor<
         f32
         0.5958333611488342
       >
-      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, beta: Nx.f32(0.5), num_classes: 3, average: :micro)
+      iex> Scholar.Metrics.Classification.fbeta_score(y_true, y_pred, Nx.f32(0.5), num_classes: 3, average: :micro)
       #Nx.Tensor<
         f32
         0.6000000238418579
       >
-      iex> Scholar.Metrics.Classification.fbeta_score(Nx.tensor([1, 0, 1, 0]), Nx.tensor([0, 1, 0, 1]), beta: Nx.tensor(0.5), num_classes: 2, average: :none)
+      iex> Scholar.Metrics.Classification.fbeta_score(Nx.tensor([1, 0, 1, 0]), Nx.tensor([0, 1, 0, 1]), Nx.tensor(0.5), num_classes: 2, average: :none)
       #Nx.Tensor<
         f32[2]
         [0.0, 0.0]
       >
-      iex> Scholar.Metrics.Classification.fbeta_score(Nx.tensor([1, 0, 1, 0]), Nx.tensor([0, 1, 0, 1]), beta: 0.5, num_classes: 2, average: :none)
+      iex> Scholar.Metrics.Classification.fbeta_score(Nx.tensor([1, 0, 1, 0]), Nx.tensor([0, 1, 0, 1]), 0.5, num_classes: 2, average: :none)
       #Nx.Tensor<
         f32[2]
         [0.0, 0.0]
       >
   """
-  deftransform fbeta_score(y_true, y_pred, opts \\ []) do
-    opts = NimbleOptions.validate!(opts, @fbeta_score_schema)
-    {beta, opts} = Keyword.pop(opts, :beta)
-    fbeta_score_n(y_true, y_pred, beta, opts)
+  deftransform fbeta_score(y_true, y_pred, beta, opts \\ []) do
+    fbeta_score_n(y_true, y_pred, beta, NimbleOptions.validate!(opts, @fbeta_score_schema))
   end
 
   defnp fbeta_score_n(y_true, y_pred, beta, opts) do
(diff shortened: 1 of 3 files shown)


lib/scholar/metrics/classification.ex
```diff
@@ -672,11 +674,70 @@ defmodule Scholar.Metrics.Classification do
...
+        f32
+        0.6000000238418579
+      >
+      iex> Scholar.Metrics.Classification.precision_recall_fscore_support(Nx.tensor([1, 0, 1, 0]), Nx.tensor([0, 1, 0, 1]), 2, num_classes: 2, average: :none)
+      #Nx.Tensor<
+        f32[2]
+        [0.0, 0.0]
+      >
+  """
+  deftransform precision_recall_fscore_support(y_true, y_pred, beta \\ 1, opts \\ [])
```
0urobor0s, line 720: Is there a better way to have this possibility of having the `beta` value as default 1?

lib/scholar/metrics/classification.ex
```diff
@@ -672,11 +674,70 @@ defmodule Scholar.Metrics.Classification do
...
+
+  deftransform precision_recall_fscore_support(y_true, y_pred, beta, opts) when not is_list(beta) and is_list(opts) do
+    opts = NimbleOptions.validate!(opts, @precision_recall_fscore_support_schema)
+
+    precision_recall_fscore_support_n(
+      y_true,
+      y_pred,
+      beta,
+      opts[:num_classes],
+      opts[:average]
```
0urobor0s, line 734: Passing the atom `:none` directly instead of the `opts` list fails with:
```
** (Protocol.UndefinedError) protocol Nx.LazyContainer not implemented for :none of type Atom. This protocol is implemented for the following type(s): Any, Atom, Complex, Float, Integer, List, Map, Nx.Batch, Nx.Tensor, Tuple
```

Should the list be passed instead of single values?

lib/scholar/metrics/classification.ex
msluszniak: Yes, there is a better way. If you want to have a default on a value that needs to be passed, then use NimbleOptions schema with `default: 1`.

msluszniak: Like [here](https://github.com/elixir-nx/scholar/blob/93d3da6bce3d47b44f5c30b7ac9c20aaa7984f00/lib/scholar/metrics/classification.ex#L81)

lib/scholar/metrics/classification.ex
msluszniak: if you want to pass any atoms to the `defn` you should do this via `opts`. So yes, use keyword list instead :)

lib/scholar/metrics/classification.ex
0urobor0s: Should the arguments default to use the NimbleOptions schema for these kinds of default arguments? I was thinking of having the API follow the same way as in `fbeta_score`. If we move to use NimbleOptions, I think that function should also change (the `beta` argument), so it's "consistent" with each other. What do you think?

msluszniak: Yes, I think it's a good idea to change `fbeta_score` to accept `:beta` parameter via `opts`. In the whole `Scholar` it's a standard to pass even very important params through opts like `k` in `K-means`

lib/scholar/metrics/classification.ex
```diff
@@ -677,7 +737,119 @@ defmodule Scholar.Metrics.Classification do
...
+  ## Options
+
+  #{NimbleOptions.docs(@precision_recall_fscore_support_schema)}
+
+  ## Examples
+
+      iex> y_true = Nx.tensor([0, 1, 1, 1, 1, 0, 2, 1, 0, 1], type: :u32)
+      iex> y_pred = Nx.tensor([0, 2, 1, 1, 2, 2, 2, 0, 0, 1], type: :u32)
+      iex> Scholar.Metrics.Classification.precision_recall_fscore_support(y_true, y_pred, num_classes: 3)
+      {#Nx.Tensor<
```
0urobor0s, line 754: Doctests seem to be failing.

lib/scholar/metrics/classification.ex
```diff
@@ -700,13 +872,15 @@ defmodule Scholar.Metrics.Classification do
...
     case average do
       :none ->
-        {precision, recall, per_class_fscore}
+        support = (y_true == Nx.iota({num_classes, 1})) |> Nx.sum(axes: [1])
+
+        {precision, recall, per_class_fscore, support}
 
       :micro ->
-        {precision, recall, per_class_fscore}
+        {precision, recall, per_class_fscore, Nx.Constants.nan()}
```
0urobor0s, line 880: Any idea on what to add here? The first try was to use `:none`, but it seems atoms can't be returned from `defn`.

msluszniak: Using NaNs seems reasonable

lib/scholar/metrics/classification.ex
msluszniak: When passing a tuple with tensors as a result you need to pass it as `Nx.tensor([...])` not with syntax printed in the terminal

lib/scholar/metrics/classification.ex
```diff
@@ -603,58 +665,56 @@ defmodule Scholar.Metrics.Classification do
...
+      iex> Scholar.Metrics.Classification.fbeta_score(Nx.tensor([1, 0, 1, 0]), Nx.tensor([0, 1, 0, 1]), beta: 0.5, num_classes: 2, average: :none)
       #Nx.Tensor<
         f32[2]
         [0.0, 0.0]
       >
   """
-  deftransform fbeta_score(y_true, y_pred, beta, opts \\ []) do
-    fbeta_score_n(y_true, y_pred, beta, NimbleOptions.validate!(opts, @fbeta_score_schema))
+  deftransform fbeta_score(y_true, y_pred, opts \\ []) do
+    fbeta_score_n(y_true, y_pred, NimbleOptions.validate!(opts, @fbeta_score_schema))
```
josevalim, line 710: We must never pass tensors as options to defn. So if you want to make beta an option, you need to extract the value and validate it here:

```suggestion
    opts = NimbleOptions.validate!(opts, @fbeta_score_schema)
    {beta, opts} = Keyword.pop(opts, :beta)
    fbeta_score_n(y_true, y_pred, beta, opts)
```

lib/scholar/metrics/classification.ex
0urobor0s: Not sure I follow, I've tried both this way:
```
iex> Scholar.Metrics.Classification.precision_recall_fscore_support(y_true, y_pred, num_classes: 3)
{Nx.Tensor([0.6666666865348816, 1.0, 0.25]),
Nx.Tensor([0.6666666865348816, 0.5, 1.0]),
Nx.Tensor([0.6666666865348816, 0.6666666865348816, 0.4000000059604645]),
Nx.Tensor([3, 6, 1])}
```

and also:
```
iex> Scholar.Metrics.Classification.precision_recall_fscore_support(y_true, y_pred, num_classes: 3)
{#Nx.Tensor<f32[3][0.6666666865348816, 1.0, 0.25]>, #Nx.Tensor<f32[3][0.6666666865348816, 0.5, 1.0]>, #Nx.Tensor<f32[3][0.6666666865348816, 0.6666666865348816, 0.4000000059604645]>, #Nx.Tensor<u64[3][3, 6, 1]>}
```

But both don't work, with the following given the following:
```
If you are planning to assert on the result of an iex> expression which contains a value inspected as #Name<...>, please make sure the inspected value is placed at the beginning of the expression; otherwise Elixir will treat it as a comment due to the leading sign #.
```

lib/scholar/metrics/classification.ex
0urobor0s: I would personally prefer to have something more clear that this value contains "nothing" useful. But with the current constraints I thought this was ok(ish). Maybe a new type/struct could be created for these cases?

lib/scholar/metrics/classification.ex
0urobor0s: Did not know that. I've made the necessary changes.

Is this because of defn elixir subset, or something else like the Nx.LazyContainer protocol and it's implications when compiling defn code?

josevalim: It is how `Nx.Defn` caches compilation. When passed as arguments, we cache based on their shape and type. However, option are cached based on their values, so a tensor with the same shape but with different values would require new compilations. :)

lib/scholar/metrics/classification.ex
```diff
@@ -677,7 +739,122 @@ defmodule Scholar.Metrics.Classification do
...
+      >,
+      #Nx.Tensor<
+       f32
+       0.6000000238418579
+      >,
+      #Nx.Tensor<
+       f32
+       NaN
+      >}
+      iex> Scholar.Metrics.Classification.precision_recall_fscore_support(Nx.tensor([1, 0, 1, 0]), Nx.tensor([0, 1, 0, 1]), beta: 2, num_classes: 2, average: :none)
```
josevalim, line 823: Please add new lines before `iex` whenever you can, so ExDoc formats them as distinct examples. Also, if you can break this line into several ones (for exmaple, by assigning the tensors and opts to variables, it can help with readability).

lib/scholar/metrics/classification.ex
```diff
@@ -603,58 +665,58 @@ defmodule Scholar.Metrics.Classification do
...
+    opts = NimbleOptions.validate!(opts, @fbeta_score_schema)
+    {beta, opts} = Keyword.pop(opts, :beta)
+    fbeta_score_n(y_true, y_pred, beta, opts)
   end
 
   defnp fbeta_score_n(y_true, y_pred, beta, opts) do
-    check_shape(y_pred, y_true)
-    num_classes = check_num_classes(opts[:num_classes])
-    average = opts[:average]
+    check_beta(beta)
```
josevalim, line 716: I think we can remove this. It has a default value and it cannot be nil as an option.

0urobor0s: I'm not defining a default value of beta in `fbeta_score_schema`, only in `precision_recall_fscore_support_schema`, which was the reason to leave the check there.

josevalim: If it is required there, why not make it an argument then?

0urobor0s: That was the idea, but we've moved to this instead in order to be consistent with the other functions that use the beta parameter.
https://github.com/elixir-nx/scholar/pull/186#discussion_r1352874935

0urobor0s: Maybe using the [NimbleOptions schema options](https://hexdocs.pm/nimble_options/NimbleOptions.html#module-schema-options) with the required as true, would be better in this case?

lib/scholar/metrics/classification.ex
0urobor0s: Apologies, I didn't understand it. I've made a commit with the changes.

lib/scholar/metrics/classification.ex
josevalim: Honestly, I would prefer to keep it as an argument, but it is not a strong opinion. Or we remove the function and make it an option on f1_score.

0urobor0s: In another kind of library I would agree, but for these kinds of data science/statistical learning libraries, I'm partial to having a consistent/predictable API.

josevalim (commented):
Hi @0urobor0s! For `fbeta_score`, I am still not convinced we should have an option as required value. I think we should either give it a default or expect it as argument.

I can think of two options:

1. Rename the function to `fscore`, to mirror `fscore` in "precision_recall" and make `beta` optional

2. Keep beta a required argument in fbeta_score (as the name says)

0urobor0s:
Hi @josevalim,
Not sure about the ergonomics of the API (and also this [case](https://github.com/elixir-nx/scholar/pull/186#discussion_r1355576748)), but I understand the rationale behind your thinking. I've made the necessary changes.

In the topic of renaming, I think `precision_recall_fscore_support`, would be better called `precision_recall_fbeta_score_support`. However there might be value in having the same name as in `scikit-learn`.

lib/scholar/options.ex
```diff
@@ -100,4 +100,12 @@ defmodule Scholar.Options do
     {:error,
      "expected metric to be a :cosine or tuple {:minkowski, p} where p is a positive number or :infinity, got: #{inspect(metric)}"}
   end
+
+  def beta(beta) do
+    if (is_number(beta) and beta >= 0) or (Nx.is_tensor(beta) and Nx.size(beta) == 1) do
```
msluszniak, line 105: You want here `beta` to be scalar because later on it is used in condition. A tensor that has `Nx.size` equal to 1 is for example ```Nx.tensor([[[[[[1]]]]]])``` ;)
```suggestion
    if (is_number(beta) and beta >= 0) or (Nx.is_tensor(beta) and Nx.rank(beta) == 0) do
```

msluszniak (approved):

josevalim:

PR #187, opened 2023-10-10
Hierarchical Clustering

billylanchantin:
## Description

This PR adds hierarchical clustering. It's one of the items with a `help wanted` label, so I thought I'd give it a go :)

## Background

Hierarchical clustering is a family of unsupervised learning methods that build clusters in a greedy fashion, either bottom-up (agglomerative) or top-down (divisive). If you're unfamiliar, I think the following tutorial is probably approachable:

https://www.kdnuggets.com/2019/09/hierarchical-clustering.html

This PR adds support for agglomerative clustering, which seems to be the most common:

* https://scikit-learn.org/stable/modules/generated/sklearn.cluster.AgglomerativeClustering.html#sklearn.cluster.AgglomerativeClustering
* https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.linkage.html#scipy.cluster.hierarchy.linkage

## Approach

The plan is to follow the approach outlined in the paper [Modern hierarchical, agglomerative clustering algorithms](https://arxiv.org/abs/1109.2378) by Daniel Müllner. It's still state of the art AFAIK, and it serves the basis for Scipy's and Scikit-learn's hierarchical clustering (though both of those have extra bells and whistles).

The core of hierarchical clustering, which the Müllner paper denotes the "primitive" algorithm, is this:

1. Start by assuming each data point is in a cluster by itself.
2. Build a (square) pairwise dissimilarity matrix from the data points.
3. Choose from among 7 dissimilarity update functions.
4. Use the update function to
   a. group the closest pair of clusters into a new cluster and
   b. recalculate the distances from the new cluster to the rest of the clusters
5. Repeat step 4 until there is a single cluster containing all data points.

The primitive algorithm is best case O(n^3). The paper's main contribution is to describe 3 optimizations to improve the complexity. Which optimization to use depends on which of the 7 update functions was chosen.

So far, I've only implemented the primitive algorithm, which isn't to be used in practice. **If what I've described so far is the direction we want to take, I can start working on the 3 optimizations.**

### To-do

- [x] Dissimilarity functions
  - [x] Euclidean
  - ~Pre-computed~
    - Future work
- [x] Dissimilarity update functions
  - [x] Average
  - ~Centroid~
    - Future work
  - [x] Complete
  - ~Median~
    - Future work
  - [x] Single
  - [x] Ward
  - [x] Weighted
- [x] Cluster functions
  - [x] "parallel-nearest-neighbor"
  - ~"primitive" (not to be used in practice)~
    - (Available in a commit)
  - ~"generic"~
    - Future work
  - ~"minimum-spanning-tree"~
  - ~"nearest-neighbor-chain"~
    - (Available in a commit)

## Livebook demo

I'm including a livebook demo since the dendrogram visualization is very useful for understanding hierarchical clusters.

Screen Shot 2023-11-20 at 5 51 51 PM


msluszniak:
Hi Billy, thank you for the PR. As far as I scanned through your PR you mostly use `def` functions. It is very important that all the functionalities are implemented with `defn`. It gives orders of magnitude improvements to performance. The other reason is that only other `defn`s and `deftransform`s can be used inside other defn.

billylanchantin:
> As far as I scanned through your PR you mostly use `def` functions. It is very important that all the functionalities are implemented with `defn`.

@msluszniak Will do! I mostly wanted a gut check on the direction before I started to "tensor-ify" the operations.

On that note, I may need some help as I get into the optimizations with making them `Nx`-friendly. But I'll call out any parts I get stuck on in the PR.

josevalim:
@billylanchantin issue #189 contains a good discussion on the challenges to optimize an algorithm to run on Nx/defn. However, for HCA, it seems there exists an algorithm already for the GPU here: https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=94392889916911b13e8a40e9cffc1d5924e93fc2

I am not familiar with HCA but I believe the computation of the distance matrix is straight-forward and already done. So you need to implement the linkage algorithm in section 2.4. Step 1 seems to be an argsort (you may need to initialize the distance matrix with infinity/neginfinity to get the proper indexes). The other steps (2-5) would be done in a while loop, so not that parallelizable, but at least should work in Nx. You could even do them in Elixir and we can help convert to Nx later. The important is to rely only on tensors (no maps, etc).

billylanchantin:
> issue https://github.com/elixir-nx/scholar/issues/189 contains a good discussion on the challenges to optimize an algorithm to run on Nx/defn.

Oh that looks like a great resource, I'll give it a read!

> However, for HCA, it seems there exists an algorithm already for the GPU here: https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=94392889916911b13e8a40e9cffc1d5924e93fc2

I wasn't aware of this one. I'll take a look at it as well.

It does appear to predate the work I was planning to implement. But it may be worth having an implementation that's less featureful, but is more GPU-friendly. I'll have to see how they compare.

> I am not familiar with HCA but I believe the computation of the distance matrix is straight-forward and already done.

Yeah unfortunately the computation of the distance function is the "easy" part. It's the various while loops that are the challenge.

At each iteration of cluster analysis, we merge two clusters. This decreases number of clusters by one (which changes shape of the distance matrix) and alters all distances involving the two merged clusters. It's a lot of bookkeeping.

Plus in the Müeller paper, he's using not-necessarily-tensor-friendly things like priority queues to speed up the lookups.

josevalim:
@billylanchantin I may be very wrong, but inspired on the paper, I wrote a version of linking for minimum HCA in a Livebook. The linking is parallel but `log2(n)` so it seems better than the paper: https://gist.github.com/josevalim/00990eb921bc1f767e8870798b5e786f

You can see it is recursive and it requires `ceiling(:math.log2(size))` operations (or `32 - Nx.count_leading_zeros(Nx.u32(size))` if you prefer integer arithmetic). I can convert it to a proper `defn` tomorrow if you don't beat me to it. The most complex part will be the logic to update the pairwise matrix but maybe the existing papers have better ideas.

josevalim:
The approach above assumes binary trees. So only two elements at a time. My algorithm requires stable sorting. So we need to add a TODO because the default will change on Nx 0.7.

billylanchantin:
@josevalim See my comment on your gist. I may have been mistaken, though. I'll take a closer look tomorrow!

josevalim:
Oh, I thought the simple did not update the matrix, but you are right. I will work on it. In that case, it is worth noting that my algorithm will only work if points do not get closer to clusters as clusters are merged. According [to this paper](https://www.sciencedirect.com/science/article/pii/016781919500017I), this is called the reducibility property (section 2.2) and luckily it applies to single, average, complete, and minimum variance.

---

For the other metrics, we will need to implement [the paper I linked earlier](https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=94392889916911b13e8a40e9cffc1d5924e93fc2). The important thing to highlight is that `argsort` is your priority queue and if you access their results, you get their values. The algorithm is roughly like this:

1. Make the pairwise matrix an upper triangular matrix where the diagonal and the lower part are made of infinity
2. Compute the argsort of the matrix, that will give the closest point to any point `i`
3. Take the values from the argsorted matrix, these are the `minimal_distances`
4.  `argsort` the minimal distance, the first element is the closest cluster that you have
5. Set the value of this cluster to infinity in the `minimal_distance`, update the pairwise matrix, and then `minimal_distances`, go back to step 2 (n-1 times)

There is an optimization in that we don't need to argsort again if the next entry in minimal distances is smaller than the updates to the pairwise matrix, this means we can skip steps 2, 3, and 4. This is probably the priority queue optimization done by other papers.

This algorithm should work on all metrics, it is just slower, so the good news is that once we implement one version, we can test and benchmark against it.

josevalim:
I am delayed a bit because we are adding new functions to Nx to make updating symmetrical matrices efficient.

billylanchantin:
No rush! I've been finding it hard to get the long blocks of time I need for this PR anyhow.

josevalim:
> Oh, I thought the simple did not update the matrix, but you are right.

I have implemented the full algorithm for simple and I can confirm that we don't need to update the matrix. The simple version always updates the distances to the minimum distance of all points, which we already get argsort. So if we put my previous gist in a loop, it will be enough to compute the single-link minimal version. :)

I will provide a gist with the full version of other algorithms later. If you want, I can also provide the loop version for minimal.

billylanchantin:
> I will provide a gist with the full version of other algorithms later. If you want, I can also provide the loop version for minimal.

That'd be fantastic! Thanks for looking into this :)

(And sorry about the delay. It was actually the impetus for the Nx exercises PR: I kept getting stuck so I decided I needed some practice.)

And I admit I'm curious how this works. Without updated distances, I don't think the dendrogram object will have the right tree heights. But your gist will answer that!

josevalim:
@billylanchantin you need the distances between clusters as well? I can work on that. :)

billylanchantin:
@josevalim Yeah there are two goals with HC:

1. The clusters themselves
2. The dendrogram visualization

For the clusters themselves, you don't necessarily need to maintain the distances. But for the dendrogram, you need them to build the heights of the tree.

One of the advantages of HC is that you can use the dendrogram to help you choose post-hoc how many clusters you want. (I can explain this further if requested 😄)

josevalim:
@billylanchantin here is the complete code for single links:

https://gist.github.com/josevalim/00990eb921bc1f767e8870798b5e786f

I can also make it work for average, complete, and minimum variance. Your goal would be to traverse the links and distances and build the clusters. I can optimize the code a bit which would require a tiny modification to the output, but it should be straight-forward to adapt a bit. So with the above, we should have everything in hand for a MVP of hierarchical clustering.

For centroids and median, we will need [the second half of this comment](https://github.com/elixir-nx/scholar/pull/187#issuecomment-1794254332), which will also have a different output for links and distances (most likely two `{n}` sized vectors). So we can consider it something completely separate.

billylanchantin:
@josevalim I just pushed a new approach.

I was reviewing your approach and I got worried about the `argsort` in the main loop. AFAICT, that makes its runtime $O(n^2 \cdot \log(n))$.

The approach I've pushed from the Müllner paper is I believe $O(n^2)$. It's a bit rough around the edges ATM, but I think the core idea is solid:

You recursively call `b = Nx.argmin(pairwise, a)`. If you ever find an `a-b-a` chain, you've found a new pair. This doesn't require sorting. Also, the expected number of calls drops every iteration because you don't ever look at old rows.

LMK what you think. If it's too much to review quickly I can explain further.

josevalim:
The problem with argmin inside the loop is that it is not paralellizable. However, I completely forgot argmin existed and I think my previous algorithm will work* by simply replacing:

```
closest = Nx.argsort(pairwise, axis: 1, stable: true, type: :u32)[[.., 0]]
```

by:

```
closest = Nx.argmin(pairwise, axis: 1)
```

However, in order for this to work, we need to use `:infinity` instead of `NaN`. I think most of the algorithm will work like this but be careful because your code to update distances may introduce `NaN`s (I would have a flag that you run for tests that check no NaNs have been added to the pairwise matrix at the end).

billylanchantin:
Got it!

Yeah, your version (after replacing w/ `argmin`) is sorta doing the same thing as nearest-neighbor-chain but in parallel. It does more work overall, but much of it happens at the same time.

One thing to note: as nearest-neighbor-chain progresses, it does at most `n - current_iteration` runs. So its expected and worst case both get better as time runs on. Yours always evaluates every row.

I wonder if we can eliminate that drawback by only ever searching through the relevant rows? I'll think about it, but the parallel version is probably preferable regardless.

josevalim:
I have updated the gist to use argmin and infinity. It uses Nx v0.6 branch so we can pass type: :u32 to argmin: https://gist.github.com/josevalim/00990eb921bc1f767e8870798b5e786f

josevalim:
In the optimal case my version will perform `ceil(log2(n))` sequential steps with `n` parallel steps inside. Your version performs `n` sequential steps with `n - current_iteration` sequential steps inside.

One alternative is to mix both approaches. We go the first 80% (or whatever configurable rate) of the steps using my approach, then the last ones using yours. I believe both versions can use the same `update_pairwise` function I defined, except in your case the links tensor will always have shape `{1}`. So we just end-up repeating the outer loop, which is minimal.

PS: I pushed a fix to the gist as we no longer need to use topk.

billylanchantin:
> In the optimal case my version will perform ceil(log2(n)) sequential steps with n parallel steps inside.

n sequential steps, right? I think we dropped the log part.

> One alternative is to mix both approaches.

Oh that's an interesting idea. In fact, you can probably work out the optimal hand-off point based on the number of processors.

(Though it's probably a bit early to get that optimal. I think we shoul just get your version working. Now that my code is in git, we can come back to this PR later and work on a version that hands off.)

Also, just to double check, is there a semi-performant way to filter the `:infinity`s out of `pairwise`? (If it's expensive, we could do it every time the iteration is a power of 2 or something.) If we were able to filter down and keep track of the indices, I think your version would always be optimal.

josevalim:
> Also, just to double check, is there a semi-performant way to filter the :infinitys out of pairwise?

The problem is that even after step n, we may have points that have never joined a cluster, so its chain is still large.

billylanchantin:
Oh I misunderstood your fix. I just re-read and now I see that `counter = counter + 1` is happening in a loop. That's how we keep the runtime down to `log(n)`.

And it's why we need to keep relying on the reducibility property. If we start with 4 clusters `{{a}, {b}, {c}, {d}}` and we return `{{a, b}, {c, d}}` from a parallel pass, we need a guarantee that something like `{{a, b}, {c}, {d}} -> {{a, b, c}, {d}}` isn't better.

> So the only way we could trim it down would be with a sparse matrix, which are not currently supported.

Gotcha. I'll lay out my thought so I don't forget. Maybe we can use it later.

```elixir
# Example data
data = [[10], [11], [20], [21], [30], [31]]
pairwise = [
  [ Inf,  1.0, 10.0, 11.0, 20.0, 21.0],
  [ 1.0,  Inf,  9.0, 10.0, 19.0, 20.0],
  [10.0,  9.0,  Inf,  1.0, 10.0, 11.0],
  [11.0, 10.0,  1.0,  Inf,  9.0, 10.0],
  [20.0, 19.0, 10.0,  9.0,  Inf,  1.0],
  [21.0, 20.0, 11.0, 10.0,  1.0,  Inf]
]

# Run parallel pass, get these clusters
new_clusters = [[0, 1], [2, 3], [4, 5]]

# Do the normal update
pairwise = [
  [ Inf,  1.0,  9.0,  9.0, 19.0, 19.0],
  [ 1.0,  Inf,  9.0,  9.0, 19.0, 19.0],
  [ 9.0,  9.0,  Inf,  1.0,  9.0,  9.0],
  [ 9.0,  9.0,  1.0,  Inf,  9.0,  9.0],
  [19.0, 19.0,  9.0,  9.0,  Inf,  1.0],
  [19.0, 19.0,  9.0,  9.0,  1.0,  Inf]
]

# Then remove row/col of the higher index in each cluster.
higher = [1, 3, 5]
pairwise = [
  [ Inf,  9.0, 19.0],
  [ 9.0,  Inf,  9.0],
  [19.0,  9.0,  Inf],
]

# Now the next time we find clusters, the indices will be wrong.
# So we need to do some bookkeeping.
actual = [0, 1, 2, 3, 4, 5]
actual = while actual, h <- higher, do: Nx.select(actual < h, actual, actual + 1)
# => [0, 2, 4, 6, 7, 8]
#              ^ past this point won't ever happen
```

polvalente:
The problem with slicing in that way is that you're dynamically changing shapes, which doesn't work inside defn. So you end up having to operate on N elements in all iterations

josevalim:
> And it's why we need to keep relying on the reducibility property. If we start with 4 clusters {{a}, {b}, {c}, {d}} and we return {{a, b}, {c, d}} from a parallel pass, we need a guarantee that something like {{a, b}, {c}, {d}} -> {{a, b, c}, {d}} isn't better.

exactly!

billylanchantin:
Bunch of updates!

* @josevalim I've added your approach. There were a few tweaks I needed to make but they were minor. I think we're good to go now.
* I changed the terminology from "cluster" to ["clade"](https://en.wikipedia.org/wiki/Cladistics)  so we could call the final clusters "clusters" without confusion.
* I wrapped the result in a struct. Not married to the specifics.

---

I did some rough benchmarking:

<details>
<summary>Code</summary>

```elixir
key = Nx.Random.key(0)

for i <- 1..20 do
  num = 50 * i
  {data, _} = Nx.Random.normal(key, 1000, 10000, shape: {num, 1})
  data = Nx.add(data, Nx.add(1, Nx.reduce_min(data)))

  {us, _} = :timer.tc(fn -> Hierarchical.fit(data) end)

  {num, us / 1_000_000}
  |> IO.inspect()
end
```

</details>

<details>
<summary>Results</summary>

   n | seconds
:--  | :--
  50 |   0.1
 100 |   0.5
 150 |   1.9
 200 |   3.9
 250 |   7.5
 300 |  12.2
 350 |  19.7
 400 |  27.8
 450 |  41.3
 500 |  56.2
 550 |  75.2
 600 | 101.5
 650 | 130.7
 700 | 171.0
 750 | 231.1
 800 | 291.5
 850 | 375.4
 900 | 512.1
 950 | 661.1
1000 | 859.0

</details>

It looks like we're getting ~quadratic behavior. Probably the parallel step doesn't yield many new clades per generation on average.

---

I had some trouble getting the mix.exs to work off a `git` ref (AFAICT it started with [this](https://github.com/elixir-nx/nx/commit/b784bf24e86c1e1b8120b68565b5fb0ac60686d8) commit). This may be a blocker to tests. I worked off a local copy of Nx while developing.

josevalim:
> I had some trouble getting the mix.exs to work off a git ref

remember to pass the :sparse option!

josevalim:
Btw, are those benchmarks over the cpu or gpu? Cpu won’t be good at leveraging parallelism.

I am currently on holidays, so I won’t be able to take a look at the code within the next days, only on my phone for now. :)

josevalim:
Unfortunately Nx for the CPU leaves a lot to be desired. At best it will use SIMD but not multi core.

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,450 @@
...
+  (Cluster labels are arbitrary, but deterministic.)
+
+  ## Examples
+
+      iex> data = Nx.tensor([[2], [7], [9], [0], [3]])
+      iex> model = Hierarchical.fit(data)
+      iex> Hierarchical.cluster(model, by: [num_clusters: 3])
+      Nx.tensor([0, 1, 1, 2, 0])
+  """
+  deftransform cluster(%__MODULE__{} = model, opts \\ []) do
```
billylanchantin, line 382: I went with `cluster/2` instead of `predict/2`. It serves a similar role as `predict/2` would, but it can only be used on the original data. There's no way to take the output of `fit/2` and apply it to new data.

(In that vein, `fit/2` may also be a misnomer.)

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,450 @@
...
+
+  ## Limitations
+
+  Due to the requirements of the current implementation, only these options are supported:
+
+    * `dissimilarity: :euclidean`
+    * `linkage: :average | :complete | :single | :ward | :weighted`
+
+  Our current algorithm is $O(\\frac{n^2}{p})$ where $n$ is the number of data points
+  and $p$ is the number of processors.
```
billylanchantin, line 28: Looking back over what I've got, I'm not totally sure what the complexity is. I think each clade-finding pass is $O(\frac{n^2}{p})$, but I'm still getting used to Nx.

mix.exs
```diff
@@ -54,17 +54,15 @@ defmodule Scholar.MixProject do
...
         "notebooks/k_nearest_neighbors.livemd",
         "notebooks/cv_gradient_boosting_tree.livemd"
       ],
       groups_for_modules: [
         Models: [
-          Scholar.Cluster.AffinityPropagation,
-          Scholar.Cluster.DBSCAN,
-          Scholar.Cluster.GaussianMixture,
-          Scholar.Cluster.KMeans,
+          ~r/Scholar.Cluster/,
```
billylanchantin, line 65: This is a bit of code golf. Doing it this way has side effects, so I'm not sure if it's desired.

notebooks/hierarchical_clustering.livemd
```diff
@@ -0,0 +1,217 @@
...
+
+## Introduction
+
+```elixir
+defmodule Kino.CanvasDendrogram do
+  use Kino.JS
+
+  def new(graph), do: Kino.JS.new(__MODULE__, graph)
+
+  asset "main.js" do
```
billylanchantin, line 25: I've kept this out of the sidebar for the moment. Despite the LOC, there's not much here except the plotting logic. See also the commented out section of the moduledoc in `Scholar.Cluster.Hierarchical`.

lib/scholar/cluster/hierarchical.ex
josevalim: @billylanchantin how is it called in scikit-learn? I would generally prefer to avoid introducing new names. If this is a single operations, perhaps we should use `fit_predict` instead? /cc @msluszniak

mix.exs
josevalim: I like this approach but we need to escape dots. Perhaps it is better done in a separate PR too:

```suggestion
          ~r/Scholar\.Cluster\./,
```

notebooks/hierarchical_clustering.livemd
josevalim: Yeah, I have been long thinking it would be nice if Scholar already included visualizations for several of its modules, so I am :+1: for getting started with this trend. Just make sure to render it as a tabbed output, so we also have access to the "Raw" data structure.

lib/scholar/cluster/hierarchical.ex
msluszniak: Yes, `fit_predict` sounds good if it already returns the clustering

mix.exs
billylanchantin: > Perhaps it is better done in a separate PR too

👍 Yeah I think that's better. Will change.

lib/scholar/cluster/hierarchical.ex
billylanchantin: Ah yeah they have `fit` and `fit_predict`. I'm good with that! @msluszniak You'll definitely want to take a pass and look for API consistency.

> Yes, `fit_predict` sounds good if it already returns the clustering

I actually had that first but switched. I'll switch back.

notebooks/hierarchical_clustering.livemd
billylanchantin: It may be worth keeping it as is (not visible in the sidebar) for now and making available in another PR. I looked at a few other guide notebooks in the repo and they're more detailed. This one doesn't even show the clusters, just the raw dendrogram. Typically a dendrogram plot will assign a color to each cluster.

josevalim: Yup, we can postpone adding it for sure. But to be clear, I was thinking about moving this Kino module itself into Scholar and declaring Kino an optional dependency. The goal is that many of our modules could come with built-in visualization, but that's a separate convo (hence let's postpone it). :)

jonatanklosko: Btw. it should be `KinoScholar.CanvasDendrogram` or `Scholar.Kino.CanvasDendrogram`, the `Kino.` prefix is reserved for "official" kino packages, since for those we make sure there are no conflicts (and there are historic reasons :).

lib/scholar/cluster/hierarchical.ex
billylanchantin: [52b1e78](https://github.com/elixir-nx/scholar/pull/187/commits/52b1e78cf38cfd62d4e352c3ffb22cb9e528f556)

notebooks/hierarchical_clustering.livemd
billylanchantin: [5589061](https://github.com/elixir-nx/scholar/pull/187/commits/5589061229fe244bc9ff3049e17d43e4cdad4a7c)

mix.exs
billylanchantin: [2f2de8e](https://github.com/elixir-nx/scholar/pull/187/commits/2f2de8e36ae560fe37bba10e41609b21dfc79954)

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,476 @@
...
+  defnp parallel_nearest_neighbor(pairwise, update_fun) do
+    {n, _} = Nx.shape(pairwise)
+    pairwise = Nx.broadcast(:infinity, {n}) |> Nx.make_diagonal() |> Nx.add(pairwise)
+    clades = Nx.broadcast(-1, {n - 1, 2})
+    sizes = Nx.broadcast(1, {2 * n - 1})
+    pointers = Nx.broadcast(-1, {2 * n - 2})
+    diss = Nx.tensor(:infinity, type: Nx.type(pairwise)) |> Nx.broadcast({n - 1})
+
+    {clades, _count, _pointers, _pairwise, diss, sizes} =
+      while {clades, count = 0, pointers, pairwise, diss, sizes}, count < n - 1 do
```
msluszniak, line 214:
```suggestion
    {clades, _} =
      while {clades, {count = 0, pointers, pairwise, diss, sizes}}, count < n - 1 do
```

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,476 @@
...
+        clades_selector = Nx.equal(nearest_of_nearest, Nx.iota({n}))
+
+        # Now let's get the links that form clades.
+        # They are bidirectional but let's keep only one side.
+        links = Nx.select(clades_selector and nearest > nearest_of_nearest, nearest, n)
+
+        {clades, count, pointers, pairwise, diss, sizes} =
+          merge_clades(clades, count, pointers, pairwise, diss, sizes, links, n, update_fun)
+
+        {clades, count, pointers, pairwise, diss, sizes}
```
msluszniak, line 231:
```suggestion
        {clades, {count, pointers, pairwise, diss, sizes}}
```

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,476 @@
...
+      end
+
+    sizes = sizes[n..(2 * n - 2)]
+    perm = Nx.argsort(diss)
+    {clades[perm], diss[perm], sizes[perm]}
+  end
+
+  defnp merge_clades(clades, count, pointers, pairwise, diss, sizes, links, n, update_fun) do
+    {clades, count, pointers, pairwise, diss, sizes, _links} =
+      while {clades, count, pointers, pairwise, diss, sizes, links},
```
msluszniak, line 241:
```suggestion
    {{clades, count, pointers, pairwise, diss, sizes}, _} =
      while {{clades, count, pointers, pairwise, diss, sizes}, links},
```

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,476 @@
...
+            i <- 0..(Nx.size(links) - 1) do
+        # i < j because of how links is formed.
+        # i will become the new clade index and we "infinity-out" j.
+        j = links[i]
+
+        if j == n do
+          {clades, count, pointers, pairwise, diss, sizes, links}
```
msluszniak, line 248:
```suggestion
          {{clades, count, pointers, pairwise, diss, sizes}, links}
```

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,476 @@
...
+
+          # Update sizes
+          sc = sizes[i] + sizes[j]
+          sizes = Nx.indexed_put(sizes, Nx.stack([i, c]) |> Nx.new_axis(-1), Nx.stack([sc, sc]))
+
+          # Update dissimilarities
+          diss = Nx.indexed_put(diss, Nx.stack([count]), pairwise[i][j])
+
+          # Update pairwise
+          {pairwise, _, _, _, _, _} =
```
msluszniak, line 267:
```suggestion
          {pairwise, _} =
```

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,476 @@
...
+            while {pairwise, x = i, y = j, sa = sizes[i], sb = sizes[j], sc = sc},
```
msluszniak, line 268:
```suggestion
            while {pairwise, {x = i, y = j, sa = sizes[i], sb = sizes[j], sc = sc}},
```

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,476 @@
...
+
+                # TODO: do this in a single call?
+                pairwise =
+                  pairwise
+                  |> Nx.indexed_put(Nx.stack([x, z]), update)
+                  |> Nx.indexed_put(Nx.stack([z, x]), update)
+                  |> Nx.indexed_put(Nx.stack([y, z]), update)
+                  |> Nx.indexed_put(Nx.stack([z, y]), update)
+
+                {pairwise, x, y, sa, sb, sc}
```
msluszniak, line 286:
```suggestion
                {pairwise, {x, y, sa, sb, sc}}
```

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,476 @@
...
+          pairwise =
+            pairwise
+            |> Nx.put_slice([j, 0], Nx.reshape(infinities, {1, n}))
+            |> Nx.put_slice([0, j], Nx.reshape(infinities, {n, 1}))
+
+          # Update pointers
+          indices = [i, j] |> Nx.stack() |> Nx.new_axis(-1)
+          pointers = Nx.indexed_put(pointers, indices, Nx.stack([c, c]))
+
+          {clades, count + 1, pointers, pairwise, diss, sizes, links}
```
msluszniak, line 301:
```suggestion
          {{clades, count + 1, pointers, pairwise, diss, sizes}, links}
```

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,476 @@
...
+  # Dissimilarity functions
+
+  defnp pairwise_euclidean(%Nx.Tensor{} = x) do
+    x |> pairwise_euclidean_sq() |> Nx.sqrt()
+  end
+
+  defnp pairwise_euclidean_sq(%Nx.Tensor{} = x) do
+    sq = Nx.sum(x ** 2, axes: [1], keep_axes: true)
+    sq + Nx.transpose(sq) - 2 * Nx.dot(x, [1], x, [1])
+  end
```
msluszniak, line 326: These functions have been already implemented in Scholar ;)

billylanchantin: Oh where? I didn't see a public function for them.

lib/scholar/cluster/hierarchical.ex
billylanchantin: For these suggestions, is the idea that the "non-returns" get stuffed into an additional tuple? That way the match can ignore them with a single `_`?

msluszniak: Exactly :))

lib/scholar/cluster/hierarchical.ex
msluszniak: Here https://github.com/elixir-nx/scholar/blob/dacf106f63125a3b57866e19449fdf15cc7a27ca/lib/scholar/metrics/distance.ex#L553-L651

lib/scholar/cluster/hierarchical.ex
billylanchantin: Great pattern! I like it :) I did have to do quite a bit of "tuple matching" to get this working.

lib/scholar/cluster/hierarchical.ex
billylanchantin: Oh so it is! Thanks!

I was thrown off because it's not on hexdocs:

https://hexdocs.pm/scholar/Scholar.Metrics.Distance.html

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,476 @@
...
+      raise ArgumentError, "Must have a minimum of 3 data points, found: #{n}."
+    end
+
+    pairwise = dissimilarity_fun.(data)
+
+    if Nx.shape(pairwise) != {n, n} do
+      raise ArgumentError, "Pairwise must be a symmetric matrix."
+    end
+
+    {clades, diss, sizes} = dendrogram_fun.(pairwise, update_fun)
```
josevalim, line 193: Nitpick: we want to avoid going in and out of defn. So I would do this:

```suggestion
    {clades, diss, sizes} = dendrogram_fun.(data, dissimilarity_fun, update_fun)
```

The matrix shape check can be moved inside the `dendogram_fun`.

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,476 @@
...
+                dab = pairwise[[x, y]]
+                update = update_fun.(dac, dbc, dab, sa, sb, sc)
+
+                # TODO: do this in a single call?
+                pairwise =
+                  pairwise
+                  |> Nx.indexed_put(Nx.stack([x, z]), update)
+                  |> Nx.indexed_put(Nx.stack([z, x]), update)
+                  |> Nx.indexed_put(Nx.stack([y, z]), update)
+                  |> Nx.indexed_put(Nx.stack([z, y]), update)
```
josevalim, line 284: Yeah, I think you can do a Nx.concatenate of the stacks and of the updates.

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,476 @@
...
+  (Cluster labels are arbitrary, but deterministic.)
+
+  ## Examples
+
+      iex> data = Nx.tensor([[2], [7], [9], [0], [3]])
+      iex> model = Hierarchical.fit(data)
+      iex> Hierarchical.fit_predict(model, cluster_by: [num_clusters: 3])
+      Nx.tensor([0, 1, 1, 2, 0])
+  """
+  deftransform fit_predict(data_or_model, opts \\ [])
```
josevalim, line 398: This isn't really a `deftransform` function. They are all assuming we can access the values in the tensor and `defn` and `deftransform` cannot access values, only types and shapes. My suggestion is to call this function `labels` or similar, no `deftransform` or `defn`, and document that their goal is to return a high-level Elixir data structure representing the hierarchies (it can even be a map, not a tensor, because this is not working inside `defn` anyway).

That's the key philosophy inside Scholar/Nx: everything that can be done inside Nx, is explicitly done inside Nx. If you can't do it inside Nx, then that's fine too, but then it should be separated and marked as such!

josevalim: The philosophy is important for the GPU, because going in and out is often expensive, especially when we consider immutability. We can only be immutable inside `defn`.

josevalim (commented):
Brilliantly done!

I have done two nitpicks and one major suggestion. My suggestion is to rename `fit_predict` and make it return the most appropriate data type you consider (it doesn't have to be a tensor).

josevalim:
Also let's rebase so we can ship it!

lib/scholar/cluster/hierarchical.ex
billylanchantin: So my hesitation on this is that I've got `dissimilarity: :precomputed` in mind. Note how Scipy supports that option:

https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.linkage.html#scipy.cluster.hierarchy.linkage

> The input y may be either a 1-D condensed distance matrix or a 2-D array of observation vectors.

(Their "condensed distance matrix" is our pairwise.)

The idea is that building a dendrogram only requires pairwise, independent of what data was used to create it. If you have a custom dissimilarity fun, you can compute pairwise yourself and pass it in instead of the original data.

We can't do it yet, but once we support it we'll need to check pairwise independent of building the dendrogram.

lib/scholar/cluster/hierarchical.ex
billylanchantin: Thanks! I _think_ I got it now: [35d0ebe](https://github.com/elixir-nx/scholar/pull/187/commits/35d0ebe8df7b9fbf77be2bb7f8b304db5cae4522). (But IDK if it's actually more performant...)

billylanchantin:
@josevalim @msluszniak I need access to the pairwise functions. I was gonna `merge main`, but José said rebase. Which do y'all prefer?

lib/scholar/cluster/hierarchical.ex
billylanchantin: Awesome thanks for making those points! I'm definitely still cargo-culting much of the nuance.

Alright, what do we want the API to be? I tried to use `affinity_propagation` as my guide, but it's not a perfect fit.

We need `fit` to return a `%Hierarchical{}` struct so we can theoretically plot the dendrogram before we choose how to cluster. But other than that, there are a bunch of ways to organize things.

We can do basically what we have now:

* `deftransform fit` : `data: %Tensor{}` -> `model: %Hierarchical{}`
* `def label` : `model: %Hierarchical{}` -> `labels: %Tensor{}`

where `labels` is what `fit_predict` returns now.

We can make `labels` field of the `%Hierarchical{}` struct. Then `label` just sets that field and returns the altered `%Hierarchical{}`.

We can have `label` take a `return_type` option. Then you could either have it return a tensor of labels or a map of `%{cluster_label => list_of_member_indices}`.

We can have label return a `%HierarchicalClusters{}` struct that has both a `labels` field and a `clusters` field which would hold the map from the previous option.

I think all of those are reasonable options (and there are more). I'm happy to pick one! But I think you all likely know better what fits best with the ecosystem.

josevalim:
Any of rebase or merge is fine. Whatever you prefer!

lib/scholar/cluster/hierarchical.ex
josevalim: @billylanchantin then we can check for the shape on the dissimlarity: :precomputed branch or functions. In any case, the important is to call defn once.

lib/scholar/cluster/hierarchical.ex
josevalim: I am not sure if it is more performant either, so go with whatever reads better (I prefer the version with several indexed_put, there are fewer operations, your call).

lib/scholar/cluster/hierarchical.ex
josevalim: > deftransform fit : data: %Tensor{} -> model: %Hierarchical{}

Agreed.

And then let's add `labels_map` that returns `%{cluster_label => list_of_member_indices}`?

lib/scholar/cluster/hierarchical.ex
billylanchantin: [ad9ddb4](https://github.com/elixir-nx/scholar/pull/187/commits/ad9ddb47f77c15e129ed67a5f89f75b43a89508f)

lib/scholar/cluster/hierarchical.ex
billylanchantin: Overtaken by: https://github.com/elixir-nx/scholar/pull/187#discussion_r1404707992

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,483 @@
...
+        links = Nx.select(clades_selector and nearest > nearest_of_nearest, nearest, n)
+
+        {clades, count, pointers, pairwise, diss, sizes} =
+          merge_clades(clades, count, pointers, pairwise, diss, sizes, links, n, update_fun)
+
+        {{clades, diss, sizes}, {count, pointers, pairwise}}
+      end
+
+    sizes = sizes[n..(2 * n - 2)]
+    perm = Nx.argsort(diss, type: :u32)
```
josevalim, line 231: If you need this to be stable, please pass `stable: true`, as the default will change on Nx v0.7 to mirror Numpy. Perhaps best to pass it as true or false anyway to keep it explicit.

billylanchantin: Noted! I don't actually think it needs to be stable, but I'll make it explicit.

billylanchantin: [cb15a91](https://github.com/elixir-nx/scholar/pull/187/commits/cb15a91423ac6560fb8aed87ee320267a070234f)

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,483 @@
...
+        {next, pointers[next], pointers}
+      end
+
+    i
+  end
+
+  # Dissimilarity functions
+
+  defnp pairwise_euclidean(%Nx.Tensor{} = x) do
+    Scholar.Metrics.Distance.pairwise_euclidean(x, x)
```
msluszniak, line 315:
```suggestion
    Scholar.Metrics.Distance.pairwise_euclidean(x)
```

billylanchantin: Oh in that case we can just get rid of that `defnp` entirely. We'll point `dissimilarity_fun` at it directly up in `fit/2`.

billylanchantin: [885c5b5](https://github.com/elixir-nx/scholar/pull/187/commits/885c5b5bd33627aea53aa283836ee1d0fca189f2)

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,477 @@
...
+    {{clades, diss, sizes}, _} =
+      while {{clades, diss, sizes}, {count = 0, pointers, pairwise}}, count < n - 1 do
+        # Indexes of who I am nearest to
+        nearest = Nx.argmin(pairwise, axis: 1)
+
+        # Take who I am nearest to is nearest to
+        nearest_of_nearest = Nx.take(nearest, nearest)
+
+        # If the entry is pointing back at me, then we are a clade
+        clades_selector = Nx.equal(nearest_of_nearest, Nx.iota({n}))
```
msluszniak, line 218:
```suggestion
        clades_selector = nearest_of_nearest == Nx.iota({n})
```

msluszniak (approved):

lib/scholar/cluster/hierarchical.ex
billylanchantin: [9a2db56](https://github.com/elixir-nx/scholar/pull/187/commits/9a2db56753949f7e3faff55331bed4b28e9d426f)

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,460 @@
...
+
+  <!-- (This isn't ready yet.)
+
+  ## Examples
+
+  See the [_Hierarchical Clustering_](./doc/notebooks/hierarchical_clustering.livemd) Livebook for
+  a worked example with a dendrogram plot.
+
+  -->
+
```
josevalim, line 19:
```suggestion
```

msluszniak:
> Ok I think we're basically good to go!
> 
> One last thing: I'm a bit worried about the complexity. I think it's actually O(n2p⋅log⁡(n)). But I'm not confident.
> 
> ~I'd prefer no docs over incorrect docs. Unless anyone can say definitively what the complexity is, I want to drop the complexity discussion from the moduledoc.~
> 
> Updated: [9a2db56](https://github.com/elixir-nx/scholar/pull/187/commits/9a2db56753949f7e3faff55331bed4b28e9d426f)

Yeah, the time complexity seems legit to me ;)

lib/scholar/cluster/hierarchical.ex
```diff
@@ -0,0 +1,451 @@
...
+          # Update dissimilarities
+          diss = Nx.indexed_put(diss, Nx.stack([count]), pairwise[i][j])
+
+          # Update pointers
+          pointers = Nx.indexed_put(pointers, indices, Nx.stack([c, c]))
+
+          # Update pairwise
+          updates =
+            update_fun.(pairwise[i], pairwise[j], pairwise[i][j], sizes[i], sizes[j], sc)
+            |> Nx.indexed_put(indices, Nx.broadcast(:infinity, {2}))
```
josevalim, line 261:
```suggestion
          sa = sizes[i]
          sb = sizes[j]
          sc = sa + sb
          sizes = Nx.indexed_put(sizes, Nx.stack([i, c]) |> Nx.new_axis(-1), Nx.stack([sc, sc]))

          # Update dissimilarities
          diss = Nx.indexed_put(diss, Nx.stack([count]), pairwise[i][j])

          # Update pointers
          pointers = Nx.indexed_put(pointers, indices, Nx.stack([c, c]))

          # Update pairwise
          updates =
            update_fun.(pairwise[i], pairwise[j], pairwise[i][j], sa, sb, sc)
            |> Nx.indexed_put(indices, Nx.broadcast(:infinity, {2}))
```

josevalim:
Beautifully done!

josevalim:

PR #193, opened 2023-10-17
Add mean_tweedie_deviance and particular cases

0urobor0s:
This PR also adds the particular cases:
- `mean_poisson_deviance`
- `mean_gamma_deviance`

And updates `mean_square_error` to use `mean_tweedie_deviance` as well.

### DIFF
--- a/lib/scholar/metrics/regression.ex
+++ b/lib/scholar/metrics/regression.ex
@@ -157,7 +157,7 @@ defmodule Scholar.Metrics.Regression do
         0.18411168456077576
       >
   """
-  deftransform mean_tweedie_deviance(y_true, y_pred, power) do
+  defn mean_tweedie_deviance(y_true, y_pred, power) do
     mean_tweedie_deviance_n(y_true, y_pred, power)
   end
 
--- a/lib/scholar/metrics/regression.ex
+++ b/lib/scholar/metrics/regression.ex
@@ -11,7 +11,6 @@ defmodule Scholar.Metrics.Regression do
   any supported `Nx` compiler.
   """
 
-  import Nx, only: [is_tensor: 1]
   import Nx.Defn, except: [assert_shape: 2, assert_shape_pattern: 2]
   import Scholar.Shared
   import Scholar.Metrics.Distance
@@ -27,24 +26,7 @@ defmodule Scholar.Metrics.Regression do
     ]
   ]
 
-  mean_tweedie_deviance_schema = [
-    check_tensors: [
-      type: :boolean,
-      default: false,
-      doc: """
-      Flag indicating if tensor inputs should be checked to conform with the
-      necessary properties for the given power value.
-      """
-    ]
-  ]
-
-  mean_poisson_deviance_schema = mean_tweedie_deviance_schema
-  mean_gamma_deviance_schema = mean_tweedie_deviance_schema
-
   @r2_schema NimbleOptions.new!(r2_schema)
-  @mean_tweedie_deviance_schema NimbleOptions.new!(mean_tweedie_deviance_schema)
-  @mean_poisson_deviance_schema NimbleOptions.new!(mean_poisson_deviance_schema)
-  @mean_gamma_deviance_schema NimbleOptions.new!(mean_gamma_deviance_schema)
 
   # Standard Metrics
 
@@ -165,10 +147,6 @@ defmodule Scholar.Metrics.Regression do
   \end{cases}$$
   '''}
 
-  ## Options
-
-  #{NimbleOptions.docs(@mean_tweedie_deviance_schema)}
-
   ## Examples
 
       iex> y_true = Nx.tensor([1, 1, 1, 1, 1, 2, 2, 1, 3, 1], type: :u32)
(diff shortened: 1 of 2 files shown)


lib/scholar/metrics/regression.ex
```diff
@@ -133,6 +132,150 @@ defmodule Scholar.Metrics.Regression do
...
+  deftransform mean_tweedie_deviance(y_true, y_pred, power) do
+    message = "Mean Tweedie deviance with power=#{power} can only be used on "
+
+    case check_tweedie_deviance_power(y_true, y_pred, power) |> Nx.to_number() do
+      2 -> raise message <> "strictly positive y_pred."
+      4 -> raise message <> "non-negative y_true and strictly positive y_pred."
+      5 -> raise message <> "strictly positive y_true and strictly positive y_pred."
+      100 -> raise "Something went wrong, branch should never appear."
+      1 -> :ok
+    end
```
msluszniak, line 169: We cannot use runtime checks, because they would fail when function would be called in another `defn`

josevalim: You can test this by calling `Nx.Defn.jit(&mean_tweedie_deviance/3).(y_true, y_pred, 1)`. You will see that jit returns an expression and you can't convert it to number.

lib/scholar/metrics/regression.ex
```diff
@@ -133,6 +132,150 @@ defmodule Scholar.Metrics.Regression do
...
+  defnp mean_tweedie_deviance_n(y_true, y_pred, power) do
+    deviance =
+      cond do
+        power < 0 ->
+          2 *
+            (
+              Nx.pow(max(y_true, 0), 2 - power) / ((1 - power) * (2 - power))
+              -y_true * Nx.pow(y_pred, 1 - power) / (1 - power)
+              +Nx.pow(y_pred, 2 - power) / (2 - power)
+            )
```
josevalim, line 183: We can remove this clause as it is the same as the last?

msluszniak: Actually, they are not the same. In the first one, there is `max(y_true,0)`

lib/scholar/metrics/regression.ex
0urobor0s: Do you mean this:
```
iex> Nx.Defn.jit(&mean_tweedie_deviance/3).(y_true, y_pred, 1)
** (Protocol.UndefinedError) protocol String.Chars not implemented for #Nx.Tensor<
  s64

  Nx.Defn.Expr
  parameter a:2   s64
> of type Nx.Tensor (a struct). This protocol is implemented for the following type(s): Atom, BitString, Complex, Date, DateTime, Float, Hex.Solver.Assignment, Hex.Solver.Constraints.Empty, Hex.Solver.Constraints.Range, Hex.Solver.Constraints.Union, Hex.Solver.Incompatibility, Hex.Solver.PackageRange, Hex.Solver.Term, Integer, List, NaiveDateTime, Time, URI, Version, Version.Requirement
    (elixir 1.14.5) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir 1.14.5) lib/string/chars.ex:22: String.Chars.to_string/1
    (scholar 0.2.1) lib/scholar/metrics/regression.ex:161: Scholar.Metrics.Regression.mean_tweedie_deviance/3
    (nx 0.7.0-dev) lib/nx/defn/compiler.ex:158: Nx.Defn.Compiler.runtime_fun/3
    (nx 0.7.0-dev) lib/nx/defn/evaluator.ex:82: Nx.Defn.Evaluator.precompile/3
    (nx 0.7.0-dev) lib/nx/defn/evaluator.ex:60: Nx.Defn.Evaluator.__compile__/4
    (nx 0.7.0-dev) lib/nx/defn/evaluator.ex:53: Nx.Defn.Evaluator.__jit__/5
    iex:4: (file)
```

I thought that by using the `deftransform`, it could be done. That assumption from reading the docs, was the ideia behing the `check_tweedie_deviance_power` returning a tensor "error value". But from what I understand now, what you mean is that this happens when a `defn` function calls the `mean_tweedie_deviance` defined with `deftransform`.

If so, do you propose any idea/solution for it. In this case, the values need to be checked in order to guarantee that the results "can be calculated" as to have probabilistic meaning.

josevalim: You need to imagine that `defn` is running inside the GPU. Inside the GPU you can't have different shapes and you can't have exceptions. The best you can do is return NaN to signal that something went wrong OR have a separate function, that is a regular `def`, that receives a tensor to check it. In this case, the best is to assume you receive valid inputs and, if not, the value is unspecified.

0urobor0s: I think having the checks might be useful, at least for sanity check. I've added an option to allow for that check to happen in a `def` function, which by default is false (no check is made). Which means that by default, if the incorrect inputs for the given power value are passed to these functions, there is no guarantee of a result (with some NaNs appearing). Let me know if this can be done like this, or if you have any other concerns.

josevalim: Maybe a better solution then is to have `mean_tweedie_deviance` (no checking) and `mean_tweedie_deviance!` (with checking)?

lib/scholar/metrics/regression.ex
```diff
@@ -133,6 +150,188 @@ defmodule Scholar.Metrics.Regression do
...
+
+      power >= 2 ->
+        if nx_to_bool(Nx.greater(y_true, 0)) and nx_to_bool(Nx.greater(y_pred, 0)) do
+          :ok
+        else
+          raise message <> "strictly positive y_true and strictly positive y_pred."
+        end
+
+      true ->
+        raise "Something went wrong, branch should never appear."
```
josevalim, line 259: error messages start in lowercase and no dots at the end.

lib/scholar/metrics/regression.ex
0urobor0s: I was following the opts approach, but if possible I agree. In that case would `mean_tweedie_deviance!` be a `def` function?

josevalim: Yeah, it would be a `def`!

josevalim: That would be great, yeah!

0urobor0s: Ok, added it as well.

josevalim (approved):

msluszniak (approved):

josevalim:

PR #199, opened 2023-10-24
Added log loss classification metric

svarqq:
I based the implementation on [sklearn's](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.log_loss.html), with these main differences:
- Dropped labels as an optional argument, since it's mostly useful for string labels. Note as a result, the class index or class number of a value in `y_true` is always given by its ordered ranking over all `y_true` values (in the language of `sklearn`, the class index is always inferred). For example, if `y_true = [4, 0, 9, 4, 9]` the indices would be `[1, 0, 2, 1, 2]`. 
- No clipping of `y_prob` values to interval `(eps, 1-eps)`. That means `log(0)` can be evaluated to `-Inf` which would immediately cause a return of `Inf` for the function. I chose to keep this possibility because it immediately signals either that the client got inputs wrong, or that they should retrain their model/forecaster. Plus, I did try clipping to the `eps` interval using `Nx.Constants.epsilon/2` for the underlying floats in the tensor, and still got `Inf`'s returned.

### DIFF
--- a/lib/scholar/metrics/classification.ex
+++ b/lib/scholar/metrics/classification.ex
@@ -1263,7 +1263,7 @@ defmodule Scholar.Metrics.Classification do
   each class, from which the log loss is computed by averaging the negative log
   of the probability forecasted for the true class over a number of samples.
 
-  `y_true` should contain `num_classes` unique values, and the sum of `y_pred`
+  `y_true` should contain `num_classes` unique values, and the sum of `y_prob`
   along axis 1 should be 1 to respect the law of total probability.
 
   ## Options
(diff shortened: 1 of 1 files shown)


josevalim (approved):

msluszniak (approved):

lib/scholar/metrics/classification.ex
```diff
@@ -1233,6 +1255,93 @@ defmodule Scholar.Metrics.Classification do
...
+      iex> y_prob = Nx.tensor([[0.9, 0.1], [0.8, 0.2], [0.3, 0.7], [0.01, 0.99]])
+      iex> Scholar.Metrics.Classification.log_loss(y_true, y_prob, num_classes: 2)
+      #Nx.Tensor<
+        f32
+        0.17380733788013458
+      >
+
+      iex> Scholar.Metrics.Classification.log_loss(y_true, y_prob, num_classes: 2, normalize: false)
+      #Nx.Tensor<
+        f32
```
msluszniak, line 1285:
```suggestion

      iex> y_true = Nx.tensor([0, 0, 1, 1])
      iex> y_prob = Nx.tensor([[0.9, 0.1], [0.8, 0.2], [0.3, 0.7], [0.01, 0.99]])
      iex> Scholar.Metrics.Classification.log_loss(y_true, y_prob, num_classes: 2, normalize: false)
      #Nx.Tensor<
        f32
```

lib/scholar/metrics/classification.ex
```diff
@@ -1233,6 +1255,93 @@ defmodule Scholar.Metrics.Classification do
...
+  ## Examples
+
+      iex> y_true = Nx.tensor([0, 0, 1, 1])
+      iex> y_prob = Nx.tensor([[0.9, 0.1], [0.8, 0.2], [0.3, 0.7], [0.01, 0.99]])
+      iex> Scholar.Metrics.Classification.log_loss(y_true, y_prob, num_classes: 2)
+      #Nx.Tensor<
+        f32
+        0.17380733788013458
+      >
+
```
josevalim, line 1282:
```suggestion
```

lib/scholar/metrics/classification.ex
```diff
@@ -1233,6 +1255,92 @@ defmodule Scholar.Metrics.Classification do
...
+      #Nx.Tensor<
+        f32
+        0.17380733788013458
+      >
+      iex> Scholar.Metrics.Classification.log_loss(y_true, y_prob, num_classes: 2, normalize: false)
+      #Nx.Tensor<
+        f32
+        0.6952293515205383
+      >
+
```
josevalim, line 1287:
```suggestion
```

lib/scholar/metrics/classification.ex
```diff
@@ -1233,6 +1255,93 @@ defmodule Scholar.Metrics.Classification do
...
+      iex> Scholar.Metrics.Classification.log_loss(y_true, y_prob, num_classes: 2, normalize: false)
+      #Nx.Tensor<
+        f32
+        0.6952293515205383
+      >
+
+      iex> weights = Nx.tensor([0.7, 2.3, 1.3, 0.34])
+      iex(361)> Scholar.Metrics.Classification.log_loss(y_true, y_prob, num_classes: 2, sample_weights: weights)
```
msluszniak:
```suggestion

      iex> y_true = Nx.tensor([0, 0, 1, 1])
      iex> y_prob = Nx.tensor([[0.9, 0.1], [0.8, 0.2], [0.3, 0.7], [0.01, 0.99]])
      iex> weights = Nx.tensor([0.7, 2.3, 1.3, 0.34])
      iex(361)> Scholar.Metrics.Classification.log_loss(y_true, y_prob, num_classes: 2, sample_weights: weights)
```

josevalim:

PR #203, opened 2023-11-02
Matthews Correlation Coefficient

paulsullivanjr:
MCC implementation.


josevalim:
Thank you @paulsullivanjr . This looks good to me, we just need to figure out the namespace to define this, as I would like to avoid one metric per module. Any suggestions?

paulsullivanjr:
> Thank you @paulsullivanjr . This looks good to me, we just need to figure out the namespace to define this, as I would like to avoid one metric per module. Any suggestions?

Yeah I agree with avoiding one metric per module. I wonder if this one should be moved into metrics/classification.ex. I'm not sure on the future plans for other metrics other than what's defined in the GitHub issue so there may be a better option.

polvalente:
I'm not sure about classification. This being a correlation coefficient sounds like a metric of similarity to me.

For instance, the Pearson correlation coefficient is quite similar to a cosine similarity of sorts (as in both are derived from some sort of inner product)

paulsullivanjr:
That makes sense, much better idea.

lib/scholar/metrics/mcc.ex
```diff
@@ -0,0 +1,63 @@
...
+    Nx.sum(Nx.equal(y_true, 0) * Nx.equal(y_pred, 0))
+  end
+
+  defnp calculate_false_positives(y_true, y_pred) do
+    Nx.sum(Nx.equal(y_true, 0) * Nx.equal(y_pred, 1))
+  end
+
+  defnp calculate_false_negatives(y_true, y_pred) do
+    Nx.sum(Nx.equal(y_true, 1) * Nx.equal(y_pred, 0))
+  end
```
msluszniak, line 62: Look that those functions are already implemented in Scholar [here](https://github.com/elixir-nx/scholar/blob/3d9e0156a884e1407241e9018b09c2083c16f7d9/lib/scholar/metrics/classification.ex#L395)

lib/scholar/metrics/mcc.ex
```diff
@@ -0,0 +1,63 @@
...
+      )
+
+    zero_tensor = Nx.tensor([0.0], type: :f32)
+
+    if Nx.all(
+         Nx.logical_and(
+           Nx.equal(true_positives, zero_tensor),
+           Nx.equal(true_negatives, zero_tensor)
+         )
+       ) do
```
msluszniak, line 37:
```suggestion
    if Nx.all(
           true_positives == zero_tensor and
           true_negatives ==  zero_tensor
       ) do
```

lib/scholar/metrics/mcc.ex
```diff
@@ -0,0 +1,63 @@
...
+      Nx.tensor([-1.0], type: :f32)
+    else
+      Nx.select(
+        Nx.equal(mcc_denominator, zero_tensor),
+        zero_tensor,
+        Nx.divide(mcc_numerator, mcc_denominator)
+      )
```
msluszniak, line 44:
```suggestion
      Nx.select(
        mcc_denominator == zero_tensor,
        zero_tensor,
        mcc_numerator / mcc_denominator
      )
```

test/scholar/metrics/mcc_test.exs
```diff
@@ -0,0 +1,47 @@
...
+      y_pred = Nx.tensor([1, 0, 1, 1, 1])
+      mcc_tensor = MCC.compute(y_true, y_pred)
+      mcc_value = Nx.reshape(mcc_tensor, {1}) |> Nx.to_list() |> hd()
+      assert mcc_value == 0.6123723983764648
+    end
+
+    test "returns 0 when TP, TN, FP, and FN are all 0" do
+      y_true = Nx.tensor([0, 0, 0, 0, 0])
+      y_pred = Nx.tensor([0, 0, 0, 0, 0])
+      assert MCC.compute(y_true, y_pred) == Nx.tensor([0.0], type: :f32)
```
msluszniak, line 44: Could you rewrite the rest of the tests as you write it here? `reshape`, `to_list`, and `hd` aren't necessary ;)

lib/scholar/metrics/mcc.ex
paulsullivanjr: Gotcha, sounds good. Would you like me to make those functions public?

josevalim: @paulsullivanjr I think instead we should move this function to Classification, no?

josevalim: Nevermind, @polvalente said it is a similarity function. Although the documentation says " provides a measure of the quality of binary classifications". So which one is it? 🤔

polvalente: I think classification is acceptable, especially if this shares common private functions with that module

msluszniak (approved):
LGTM!

lib/scholar/metrics/classification.ex
```diff
@@ -1448,4 +1448,41 @@ defmodule Scholar.Metrics.Classification do
     assert_rank!(y_true, 1)
     assert_same_shape!(y_true, y_pred)
   end
+
+  @doc """
+  Matthews Correlation Coefficient (MCC) provides a measure of the quality of binary classifications.
```
josevalim, line 1453:
```suggestion
  Matthews Correlation Coefficient (MCC) provides a measure of the quality of binary classifications.

```

josevalim:

PR #204, opened 2023-11-02
Improve cubic interpolation

### LINKED ISSUE #200: Cublic Splines works too slow
One of the reasons is that currently, we use `Nx.solve` which is roughly O(n^3), but for tridiagonal matrices, there is a [linear algorithm](https://en.wikipedia.org/wiki/Tridiagonal_matrix_algorithm) that solves the problem.

msluszniak:
Closes #200


josevalim (approved):
Run `mix format` and it looks good to me!

polvalente (approved):

josevalim:

PR #206, opened 2023-11-05
KDTrees

josevalim:


lib/scholar/neighbors/kd_tree.ex
```diff
@@ -0,0 +1,318 @@
...
+        Nx.from_binary(acc, :u32)
+      else
+        Nx.argsort(tensor[[.., 0]], direction: :desc, type: :u32)
+      end
+
+    %__MODULE__{levels: levels, indexes: indexes}
+  end
+
+  defp recur([{_i, %Nx.Tensor{shape: {1}} = leaf} | rest], next, acc, tensor, level, levels, opts) do
+    [leaf] = Nx.to_flat_list(leaf)
```
polvalente, line 73: Any special reason to use `Nx.to_flat_list` instead of `Nx.to_number` here?

josevalim: to_number only works on scalars.

josevalim:
GPU, N=1_000_00 (ht @polvalente)

```
Operating System: Linux
CPU Information: 13th Gen Intel(R) Core(TM) i9-13900KS
Number of Available Cores: 32
Available memory: 125.53 GB
Elixir 1.15.1
Erlang 26.0.2

Benchmark suite executing with the following configuration:
warmup: 2 s
time: 10 s
memory time: 2 s
reduction time: 0 ns
parallel: 1
inputs: none specified
Estimated total run time: 28 s

Benchmarking banded 1M ...
Benchmarking unbanded 1M ...
    
Name                  ips        average  deviation         median         99th %
banded 1M           55.01    0.00030 min     ±2.15%    0.00030 min    0.00032 min
unbanded 1M        0.0167       1.00 min     ±0.00%       1.00 min       1.00 min

Comparison: 
banded 1M           55.01
unbanded 1M        0.0167 - 3301.18x slower +1.00 min

Memory usage statistics:

Name           Memory usage
banded 1M        0.00001 GB
unbanded 1M         4.44 GB - 825455.53x memory usage +4.44 GB

**All measurements for memory usage were the same**
```

krstopro:
What is the dimension of the data you used for benchmarks?

josevalim:
To add to your question/answer, the dimension should not impact the results, as we don't store the tensors and we only cycle through them.

lib/scholar/neighbors/kd_tree.ex
```diff
@@ -0,0 +1,370 @@
...
+  import Nx.Defn
+
+  @derive {Nx.Container, keep: [:levels], containers: [:indexes, :data]}
+  @enforce_keys [:levels, :indexes, :data]
+  defstruct [:levels, :indexes, :data]
+
+  @doc """
+  Builds a KDTree without known min-max bounds.
+
+  If your tensor has a known band (for example, -1 and 1),
```
jonatanklosko, line 39: Is this the wording they use in one of the papers? I would expect "bounds" and `bounded/2`.

test/scholar/neighbors/kd_tree_test.exs
```diff
@@ -0,0 +1,72 @@
...
+      [45, 40],
+      [62, 69],
+      [53, 67]
+    ])
+  end
+
+  describe "unbanded" do
+    test "sample" do
+      assert %Scholar.Neighbors.KDTree{levels: 4, indexes: indexes} =
+               Scholar.Neighbors.KDTree.unbanded(example(), compiler: EXLA.Defn)
```
jonatanklosko, line 23: ?

```suggestion
               Scholar.Neighbors.KDTree.unbanded(example(), compiler: EXLA)
```

test/test_helper.exs
```diff
@@ -0,0 +1 @@
+Application.ensure_all_started(:exla)
```
jonatanklosko, line 1: Why do we need to start explicitly?

josevalim: Earlier Elixir versions did not start optional applications.

lib/scholar/neighbors/kd_tree.ex
josevalim: No paper uses it, the idea is that it would reflect "frequency bands" but I think bound and unbound is more common. It may be a bit confusing because they are not really unbound and more of a question if the bounds are known and reasonable.

test/scholar/neighbors/kd_tree_test.exs
josevalim: Good call, fixed.

josevalim:

PR #211, opened 2023-11-23
Knn via kdtree

msluszniak:
I need to add a testing module and benchmarks, but as far as I've been testing it, it works fine, and is quick where classical KNN runs out of memory (100000x4) :))

### DIFF
--- a/benchmarks/knn.exs
+++ b/benchmarks/knn.exs
@@ -1,0 +1,27 @@
+# mix run benchmarks/knn.exs
+Nx.global_default_backend(EXLA.Backend)
+Nx.Defn.global_default_options(compiler: EXLA)
+
+key = Nx.Random.key(System.os_time())
+
+inputs_knn = %{
+  "100x10" => elem(Nx.Random.uniform(key, 0, 100, shape: {100, 10}), 0),
+  "1000x10" => elem(Nx.Random.uniform(key, 0, 1000, shape: {1000, 10}), 0),
+  "10000x10" => elem(Nx.Random.uniform(key, 0, 10000, shape: {10000, 10}), 0)
+}
+
+Benchee.run(
+  %{
+    "kdtree" => fn x ->
+      kdtree = Scholar.Neighbors.KDTree.bounded(x, Nx.axis_size(x, 0))
+      Scholar.Neighbors.KDTree.query(kdtree, x, k: 4)
+    end,
+    "brute force knn" => fn x ->
+      model = Scholar.Neighbors.KNearestNeighbors.fit(x, Nx.broadcast(1, {Nx.axis_size(x, 0)}), num_classes: 2, num_neighbors: 4)
+      Scholar.Neighbors.KNearestNeighbors.k_neighbors(model, x)
+    end
+  },
+  time: 10,
+  memory_time: 2,
+  inputs: inputs_knn
+)
--- a/benchmarks/knn.exs
+++ b/benchmarks/knn.exs
@@ -13,8 +13,8 @@ inputs_knn = %{
 Benchee.run(
   %{
     "kdtree" => fn x ->
-      kdtree = Scholar.Neighbors.KDTree.bounded(x, Nx.axis_size(x, 0))
-      Scholar.Neighbors.KDTree.query(kdtree, x, k: 4)
+      kdtree = Scholar.Neighbors.KDTree.fit_bounded(x, Nx.axis_size(x, 0))
+      Scholar.Neighbors.KDTree.predict(kdtree, x, k: 4)
     end,
     "brute force knn" => fn x ->
       model = Scholar.Neighbors.KNearestNeighbors.fit(x, Nx.broadcast(1, {Nx.axis_size(x, 0)}), num_classes: 2, num_neighbors: 4)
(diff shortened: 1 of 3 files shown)


lib/scholar/neighbors/kd_tree.ex
```diff
@@ -171,7 +181,7 @@ defmodule Scholar.Neighbors.KDTree do
       end
 
     k = rem(level, dims)
-    Nx.argsort(tensor[[.., k]] + band * tags, type: :u32)
+    Nx.argsort(tensor[[.., k]] + band * tags, type: :u32) |> Nx.as_type(:s64)
```
msluszniak, line 184: This is a temporary change, because when I ran the code with `EXLA` I got the following type mismatch:

```
** (RuntimeError) Dynamic slice start indices must all have the same shape, got mismatching indices with shapes u32[] and s64[].
    (exla 0.7.0-dev) lib/exla/op.ex:876: EXLA.Op.unwrap!/1
    (exla 0.7.0-dev) lib/exla/op.ex:121: EXLA.Op.get_shape/1
    (exla 0.7.0-dev) lib/exla/defn.ex:1043: EXLA.Defn.to_operator/4
    (exla 0.7.0-dev) lib/exla/defn.ex:646: EXLA.Defn.cached_recur_operator/4
    (exla 0.7.0-dev) lib/exla/defn.ex:490: EXLA.Defn.recur_operator/3
    (elixir 1.15.5) lib/enum.ex:1819: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (elixir 1.15.5) lib/enum.ex:1819: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    iex:10: (file)
```

lib/scholar/neighbors/kd_tree.ex
```diff
@@ -347,4 +361,156 @@ defmodule Scholar.Neighbors.KDTree do
...
+    k = opts[:k]
+    node = Nx.as_type(root(), :s64)
+    {size, dims} = Nx.shape(tree.data)
+    nearest_neighbors = Nx.broadcast(Nx.s64(0), {k})
+    distances = Nx.broadcast(Nx.Constants.infinity(), {k})
+    visited = Nx.broadcast(Nx.u8(0), {size})
+    down = 0
+    up = 1
+    mode = down
+    cnt = 0
```
msluszniak, line 428: This variable was only for debugging purpose, I'll delete it

lib/scholar/neighbors/kd_tree.ex
```diff
@@ -347,4 +361,156 @@ defmodule Scholar.Neighbors.KDTree do
...
+                  end
+
+                # Should be not reachable
+                true ->
+                  {node, i, visited, nearest_neighbors, distances, mode}
+              end
+
+            # Should be not reachable
+            true ->
+              {node, i + 1, visited, nearest_neighbors, distances, down}
```
msluszniak, line 508:
```suggestion
              {node, i , visited, nearest_neighbors, distances, mode}
```

msluszniak:
In fact, the exact KDtree knn search scales poorly with k. Calculating results for {100000, 10} takes much longer than {100000, 4}.

polvalente:
> In fact, the exact KDtree knn search scales poorly with k. Calculating results for {100000, 10} takes much longer than {100000, 4}.

Maybe that has to do with compounding factors of 100000 for each k unit instead of k itself?

msluszniak:
Yes, but it's not only our problem, this is a general problem of the exact knn using dktree. It's more like eh, indeed works slower 😅 
image

msluszniak:
And the natural continuation is to check the approximated versions which I researching now

josevalim:
Beautifully done, we just need to discuss the API. I can think of two options:

1. Rename `KDTree.unbound` and `KDTree.bound` to `KDTree.fit_bound` and `KDTree.fit_unbound`, rename `query` to `predict`

2. Keep `KDTree.unbound` and `KDTree.bound` but then pass the KDTree as first argument to `KNearestNeighbours.fit` and use it on `predict`

Thoughts?

msluszniak:
I think that I like the first option more. It will separate logically the brute force implementation of KNN with the KDTrees

josevalim:
So let’s do it!

benchmarks/knn.exs
```diff
@@ -0,0 +1,27 @@
...
+  "100x10" => elem(Nx.Random.uniform(key, 0, 100, shape: {100, 10}), 0),
+  "1000x10" => elem(Nx.Random.uniform(key, 0, 1000, shape: {1000, 10}), 0),
+  "10000x10" => elem(Nx.Random.uniform(key, 0, 10000, shape: {10000, 10}), 0)
+}
+
+Benchee.run(
+  %{
+    "kdtree" => fn x ->
+      kdtree = Scholar.Neighbors.KDTree.bounded(x, Nx.axis_size(x, 0))
+      Scholar.Neighbors.KDTree.query(kdtree, x, k: 4)
```
josevalim, line 17: We have to update the benchmark too!

josevalim (approved):
One tiny comment and ship it

Issue #207, opened 2023-11-05
Alternatives to kNN + t-SNE


josevalim:
We don't need to implement all options below but enough to have some efficient alternatives to the linear/brute force versions we have today. My suggestion is to pick whatever you find to be simplest.

Alternatives to kNN:

* [x] kNN with kd-trees
* [x] Random projection trees/forests (see #215)
* [x] [kNNDescent](https://github.com/elixir-nx/scholar/issues/189#issuecomment-1791453422)

Alternatives to tSNE:

* [x] [Largevis](https://arxiv.org/abs/1602.00370) (comes with a kNNG algorithm as a bonus)
* [x] [TriMap](https://github.com/google-research/google-research/tree/master/trimap)
* [ ] [PaCMAP](https://github.com/YingfanWang/PaCMAP)

Bonus:

* [ ] A notebook comparing the outputs of different neighbors algorithms ([example](https://www.researchgate.net/figure/Application-of-t-SNE-LargeVis-UMAP-TriMAP-and-PaCMAP-to-the-Mammoth-dataset-In_fig1_346766201))

josevalim:
I would have to see the algorithm in order to give feedback, but realistically speaking, I would only be able to look into this in 3 weeks (going on holidays).

josevalim:
My suggestion for now is to question the assumption if you really need to remove the duplicates. [In this algorithm here](https://github.com/elixir-nx/scholar/issues/189), the `updateNN` operation is idempotent, so you could call it for duplicates without loss or repeated information.

jonatanklosko:
> It boils down to taking first k distinct elements from a tensor.

@krstopro

1. Does it have to be k first going left-to-right or can it be any subset (specifically sorted)?
2. Is it guaranteed there are k distinct elements in the tensor? If not, what should happen?
3. Is the subset of k elements the final result, or is there some immediate operation we want to do with these elements?

krstopro:
> > It boils down to taking first k distinct elements from a tensor.
> 
> @krstopro
> 
> 1. Does it have to be k first going left-to-right or can it be any subset (specifically sorted)?
> 2. Is it guaranteed there are k distinct elements in the tensor? If not, what should happen?
> 3. Is the subset of k elements the final result, or is there some immediate operation we want to do with these elements?

@jonatanklosko 

[...]

PR #215, opened 2023-12-09
Add Random Projection Forests

krstopro:
Adds Random Projection Forest as in [Random projection trees and low dimensional manifolds](https://cseweb.ucsd.edu/~dasgupta/papers/rptree-stoc.pdf).
Things I am not sure about:
- Function used to construct the forest is called `grow`. Perhaps I should have used `fit` as in rest of `Scholar`?
- Should arguments `num_trees` and `min_leaf_size` be passed as options (and validated)?

Docstrings are missing, I will gladly add them when the questions above are answered. More unit tests might be needed as well.


lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -0,0 +1,140 @@
...
+
+    {hyperplanes, _key} =
+      Nx.Random.normal(key, type: to_float_type(tensor), shape: {num_trees, depth, dim})
+
+    medians = Nx.broadcast(:nan, {num_trees, num_nodes})
+    # TODO: Maybe rename acc to indices
+    acc = Nx.broadcast(Nx.u32(0), {num_trees, size})
+    root = 0
+    start_index = 0
+    indices = Nx.iota({size}, type: :u32) |> Nx.new_axis(0) |> Nx.broadcast({num_trees, size})
```
msluszniak, line 47:
```suggestion
    indices = Nx.iota({1, size}, type: :u32) |> Nx.broadcast({num_trees, size})
```

lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -0,0 +1,140 @@
...
+    median =
+      if rem(size, 2) == 1 do
+        proj[[.., div(size, 2)]]
+      else
+        mid = Nx.slice_along_axis(proj, div(size, 2) - 1, 2, axis: 1)
+        Nx.mean(mid, axes: [1])
+      end
+
+    indices = Nx.take_along_axis(indices, sorted_indices, axis: 1)
+    left_size = div(size, 2) + rem(size, 2)
```
msluszniak, line 117: I would move div(size, 2) and rem(size, 2) to a separate variables since they are used multiple times

msluszniak:
> Should arguments num_trees and min_leaf_size be passed as options (and validated)?
Yes, I think it will be better to pass these args via opts

> Function used to construct the forest is called grow. Perhaps I should have used fit as in rest of Scholar?
I don't have a strong preference, but I think it will be consistent if we rename it to `fit`

msluszniak (approved):

josevalim:
Thank you @krstopro! This is a great starting point, the big question is if we can fully convert this to `defn` or not. Unfortunately I will only be able to look at the paper much later, so I don't have an answer right now. If you have any thoughts, please let us know!

josevalim:
@krstopro I had some time to look into this and I *think* we may be able to move this to `defn`. The paper says that kdtree and rptree are the same, except they choose a different splitting rule. However, rptree also wants to customize the `min_size`. Perhaps we could submit both of these changes to the current KDTree implementation?

1. First allow the splitting rule to be passed as an anonymous function to both `fit_bounded` and `fit_unbounded`.

2. Change the construction algorithm and the KDTree representation to store a `min_size` and build leaf nodes of at least `min_size`.

Given 2 is much harder than one, perhaps I would start with it first. :)

josevalim:
One last comment, I took a look at what is the cost of implementing task item 2 and this parameter exists in scikit-learn as `leafsize`. All entries inside the `leafsize` in a regular KDtree are brute-forced in the query algorithm.

The implementation for `fit_bounded` is actually quite trivial, we just need to compute the number of effective levels to traverse. That's because the `leafsize` does not actually change the result, just how we traverse it.

For example, imagine we are storing 0..9 into the kdtree. With leafsize=1, we will have:

```
         6
     3       8
   1   5   7   9
  0 2 4
```

with leafsize=2 and leafsize=3, we have the same results too. The only difference is with leafsize=4:

```
         6
     3       8
  012 45   7   9
```

And we can easily achieve this by changing the algorithm to stop earlier. The hardest part would be knowing how many children there are in the left side of `3` (and its right side), but the original algorithm already implements this as `subtree_size`. So overall we need to:

1. Change `fit_unbounded` to accept a leafsize (@krstopro you already did this in your algorithm, so perhaps you can submit a PR?)

2. Change `fit_bounded` to accept a leafsize (I will submit a PR for this soon)

3. Change querying to consider the leafsize. The trick is to detect if we are in the last level. If so, you will read `subtree_size` entries and compare them using bruteforce.

krstopro:
Thanks for the feedback @josevalim.

This might be possible to write fully within `defn`; I went with the unbounded version first until I see the accuracy - it seems fine.
I am still trying the fully `defn` version, but there are some issues with it.

> One last comment, I took a look at what is the cost of implementing task item 2 and this parameter exists in scikit-learn as leafsize. All entries inside the leafsize in a regular KDtree are brute-forced in the query algorithm.

There might be a problem with querying such a tree (or even forest) in parallel in parallel if leaves are of different size. I am avoiding this by making the forest complete and making leaf size equal for all the leaves (left subtree might be larger by 1 than the right subtree, and when querying the right leaf we might query 1 element from the left leaf; this is no issue at all).

josevalim:
Ah, my comment was written from the perspective that leafsize is a maximum, not a minimum. But it is easy to flip it around and make it a minimum. 

> There might be a problem with querying such a tree (or even forest) in parallel if leaves are of different size.

Perhaps a separate algorithm is indeed best, because the current one leads to left balanced trees which will lead to many more elements on the left side.

The bounded algorithm is not complicated. The hardest part is deriving the formulas for the pivot (both bounded_segment_begin and bounded_subtree_size) but perhaps there are already definitions for these elsewhere.

lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -0,0 +1,212 @@
...
+       ) do
+    size = Nx.axis_size(indices, 1)
+
+    if size > leaf_size do
+      hyperplane = hyperplanes[[.., level]]
+
+      {median, left_indices, right_indices} =
+        Nx.Defn.jit_apply(&split(&1, &2, &3), [tensor, indices, hyperplane])
+
+      medians = Nx.put_slice(medians, [0, node], Nx.new_axis(median, 1))
```
josevalim, line 156: Move this inside `split`. We should have as many operations inside `defn` as possible.

josevalim:
> Function used to construct the forest is called grow. Perhaps I should have used fit as in rest of Scholar?

Yes!

> Should arguments num_trees and min_leaf_size be passed as options (and validated)?

As far as I understand they are already currently validated and accepted as options. I would only pass a default of 1  to `min_leaf_size`.

So overall, this looks great. The biggest question is indeed only the `defn` parts.

lib/scholar/neighbors/random_projection_forest.ex
krstopro: This is easy to do, but I don't see exactly why is it better given that `put_slice` is `Nx` function. Can I ask for an explanation?

krstopro:
> @krstopro btw, can you confirm that, for a given size N and min_leaf_size, this is how your trees look like

Something different. Here the nodes do not correspond to the points in the dataset. `indices` field specifies how the points in leaves are arranged. For the example above, if all the `hyperplanes` are positive indices will be `1 2 ... 20`. The first five points belongs to the first leaf, second five to the second, etc.

lib/scholar/neighbors/random_projection_forest.ex
josevalim: Nx in Elixir land needs to preserve the immutable semantics of the language. So every time you receive an output, the output will be fully copied into memory. In this case, with the two operations in the snippet, you are making a full copy of `median`, `left_indices`, `right_indices`, and `medians`. By moving this operation inside, at least you don't need to copy `median`.

This is exactly what moving the whole thing to a `defn` is beneficial. You don't have to worry about the copying of variables in a loop.

krstopro: I see, thanks.

krstopro:
Alright, the bounded version had been added. There are still some TODOs (including documentation and unit tests), but it seems to be working fine. What I am not sure about is how to handle amplitude; see the comment below.

krstopro:
For example, if
```
key = Nx.Random.key(12)
{x, _key} = Nx.Random.uniform(key, shape: {135, 1})
```
then `x[79] = 0.12850868701934814` and `x[133] = 0.1285092830657959`. The last hyperplane in the forest created with `fit_bounded(x, min_leaf_size: 1, num_trees: 1, key: key)` will be `-1.2941265106201172` and the projections of `x[79]` and `x[133]` will be `-0.1663064956665039` and `-0.166307270526886` respectively.
Furthermore, `amplitude = -1.2941265106201172` and `band = 2.2841360569000244` as they are calculated now and `tags` are `43` for both points on `level = 6`. We then get that `band * tags + proj` is equal to `98.05154418945312` for both points.

Similarly, for kd-trees `amplitude = 0.992280125617981`, `band = 1.992280125617981`, and if `tags` are `43` for both points (some other number might work as well) we get that `tensor[[.., 0]] + band * tags` is equal to `85.79655456542969` for both points.

I am not exactly sure how to proceed with this happening. Any comments?

msluszniak:
> For example, if
> 
> ```
> key = Nx.Random.key(12)
> {x, _key} = Nx.Random.uniform(key, shape: {135, 1})
> ```
> 
> then `x[79] = 0.12850868701934814` and `x[133] = 0.1285092830657959` (they are very close). The last hyperplane in the forest created with `fit_bounded(x, min_leaf_size: 1, num_trees: 1, key: key)` will be `-1.2941265106201172` and the projections of `x[79]` and `x[133]` will be `-0.1663064956665039` and `-0.166307270526886` respectively. Furthermore, `amplitude = -1.2941265106201172` and `band = 2.2841360569000244` as they are calculated now and `tags` are `43` for both points on `level = 6`. We then get that `band * tags + proj` is equal to `98.05154418945312` for both points.
> 
> Similarly, for kd-trees `amplitude = 0.992280125617981`, `band = 1.992280125617981`, and if `tags` are `43` for both points (some other number might work as well) we get that `tensor[[.., 0]] + band * tags` is equal to `85.79655456542969` for both of them.
> 
> I am not exactly sure how to proceed with this happening. Any thoughts?

Maybe we should use only `:f64` for higher precision of comparisons?

josevalim:
Or alternatively, could we avoid mixing values without imprecision? 

* Can you use exact 2.0 as the band for the random case? Would that help?

* What about increasing the amplitude of Nx.Random.uniform itself?

* Or, if what you need is to randomly sort the input, what if you use Nx.Random.shuffle(Nx.iota({n})) instead of working with floats? This will give you an exact integer for each element.

krstopro:
@msluszniak @josevalim 

> Maybe we should use only :f64 for higher precision of comparisons?

This should work, but I would try saving the memory.

> Can you use exact 2.0 as the band for the random case? Would that help?

You mean taking the ceiling of the band? Generally, we calculate `band` as `amplitude + 1` while any other positive value instead of 1 should work, right?. Maybe we should use the minimum value for the type used and then take the ceiling?

> What about increasing the amplitude of Nx.Random.uniform itself?

This is another possibility.

> Or, if what you need is to randomly sort the input, what if you use Nx.Random.shuffle(Nx.iota({n})) instead of working with floats? This will give you an exact integer for each element.

No, they need to be sorted by `band * tags + proj` (or `band * tags + tensor[[.., k]]` for k-d trees).

josevalim:
> You mean taking the ceiling of the band?

The ceiling but make it a power of two, so it doesn't use mantissa bits of the floating point.

Btw, why do you need to compute the band on every iteration?

krstopro:
> The ceiling but make it a power of two, so it doesn't use mantissa bits of the floating point.

I see, that could work. The only issue is that we need to 

> Btw, why do you need to compute the band on every iteration?

I don't - I just used it as a temporary solution. I want to make amplitude the argument of the function as you did for k-d trees. The issue was that I needed to compute the amplitude of the projections, not the original tensors. However, there is a way to bound the absolute value of the projections.

There is one thing I am not sure about. Does the amplitude in k-trees
https://github.com/elixir-nx/scholar/blob/b9ff291dc2fc0f10347841f2e5400ae06420064e/lib/scholar/neighbors/kd_tree.ex#L267-L278
assume that the tensor is centered, i.e. zero mean?

krstopro:
> There is one thing I am not sure about. Does the amplitude in k-trees
> 
> https://github.com/elixir-nx/scholar/blob/b9ff291dc2fc0f10347841f2e5400ae06420064e/lib/scholar/neighbors/kd_tree.ex#L267-L278
> 
> 
> assume that the tensor is centered, i.e. zero mean?

Nevermind, I think I got it.

krstopro:
Setting band to the power of two doesn't work as it still needs to be multiplied by tags. Perhaps the best way to proceed is to increase the amplitude of the data.

krstopro:
And even that doesn't seem to be working. :(

josevalim:
I will try to read the paper again and see if I come up with ideas.

krstopro:
> I will try to read the paper again and see if I come up with ideas.

It might be fine as it is now, because these points are rather close anyway and they will end up in the same leaf.

krstopro:
To add, I think k-d trees as you implemented in https://github.com/elixir-nx/scholar/blob/main/lib/scholar/neighbors/kd_tree.ex will have the same problem.

josevalim:
It may be that indeed you need to either bump the precision or the amplitude to not fall within numerical instability of floating points. :(

I have also written an exponential version of tags but I don't think it will make a difference:

```
diff --git a/lib/scholar/neighbors/kd_tree.ex b/lib/scholar/neighbors/kd_tree.ex
index 47d7ae0..64bbd4f 100644
--- a/lib/scholar/neighbors/kd_tree.ex
+++ b/lib/scholar/neighbors/kd_tree.ex
@@ -185,12 +185,12 @@ defmodule Scholar.Neighbors.KDTree do
     band = amplitude + 1
     tags = Nx.broadcast(Nx.u32(0), {size})
 
-    {level, tags, _tensor, _band} =
-      while {level = Nx.u32(0), tags, tensor, band}, level < levels - 1 do
+    {level, tags, _tensor, _band, _mask} =
+      while {level = Nx.u32(0), tags, tensor, band, mask = tags}, level < levels - 1 do
         k = rem(level, dims)
-        indices = Nx.argsort(tensor[[.., k]] + band * tags, type: :u32, stable: true)
-        tags = update_tags(tags, indices, level, levels, size)
-        {level + 1, tags, tensor, band}
+        indices = Nx.argsort(tensor[[.., k]] + band * print_value(mask), type: :u32, stable: true)
+        {tags, mask} = update_tags(tags, indices, level, levels, size)
+        {level + 1, tags, tensor, band, mask}
       end
 
     k = rem(level, dims)
@@ -204,19 +204,26 @@ defmodule Scholar.Neighbors.KDTree do
       bounded_segment_begin(tags, levels, size) +
         bounded_subtree_size(left_child(tags), levels, size)
 
-    Nx.select(
-      pos < (1 <<< level) - 1,
-      tags,
+    filter = pos < (1 <<< level) - 1
+
+    tags =
       Nx.select(
-        pos < pivot,
-        left_child(tags),
+        filter,
+        tags,
         Nx.select(
-          pos > pivot,
-          right_child(tags),
-          tags
+          pos < pivot,
+          left_child(tags),
+          Nx.select(
+            pos > pivot,
+            right_child(tags),
+            tags
+          )
         )
       )
-    )
+
+    base = (1 <<< level) - 1
+    mask = Nx.select(tags < (1 <<< (level + 1)) - 1, base, tags) - base
+    {tags, Nx.pow(2, mask)}
   end
 
   defnp bounded_subtree_size(i, levels, size) do
```

It limits us to 32 levels (4294967296 points) but we can increase it to u64 as well.

krstopro:
I think it would be better to sort the orders of the points in the tensor instead of points themselves. What I mean is the following. Suppose we have the tensor
```
[2, 5, 3, 1, 4]
```
For every point we compute its order in the tensor (i.e. how many points are smaller than it). In this case we get
```
[1, 4, 2, 0, 3]
```
This can efficiently be computed as
```
orders = Nx.broadcast(0, {size})
Nx.indexed_put(orders, tensor |> Nx.argsort() |> Nx.new_axis(1), Nx.iota({size}))
```
Then we sort `size * tags + orders`. This translates the amplitude constraint to the size constraint but it avoids the numerical problems. If we use `u64` for this tensor we can have the size up to 2 ** 32 which should be large enough. What do you think?

jonatanklosko:
> (i.e. how many points are smaller than it)

I think that's `Nx.argsort(tensor, direction: :desc)` :D

krstopro:
> orders = Nx.broadcast(0, {size})
> Nx.indexed_put(orders, tensor |> Nx.argsort() |> Nx.new_axis(1), Nx.iota({size}))

No? :)
```
tensor = Nx.tensor([2, 5, 3, 1, 4])
size = 5
orders = Nx.broadcast(0, {size})
Nx.indexed_put(orders, tensor |> Nx.argsort() |> Nx.new_axis(1), Nx.iota({size}))
```
gives
```
#Nx.Tensor<
  s64[5]
  [1, 4, 2, 0, 3]
>
```

jonatanklosko:
@krstopro what I meant is literally `orders = Nx.argsort(tensor, direction: :desc)`, no?

josevalim:
The issue is that we do argsort twice, which is the most expensive operation in the algorithm?

krstopro:
> I think that's `Nx.argsort(tensor, direction: :desc)` :D

ahahaha I need to take the coffee for today I guess. Thanks!

krstopro:
> The issue is that we do argsort twice, which is the most expensive operation in the algorithm?

Indeed. And we should use `u64` for this array.

jonatanklosko:
I didn't read the whole thread, but just a random additional observation that `Nx.argsort(x, direction: :desc)` is the same as `(size - 1) - Nx.argsort(x)`.

jonatanklosko:
Sorry, actually both of my observations are wrong, I had wrong idea and it just worked for this specific example. I guess it's me who needs coffee 😂

krstopro:
> I think it would be better to sort the orders of the points in the tensor instead of points themselves. What I mean is the following. Suppose we have the tensor
> 
> ```
> [2, 5, 3, 1, 4]
> ```
> 
> For every point we compute its order in the tensor (i.e. how many points are smaller than it). In this case we get
> 
> ```
> [1, 4, 2, 0, 3]
> ```
> 
> This can efficiently be computed as
> 
> ```
> orders = Nx.broadcast(0, {size})
> Nx.indexed_put(orders, tensor |> Nx.argsort() |> Nx.new_axis(1), Nx.iota({size}))
> ```
> 
> Then we argsort `size * tags + orders`. This translates the amplitude constraint to the size constraint but it avoids the numerical problems with floating points and gives the same result as the unbounded version. If we use `u64` for this tensor we can have the size up to 2 ** 32 which should be large enough. What do you think?

@josevalim I would still go with this version. It uses more time and memory, but it avoids problems with amplitude and floating points.

josevalim:
@krstopro, great insight!!!

I have just pushed something along these lines here: https://github.com/elixir-nx/scholar/commit/e8fee479c20fc407c5ffa2b03abf7c62a4003367

It is only 8% slower, I was able to avoid the double argsort, and it has roughly the memory constraints. :) I think we could even optimize it in EXLA if we really want, by allowing to argsort by multiple entries at once, but good enough for now.

krstopro:
@josevalim Did you test it with more than 1 dimension? Should sorting here
https://github.com/elixir-nx/scholar/blob/e8fee479c20fc407c5ffa2b03abf7c62a4003367/lib/scholar/neighbors/kd_tree.ex#L67C58-L67C58
go with `axis: 1`?

josevalim:
Btw, I pushed some other commits, so please check main, but we can likely apply the technique here now, which is very exciting!

krstopro:
> Btw, I pushed some other commits, so please check main, but we can likely apply the technique here now, which is very exciting!

Yes, I am trying it here. The main difference with random projection trees is that the projections are not known in advance (unlike tensor dimensions in k-d tree which we can sort immediately) which means I will probably have to do the sorting on them inside the loop.

I think `parent`, `left_child`, and `right_child` can be defined using `defn` now that the unbounded version is removed.

josevalim:
@krstopro pls let me know whenever this is ready for another round of review. We can probably unify this into a single `fit` function as well.

krstopro:
> @krstopro pls let me know whenever this is ready for another round of review. We can probably unify this into a single `fit` function as well.

@josevalim Yes, I am comparing the two functions at the moment. If they happen to be the same then the unbounded one can be removed.

krstopro:
@josevalim The updated code is here. Unbounded version is removed and predict function is added. There are few things I am not sure about.

krstopro:
One thing I noticed while comparing unbounded and bounded version of fit. After running
```
Nx.default_backend(EXLA.Backend)
key = Nx.Random.key(12)
{matrix, _key} = Nx.Random.uniform(key, type: :f32, shape: {10, 10})
{vector, _key} = Nx.Random.normal(key, type: :f32, shape: {10})
```
doing
```
Nx.dot(matrix, vector)
```
and
```
Nx.dot(Nx.new_axis(matrix, 0), [2], Nx.new_axis(vector, 0), [1])[[0, .., 0]]
```
won't give exactly the same result which I found slightly surprising. I don't know if this is a bug or just an XLA thing.

polvalente:
> won't give exactly the same result which I found slightly surprising. I don't know if this is a bug or just an XLA thing.

@krstopro This is just down to "XLA thing". Because if I run the same code in the GPU I do actually get the exact same results, even though the CPU diverges slightly. I believe it might be related to the way XLA is optimizing memory access down at the lowest levels.

f64 will also give the exact same results for both implementations, but I wouldn't worry too much about that

krstopro:
> @krstopro This is just down to "XLA thing". Because if I run the same code in the GPU I do actually get the exact same results, even though the CPU diverges slightly. I believe it might be related to the way XLA is optimizing memory access down at the lowest levels.

Yeah, I also thought it has to do with how XLA works at low level. Thanks for confirming.

test/scholar/neighbors/random_projection_forest_test.exs
```diff
@@ -0,0 +1,56 @@
...
+      tensor = example()
+      forest = RandomProjectionForest.fit(tensor, num_trees: 4, min_leaf_size: 3)
+      leaf_indices = RandomProjectionForest.predict(forest, Nx.tensor([[20, 30], [30, 50]]))
+      assert Nx.shape(leaf_indices) == {2, forest.num_trees, forest.leaf_size}
+    end
+
+    test "every point is its own leaf when leaf_size is 1" do
+      key = Nx.Random.key(12)
+      tensor = x()
+      forest = RandomProjectionForest.fit(tensor, num_trees: 1, min_leaf_size: 1)
```
msluszniak, line 51:
```suggestion
      forest = RandomProjectionForest.fit(tensor, num_trees: 1, min_leaf_size: 1, key: key)
```

lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -0,0 +1,363 @@
+defmodule Scholar.Neighbors.RandomProjectionForest do
+  @moduledoc """
+  Random Projection Forest.
```
josevalim, line 3:
```suggestion
  Random Projection Forest for approximate k-nearest neighbors.
```

lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -0,0 +1,363 @@
...
+
+  Each tree in a forest is constructed using a divide and conquer approach.
+  We start with the entire dataset and at every node we project the data onto a random
+  hyperplane and split it in the following way: the points with the projection smaller
+  than or equal to the median are put into the left subtree and the points with projection
+  greater than the median are put into the right subtree. We then proceed
+  recursively with the left and right subtree.
```
josevalim, line 10:
```suggestion
  recursively with the left and right subtree.

```

lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -0,0 +1,363 @@
...
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Examples
+
+      iex> key = Nx.Random.key(12)
+      iex> tensor = Nx.iota({5, 2})
+      iex> Scholar.Neighbors.RandomProjectionForest.fit(tensor, num_trees: 3, min_leaf_size: 2, key: key).indices
```
josevalim, line 62: Instead of returning only the indices, you can assign the forest to a variable and then assert on multiple entries?

```suggestion
      iex> forest = Scholar.Neighbors.RandomProjectionForest.fit(tensor, num_trees: 3, min_leaf_size: 2, key: key)
      iex> forest.indices
```

Things like depth, data, and so on would be nice to show.

lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -0,0 +1,363 @@
...
+
+    Nx.select(
+      level_mask,
+      level_medians,
+      medians
+    )
+  end
+
+  @doc """
+  Computes the leaf indices for every point in the input tensor.
```
josevalim, line 245:
```suggestion
  Computes the leaf indices for every point in the input tensor.

```

lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -0,0 +1,363 @@
...
+      #Nx.Tensor<
+        u32[1][3][3]
+        [
+          [
+            [0, 1, 2],
+            [0, 1, 2],
+            [4, 3, 2]
+          ]
+        ]
+      >
```
josevalim, line 264: How does this relate to a-knn? Or am I missing something and this is not yet the full aKNN implementation?

lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -0,0 +1,363 @@
...
+  """
+  deftransform predict(forest, x) do
```
josevalim, line 266:
```suggestion
  deftransform predict(%__MODULE__{} = forest, x) do
```

lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -0,0 +1,363 @@
...
+          left?,
+          cell_sizes,
+          {x, hyperplanes, medians, level + 1, nodes}
+        }
+      end
+
+    Nx.select(not left? and cell_sizes < leaf_size, start_indices - 1, start_indices)
+  end
+
+  defn left_child(nodes) do
```
josevalim, line 357: We can make them private here if you are not expecting people to navigate this tree (I think it is harder to navigate than the kd-tree ones, no?)

lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -0,0 +1,364 @@
...
+    Nx.select(
+      level_mask,
+      level_medians,
+      medians
+    )
+  end
+
+  @doc """
+  Computes the leaf indices for every point in the input tensor.
+  If the input tensor contains n points, then the result has shape {n, num_trees, leaf_size}.
```
josevalim, line 247:
```suggestion
  If the input tensor contains n points, then the result has shape `{n, num_trees, leaf_size}`.
  These are the indices of each tree in the forest that are closest to the input data.
```

lib/scholar/neighbors/random_projection_forest.ex
krstopro: I would separate that from Random Projection forest implementation. The reason is that we should have a separate functions that find nearest neighbors of query points and those that construct k-nn graph of the data.
Like this we can just do `forest = Forest.fit(x, ...)` and `Forest.predict(forest, x)` where `x` has been used both for fitting and prediction. Then we can process the result, which might include removing `i` from `i`-th row (we don't want `i` to be among k-nearest neighbors of itself in a k-NN graph).

josevalim: Actually, I dropped a comment above with my understanding of how this relates to kNN. Should we have a `predict` function that receives the distance function and then traverses the leaves? Or do we leave it as an exercise to the user? If the second, should we provide a code snippet?

josevalim: Oops, we commented at the same time. What you said makes sense. :) I suggested a comment above which I believe is pretty much what you said. :) I will resolve this one.

krstopro: I actually wrote graph.ex that contains functions for creating k-nn graph from data, both exact (using brute-force) and approximate using random projection forest and neighborhood expansion. Thought of submitting a separate PR for that.

lib/scholar/neighbors/random_projection_forest.ex
josevalim: Or alternatively:

```suggestion
  Random Projection Forest for (approximate) nearest neighbor searches.
```

Or similar. :D

josevalim (approved):
Beautiful work @krstopro! I have only added some minor comments and we can ship this soon.

Quick question: what is the next step from here? Which algorithm do you think we could implement on top of them?

Happy holidays!

lib/scholar/neighbors/random_projection_forest.ex
krstopro: Another possibility is to really return nearest neighbors in predict function of RandomProjectionForest, but when creating the graph, the user should get `k + 1` neighbors then filter out `i` from `i`-th row.

josevalim: Do you have a link to the paper so I can update our references in #207? :)

josevalim: > Another possibility is to really return nearest neighbors in predict function of RandomProjectionForest

We can totally decide that later and move the code around later.

krstopro:
@josevalim 

> Beautiful work @krstopro! I have only added some minor comments and we can ship this soon.

Thanks!

> Quick question: what is the next step from here? Which algorithm do you think we could implement on top of them?

I have submitted another pull request for that, see https://github.com/elixir-nx/scholar/pull/226.
It implements NN-expansion on top of random projection forest. This is similar to the algorithm used in [LargeVis](https://arxiv.org/pdf/1602.00370.pdf). Also, [PaCMAP uses Annoy](https://github.com/YingfanWang/PaCMAP/blob/8e22b7cb48ccbbbcb3feae90db79a2eaac19cf14/source/pacmap/pacmap.py#L14), which is based on random projection trees (though in Annoy they are constructed in a different way than here) so we might be able to proceed with the dimensionality reduction methods. Ideally, we would have NNDescent on top of random projection forest, [as done in PyNNDescent](https://youtu.be/xPadY4_kt3o?si=JILXIcSITWj7Z9Ay&t=853). I am still thinking of implementing it (I need to see if we can avoid looping over all points using `while`), but NN-expansion might work just fine as it is simple and quick.

> Happy holidays!

Thanks, same for you too!

lib/scholar/neighbors/random_projection_forest.ex
krstopro: Agreed.

msluszniak:
> @josevalim
> 
> > Beautiful work @krstopro! I have only added some minor comments and we can ship this soon.
> 
> Thanks!
> 
> > Quick question: what is the next step from here? Which algorithm do you think we could implement on top of them?
> 
> I have submitted another pull request for that, see #226. It implements NN-expansion on top of random projection forest. This is similar to the algorithm used in [LargeVis](https://arxiv.org/pdf/1602.00370.pdf). Also, [PaCMAP uses Annoy](https://github.com/YingfanWang/PaCMAP/blob/8e22b7cb48ccbbbcb3feae90db79a2eaac19cf14/source/pacmap/pacmap.py#L14), which is based on random projection trees (though in Annoy they are constructed in a different way than here) so we might be able to proceed with the dimensionality reduction methods. Ideally, we would have NNDescent on top of random projection forest, [as done in PyNNDescent](https://youtu.be/xPadY4_kt3o?si=JILXIcSITWj7Z9Ay&t=853). I am still thinking of implementing it (I need to see if we can avoid looping over all points using `while`), but NN-expansion might work just fine as it is simple and quick.
> 
> > Happy holidays!
> 
> Thanks, same for you too!

Yeah, I think that the next step should be NNDescent rather than Annoy.

josevalim:

PR #216, opened 2023-12-11
Add implementation of lazy_select via defn

msluszniak:
I implemented `defn` version of lazy_select. This implementation is better, so I'll close the PR in `Nx` in favour of this PR. I also pasted the results of benchmarks of `lazy_select` and `Nx.sort`. `lazy_select` was better only for small tensors. I'm thinking if we want to use it instead of sort for small tensors?


josevalim:
Given it gets slower for larger values, which is when it would matter the most, I am thinking we should give up exploring this route for now. :) Thank you for exploring all different options though!

msluszniak:
Sure :)

PR #224, opened 2023-12-19
Implement parent, left_child, and right_child as defn

krstopro:
Now that there is a single function for fitting trees and it is implemented within `defn` I think that `parent`, `left_child`, and `right_child` can go under `defn` as well. Perhaps they should be private as well?


krstopro:
I checked out from the wrong branch, so gotta close this for now.

josevalim:
The reason I keep them public (and non-defn) is because folks may need them to traverse the KDTree. Given they are small, I am fine with keeping them as is. :)

krstopro:
Makes sense!

PR #225, opened 2023-12-23
Introduce encoders in separate modules

msluszniak:


josevalim (approved):
We just need some basic module docs and we are good to go!

josevalim (approved):

PR #226, opened 2023-12-24
LargeVis

krstopro:
Adds methods for constructing k-nearest neighbor graph from data, both exact and approximate.
This is just a draft for now.


lib/scholar/metrics/neighbors.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Scholar.Metrics.Neighbors do
```
josevalim, line 1: Neighbors or graph?

josevalim: Or rather: what is the shape of the inputs expected here? Is it expecting a graph shape? If so, how is the graph laid out? Or can we can pass it the result of any Neighbors classification?

lib/scholar/neighbors/graph.ex
```diff
@@ -0,0 +1,182 @@
...
+    |> Nx.negate()
+    |> Nx.top_k(k: k)
+    |> elem(1)
+  end
+
+  @doc """
+  Constructs the approximate k-nearest neighbor graph using random projection forest
+  and neighborhood expansion.
+  """
+  deftransform approximate_random_projection_forest_and_expansion(x, opts) do
```
josevalim, line 66: So what is the algorithm implemented here? LargeVis? NNExpansion? I think we should aim to implement some particular algorithms and then, if there are shared paths, for example, "expansion" logic, we move that to the forests module?

lib/scholar/metrics/neighbors.ex
krstopro: hmmm I would set it to neighbors as it can work with the queries as well, not only graph. I guess I should rename the arguments to something else.

lib/scholar/neighbors/graph.ex
krstopro: The algorithm works as follows: first it initializes the k-NN graph using random projection forest, then refines it using NN-expansion. The algorithm is similar to the one used in LargeVis (although the idea is pretty natural and I had it even before stumbling upon that paper; something similar is used in other algorithms). The main difference is how the trees are constructed. LargeVis uses Annoy which, at every node, randomly picks two points, separates them with a hyperplane, and then splits the data depending on which side of the hyperplane each point is.

I would keep the expansion in this module or create a separate one for it as expansion itself can give decent results for k-NN graph construction (see e.g. EFANNA paper).

lib/scholar/metrics/neighbors.ex
krstopro: > Or rather: what is the shape of the inputs expected here? Is it expecting a graph shape? If so, how is the graph laid out? Or can we can pass it the result of any Neighbors classification?

The shape of both arguments is `{n, k}`. The function calculates how many nearest neighbors out of `k` possible are predicted correctly, then averages over all `n` points.
It doesn't work with classification, so maybe changing the module name to Graph or something else might a good idea.

krstopro: Or keep it as Neighbors as we have a separate module for metrics used for classification and regression.

josevalim: Yeah, agreed on Neighbors. We just need to improve the docs with more concrete examples!

krstopro: Yes, of course. I just submitted a PR draft to show why I didn't include k-NN search inside RandomProjectionForest module.

lib/scholar/neighbors/graph.ex
josevalim: > Ideally, we would have NNDescent instead of NN-expansion.

It is ok to have multiple implementations. The only trouble is figuring out how we are going to expose this. Calling it `Graph` is a bit generic. Do we expect all implementations of kNN-graph to be in this module or follow the struct returned by this module? I would say that this is too early to say and, generally speaking, I can easily see us having different knn-graphs in the future.

I would prefer if we could stick with the fit/predict APIs, but you are more knowledgeable than me on what is the fit/predict here. I assume this function is a `fit` and we need to add a `predict` to it? If we do such, we just call it `NNExpansion`?

krstopro: > > Ideally, we would have NNDescent instead of NN-expansion.
> 
> It is ok to have multiple implementations. The only trouble is figuring out how we are going to expose this. Calling it `Graph` is a bit generic. Do we expect all implementations of kNN-graph to be in this module or follow the struct returned by this module? I would say that this is too early to say and, generally speaking, I can easily see us having different knn-graphs in the future.
> 
> I would prefer if we could stick with the fit/predict APIs, but you are more knowledgeable than me on what is the fit/predict here. I assume this function is a `fit` and we need to add a `predict` to it? If we do such, we just call it `NNExpansion`?

Yes, agreed, `Graph` is too generic. Perhaps there should be a module for every k-NN graph construction algorithm. Also, calling this module `NNExpansion` seems like a good idea as long as we mention that we are using random projection forest for initialization. I think we need only `fit` function that takes a tensor and opts and returns a tensor of indices of shape `{n, k}` where `i`-th row contains `k`-nearest neighbors of the `i`-th point. Don't think `predict` is needed here.

josevalim: So to make sure I get it:

1. We have different ways to build graphs

2. NNExpansion is one of such ways and it *may* use RandomForests

3. RandomForests can be used to build graphs but NNExpansion is only one of the ways to do that with RandomForests

So here is the question: can I run NNExpansion with kdtrees? Because if we can, then I think we should make the tree/forest algorithm an argument of the NNExpansion algorithm. This way they can compose.

krstopro: Not sure if the current implementation of k-d trees can construct an approximate k-NN graph, but in principle yes, we could use them as well. Another possibility is to use random initialization, i.e. for every point randomly choose `k` other points as its `k`-nearest neighbors.

lib/scholar/neighbors/graph.ex
```diff
@@ -0,0 +1,182 @@
...
+      end
+
+    size = Nx.axis_size(x, 0)
+    num_trees = if opts[:num_trees], do: opts[:num_trees], else: 5 + round(:math.pow(size, 0.25))
+    num_iters = opts[:num_iters]
+    key = Keyword.get_lazy(opts, :key, fn -> Nx.Random.key(System.system_time()) end)
+    forest = Forest.fit(x, num_trees: num_trees, min_leaf_size: min_leaf_size, key: key)
+    IO.inspect(forest.leaf_size)
+    IO.inspect(num_trees)
+    leaves = Forest.predict(forest, x) |> Nx.reshape({size, num_trees * forest.leaf_size})
```
josevalim, line 93: Perhaps if you convert this to:

```suggestion
    leaves = Forest.predict(forest, x) |> Nx.reshape({size, :auto})
```

Then you can say: give me any data structure that implements `predict`? Then it is up to you if you pass a random forest or kd-tree or whatever?

josevalim: The code would be something like:

```
tensor
|> Forest.fit()
|> NNExpand.fit()
```

lib/scholar/metrics/neighbors.ex
```diff
@@ -0,0 +1,35 @@
...
+    end
+    recall_n(graph_true, graph_pred)
+  end
+
+  defn recall_n(graph_true, graph_pred) do
+    {n, k} = Nx.shape(graph_true)
+    concatenated = Nx.concatenate([graph_true, graph_pred], axis: 1) |> Nx.sort(axis: 1)
+    duplicate_mask = concatenated[[.., 0..(2 * k - 2)]] == concatenated[[.., 1..(2 * k - 1)]]
+    duplicate_mask |> Nx.sum() |> Nx.divide(n * k)
+  end
```
msluszniak, line 34: Does this function calculate something different than recall from `Scholar.Metrics`?

lib/scholar/neighbors/graph.ex
```diff
@@ -0,0 +1,182 @@
...
+  end
+
+  defn exact_brute_force_n(x, opts) do
+    k = opts[:k]
+    # TODO: Add other metrics other than Euclidean
+    x
+    |> Nx.new_axis(1)
+    |> Nx.subtract(Nx.new_axis(x, 0))
+    |> Nx.pow(2)
+    |> Nx.sum(axes: [2])
```
msluszniak, line 55: Here you have all other metrics implemented
```elixir
  defn pairwise_dist(x, opts) do
    broadcast_shape = {num_samples, num_samples, num_features}


    t1 =
      x
      |> Nx.reshape({1, num_samples, num_features})
      |> Nx.broadcast(broadcast_shape)


    t2 =
      x
      |> Nx.reshape({num_samples, 1, num_features})
      |> Nx.broadcast(broadcast_shape)


    case opts[:metric] do
      :squared_euclidean ->
        Distance.pairwise_squared_euclidean(x)


      :euclidean ->
        Distance.pairwise_euclidean(x)


      :manhattan ->
        Distance.manhattan(t1, t2, axes: [2])


      :cosine ->
        Distance.pairwise_cosine(x)


      :chebyshev ->
        Distance.chebyshev(t1, t2, axes: [2])
    end
  end
```

lib/scholar/neighbors/graph.ex
```diff
@@ -0,0 +1,182 @@
...
+        max(10, 2 * k + 1)
+      end
+
+    size = Nx.axis_size(x, 0)
+    num_trees = if opts[:num_trees], do: opts[:num_trees], else: 5 + round(:math.pow(size, 0.25))
+    num_iters = opts[:num_iters]
+    key = Keyword.get_lazy(opts, :key, fn -> Nx.Random.key(System.system_time()) end)
+    forest = Forest.fit(x, num_trees: num_trees, min_leaf_size: min_leaf_size, key: key)
+    IO.inspect(forest.leaf_size)
+    IO.inspect(num_trees)
```
msluszniak, line 92:
```suggestion
```

lib/scholar/neighbors/graph.ex
```diff
@@ -0,0 +1,182 @@
...
+    leaves = Forest.predict(forest, x) |> Nx.reshape({size, num_trees * forest.leaf_size})
+    graph = make_graph(leaves, x, k: k)
```
msluszniak, line 94: I guess, that this should be moved into defn

lib/scholar/neighbors/graph.ex
```diff
@@ -0,0 +1,182 @@
...
+      while {
+              graph,
+              {x, iter = 0}
+            },
+            iter < num_iters do
+        {expansion_iter(graph, x), {x, iter + 1}}
+      end
+
+    graph
+  end
```
msluszniak, line 114: This code might be vectorized

lib/scholar/metrics/neighbors.ex
krstopro: @msluszniak Merry Christmas!

I suppose you mean `Scholar.Metrics.Classification.recall/3`. Yes, this one is specific for algorithms that compute approximate k-nearest neighbors. It calculates the fraction of true nearest neighbors returned by the algorithm. In other words, it calculates how many nearest neighbors out of k possible did it get correctly, then averages over all the points. The meaning is similar to the one used for classification, but the way it is calculated is different (this one has to use sorting).

lib/scholar/neighbors/graph.ex
krstopro: Thanks! I've seen this somewhere implemented somewhere else, perhaps we can move it in a shared module?

lib/scholar/neighbors/graph.ex
```diff
@@ -0,0 +1,182 @@
...
+    sorted_indices = Nx.argsort(candidate_lists, axis: 1, stable: true)
+    inverse = inverse_permutation(sorted_indices)
+    sorted = Nx.take_along_axis(candidate_lists, sorted_indices, axis: 1)
+
+    duplicate_mask =
+      Nx.concatenate(
+        [
+          Nx.broadcast(0, {n, 1}),
+          Nx.equal(sorted[[.., 0..-2//1]], sorted[[.., 1..-1//1]])
+        ],
```
msluszniak, line 140:
```suggestion
          sorted[[.., 0..-2//1]] == sorted[[.., 1..-1//1]]
        ],
```

lib/scholar/metrics/neighbors.ex
msluszniak: Merry Christmas!
Ok, I see now :)

lib/scholar/neighbors/graph.ex
msluszniak: Yeah, this function is directly used in `t-SNE`. We may consider having this as a separate function.

lib/scholar/neighbors/graph.ex
krstopro: > The code would be something like:
> 
> ```
> tensor
> |> Forest.fit()
> |> NNExpansion.fit()
> ```

This could work, but then `Forest.fit/2` should return k-nearest neighbors or the graph and `NNExpansion.fit/2` should take a graph as input.

josevalim: Hrm, perhaps it doesn't make sense to think about how to abstract it unless we have another data structure we can use as the foundation for NNExpand. Maybe we keep this PR open until we have NNDescent or LargeVis or similar?

krstopro: > Hrm, perhaps it doesn't make sense to think about how to abstract it unless we have another data structure we can use as the foundation for NNExpand. Maybe we keep this PR open until we have NNDescent or LargeVis or similar?

Don't think having NNDescent or LargeVis is going to change it. LargeVis is basically random projection forest + NN-expansion - that is kinda what's done here.
What we could have instead is `Forest.fit_predict` (`fit_predict` is implemented in many scikit-learn modules). Now I am not sure what would I do with `NNExpansion`. It doesn't really go with fit/predict API.

josevalim: >  LargeVis is basically random projection forest + NN-expansion - that is kinda what's done here.

So to me that's perfect. We implement LargeVis and we embed this code inside. If someone wants to extract it out in the future for other use cases, we can do it once needed?

krstopro: > > LargeVis is basically random projection forest + NN-expansion - that is kinda what's done here.
> 
> So to me that's perfect. We implement LargeVis and we embed this code inside. If someone wants to extract it out in the future for other use cases, we can do it once needed?

Yeah, I think so. I don't really like the word LargeVis, especially since the algorithm seems natural to me (I had the same idea even before reading the paper), but then again the authors did it and not that I have a better name.

josevalim: At least is how people will recognize/search for it, so at least it helps on that aspect :)

krstopro: Agreed.

krstopro:
Alright, updated the code. Not sure if I should have opened a new pull requests as the comments here are outdated.

krstopro:
Closing this pull request and opening a new one as the comments here are outdated.

Issue #230, opened 2024-01-10
Bug in RandomProjectionForest due to Nx.dot with EXLA Backend


krstopro:
Currently, there is a bug in RandomProjectionForest that I am not exactly sure how to resolve. Namely, when querying the tree, the projection value might not be the same as when fitting the tree. This happens due to inner workings of XLA, as mentioned [here](https://github.com/elixir-nx/scholar/pull/215#issuecomment-1867836637). For example, if we have
```
Nx.global_default_backend(EXLA.Backend)
key = Nx.Random.key(12)
{tensor, key} = Nx.Random.uniform(key, shape: {10, 7})
forest = Scholar.Neighbors.RandomProjectionForest.fit(tensor, num_neighbors: 1, num_trees: 1, key: key)
```
then doing
```
Nx.dot(forest.hyperplanes[[.., 1]], [1], tensor, [1])[0][3]
```
gives
```
#Nx.Tensor<
  f32
  EXLA.Backend<host:0, 0.1762387501.1096941588.214884>
  1.3310179710388184
>
```
On the other hand, doing
```
query = Nx.new_axis(tensor[3], 0)
Nx.dot(forest.hyperplanes[[.., 1]], [1], query, [1])[0][0]
```
gives
```
#Nx.Tensor<
  f32
  EXLA.Backend<host:0, 0.1762387501.1096941588.214899>
  1.331018090248108
>
```
Also, `forest.medians[0][1]` gives
```
#Nx.Tensor<
  f32
  EXLA.Backend<host:0, 0.1762387501.1096941588.214903>
  1.3310179710388184
>
```
which is the same as the first result.
This is a problem when comparing projections to medians, as done in the module.
https://github.com/elixir-nx/scholar/blob/58ee5e32e96e5830c7d0da553786879ad4aa2a4f/lib/scholar/neighbors/random_projection_forest.ex#L323

One way to resolve this could be to use `:f64` for both hyperplanes and medians, albeit more memory expensive.

Any thoughts on this?

josevalim:
Keep in mind that Nx shows the wrong representation for f32. There is an open issue about that. I am not sure if it affects this, but it is something to keep in mind. Other than that, I am not sure how much we should sweat about this. Perhaps add a warning that we recommend f64 input tensor for better precision and we just make sure to preserve the input precision?

polvalente:
[...]

PR #231, opened 2024-01-15
Random Projection Forest Update

krstopro:
Changelog:
- Fixed the bug inside `predict/2` that assumed that query is of the same size as tensor used to grow the forest.
- Predict now returns the indices of (approximate) k-nearest neigbors as well as distances from the query points.
- ~~Hyperplanes and medians are now `f64`.~~
- Incorporated some feedback from #215 I didn't manage to before the pull request got closed. E.g. `left_child` and `right_child` are private now.
- Fixes https://github.com/elixir-nx/scholar/issues/230.

One thing I am still not sure about. `num_neighbors` is passed as an argument to `fit/2`. I prefer it like this, but it also makes sense to pass it as an argument to `predict/2`. There is also a small inconsistency with the option name used for the number of neighbors: [KNearestNeighbors uses num_neighbors](https://github.com/elixir-nx/scholar/blob/58ee5e32e96e5830c7d0da553786879ad4aa2a4f/lib/scholar/neighbors/k_nearest_neighbors.ex#L20), while [KDTree uses k](https://github.com/elixir-nx/scholar/blob/58ee5e32e96e5830c7d0da553786879ad4aa2a4f/lib/scholar/neighbors/kd_tree.ex#L28).

### DIFF
--- a/lib/scholar/neighbors/random_projection_forest.ex
+++ b/lib/scholar/neighbors/random_projection_forest.ex
@@ -253,9 +253,9 @@ defmodule Scholar.Neighbors.RandomProjectionForest do
 
     right_first = Nx.take_along_axis(level_proj, right_indices, axis: 1)
 
-    nodes = Nx.iota({num_nodes}, type: :u32)
     medians_first = (left_first + right_first) / 2
 
+    nodes = Nx.iota({num_nodes}, type: :u32)
     median_mask = width <= nodes and nodes < width + median_offset
     median_pos = Nx.argsort(median_mask, direction: :desc, stable: true, type: :u32)
     level_medians = Nx.take(medians_first, median_pos, axis: 1)
--- a/lib/scholar/neighbors/random_projection_forest.ex
+++ b/lib/scholar/neighbors/random_projection_forest.ex
@@ -48,7 +48,7 @@ defmodule Scholar.Neighbors.RandomProjectionForest do
     num_neighbors: [
       required: true,
       type: :pos_integer,
-      doc: "The number of nearest neighbors ..."
+      doc: "The number of nearest neighbors."
     ],
     min_leaf_size: [
       type: :pos_integer,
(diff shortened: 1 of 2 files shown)


lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -130,17 +170,16 @@ defmodule Scholar.Neighbors.RandomProjectionForest do
   defn fit_n(tensor, key, opts) do
     depth = opts[:depth]
     num_trees = opts[:num_trees]
-    type = to_float_type(tensor)
     {size, dim} = Nx.shape(tensor)
     num_nodes = 2 ** depth - 1
 
     {hyperplanes, _key} =
-      Nx.Random.normal(key, type: type, shape: {num_trees, depth, dim})
+      Nx.Random.normal(key, type: :f64, shape: {num_trees, depth, dim})
```
josevalim, line 177: I don't think we should default to f64, What if we allow the type as an argument? So we do: `type = opts[:type] || to_float_type(tensor)`?

josevalim (approved):
Beautiful, just one tiny comment about the type.

lib/scholar/neighbors/random_projection_forest.ex
krstopro: It seems that initializing hyperplanes and medians to `f64` might not be necessary. I changed the code in `predict/2` from 
```
proj = Nx.dot(query, hyperplanes[level])
```
(hyperplanes were vectorized by first coordinates which corresponds to trees) to
```
proj = Nx.dot(hyperplanes[[.., level]], [1], query, [1]) |> Nx.vectorize(:trees)
```
which is kinda the same as done in `fit/2`. Now the projections are the same inside `fit/2` and `predict/2`, even with `f32`; it seems the order of the arguments was important as well. Should have done this before, sorry about that!
It will work for constructing the k-NN graph, i.e. when `query = tensor`, where `tensor` is used to grow the forest. Might not work for arbitrary query (i.e. there might be some issues with projections as described in https://github.com/elixir-nx/scholar/issues/230). Perhaps adding a note in the docs might be a good idea?

josevalim:
More docs are always welcome too! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

msluszniak:
I didn't make it on time but great job!

PR #232, opened 2024-01-17
LargeVis

krstopro:
LargeVis algorithm for k-nearest neighbor graph construction.


lib/scholar/neighbors/large_vis.ex
```diff
@@ -0,0 +1,146 @@
...
+              graph,
+              _distances = Nx.broadcast(Nx.tensor(:nan, type: to_float_type(tensor)), {n, k}),
+              {tensor, iter = 0}
+            },
+            iter < num_iters do
+        {graph, distances} = expansion_iter(graph, tensor)
+        {graph, distances, {tensor, iter + 1}}
+      end
+
+    {graph, distances}
```
msluszniak, line 136:
```suggestion
    {result, _} =
      while {
              {graph,
              _distances = Nx.broadcast(Nx.tensor(:nan, type: to_float_type(tensor))}, {n, k}),
              {tensor, iter = 0}
            },
            iter < num_iters do
        {expansion_iter(graph, tensor), {tensor, iter + 1}}
      end

    result
```

lib/scholar/metrics/neighbors.ex
```diff
@@ -0,0 +1,38 @@
+defmodule Scholar.Metrics.Neighbors do
+  import Nx.Defn
+
+  deftransform recall(graph_true, graph_pred) do
```
josevalim, line 4: I believe these conditionals would work inside `defn`, no?

lib/scholar/metrics/neighbors.ex
```diff
@@ -0,0 +1,38 @@
+defmodule Scholar.Metrics.Neighbors do
```
josevalim, line 1: We need `@moduledoc` and `@doc` for the function. :)

lib/scholar/neighbors/utils.ex
```diff
@@ -0,0 +1,64 @@
+defmodule Scholar.Neighbors.Utils do
```
josevalim, line 1:
```suggestion
defmodule Scholar.Neighbors.Utils do
  @moduledoc false
```

lib/scholar/neighbors/utils.ex
```diff
@@ -0,0 +1,64 @@
...
+  import Nx.Defn
+  require Nx
+
+  defn find_neighbors(query, data, candidate_indices, opts) do
```
josevalim, line 5: Quick question: does it work on any query/data/candidate_indices or it has to be related to random forests? We probably don't want to make this public for now (hence `@moduledoc false`), but we should most likely add a `@doc` entry to this one as well, explaining the shape and the role of each argument.

lib/scholar/neighbors/large_vis.ex
```diff
@@ -0,0 +1,147 @@
...
+      If the key is not provided, it is set to `Nx.Random.key(System.system_time())`.
+      """
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(opts)
+
+  @doc """
+  Constructs the approximate k-NN graph with LargeVis.
+  Returns neighbor indices and distances.
```
josevalim, line 58:
```suggestion

  Returns neighbor indices and distances.
```

josevalim:
Looks great, I have dropped some comments, but I think they are minor.

Btw, it would be really cool if we had a Livebook comparing our different solutions here, similar to this image:

image

[Source](https://www.researchgate.net/figure/Application-of-t-SNE-LargeVis-UMAP-TriMAP-and-PaCMAP-to-the-Mammoth-dataset-In_fig1_346766201).

lib/scholar/neighbors/utils.ex
krstopro: Yeah, it is kinda general. For example, brute-force k-NN search can be implemented as `find_neighbors(query, tensor, tensor, num_neighbors: k)`. Two things that should be added:
- More distances other than Euclidean.
- A flag whether there is a need to remove duplicates from candidate lists or not. For example, brute-force search as specified above doesn't need duplicate removal. Similarly for random projection forest when there is a single tree (there can be no duplicates in a leaf). Or should there be two functions: one that removes the duplicates and the other that does not?

I called the function `find_neighbors` simply because I cannot think of a better name. Something like `find_nearest_neighbors` or `nn_search` might be better.

krstopro:
> Looks great, I have dropped some comments, but I think they are minor.
> 
> Btw, it would be really cool if we had a Livebook comparing our different solutions here, similar to this image:

That would be the goal. It remains to be seen if the current implementation of k-NN graph construction will work for dimensionality reduction methods. One issue might be the memory cost of `predict/2` in `RandomProjectionForest`. Currently, it uses $O(T \times N \times L)$ memory, for $T$ trees, sample size $N$ and leaf size $L$. This can be reduced to $O(N \times L)$ by looping over trees, but then I am afraid of #230 coming back to haunt me. :)

lib/scholar/metrics/neighbors.ex
krstopro: This one won't work
```
if Nx.shape(neighbors_true) != Nx.shape(neighbors_pred) do
```
as shape tuples cannot be converted to tensors. So either I leave it like this or compare each axis separately.

josevalim:
> This can be reduced to by looping over trees, but then I am afraid of https://github.com/elixir-nx/scholar/issues/230 coming back to haunt me. :)

Probably worth giving it a shot anyway.

josevalim:
> This can be reduced to by looping over trees, but then I am afraid of https://github.com/elixir-nx/scholar/issues/230 coming back to haunt me. :)

Probably worth giving it a shot anyway, especially as we allow precision to be given as parameter.

josevalim:
After all the feedback, this looks good to go. Just docs missing and I will merge it!

josevalim:

PR #233, opened 2024-01-24
Nn descent

msluszniak:


lib/scholar/neighbors/nn_descent.ex
```diff
@@ -0,0 +1,817 @@
...
+      flags =
+        if flag != Nx.s8(-1),
+          do: Nx.indexed_put(flags, Nx.concatenate([index0_ext, curr_ext]), flag),
+          else: flags
+
+      {indices, keys, flags}
+    end
+  end
+
+  defn add_neighbor_vectorized(curr_graph, index1, key, flag, num_nodes) do
```
jonatanklosko, line 729: Technically we could use `add_neighbor` by having vectoried `index0` with all indices. I guess it may be slightly more work, since we would use two indices to slice instead of one, etc, but perhaps it's negligible and worth simplifying the implementation.

josevalim:
Beautifully done! /cc @krstopro in case they have feedback or find places where we could share logic with other algorithms.

lib/scholar/neighbors/nn_descent.ex
```diff
@@ -0,0 +1,838 @@
...
+
+    {size_new, new_candidates_num_nodes} = Nx.shape(new_candidates_indices)
+    old_candidates_num_nodes = Nx.axis_size(old_candidates_indices, 1)
+    num_samples = Nx.axis_size(data, 0)
+    update_index = Nx.s64(0)
+
+    # Normally there would be a stack of that will dynamically grow
+    # so we need to preallocate it with a fixed size
+    expand_factor = 50
+    updates_indices = Nx.broadcast(Nx.s64(0), {expand_factor * num_samples, 2})
```
msluszniak, line 452: I'm thinking about how to tackle this one.

msluszniak:
I want to also implement `query` analogical to what we have in `KDTree`. The current implementation has many nested whiles, which might have an impact on performance. I'll think more about how to get rid of them, but for now, I have no good idea.

krstopro:
This looks great, but I have yet to go through bigger part of it. Few comments for now:
- `sum_samples` -> `num_samples` (guess this is a typo)
- I have some questions regarding this part: https://github.com/msluszniak/scholar/blob/e28d50ac81e96e82c4c7d23f24690ba67e0ee01f/lib/scholar/neighbors/nn_descent.ex#L134-L154
If I understood correctly, graph is first initialized to a matrix filled with `-1`. Then updated using random projection forest. Then updated with random neighbors. Is the first step necessary? Can graph be initialized using `predict/2` of random projection forest? Similarly, is `update_by_leaves/3` needed? I am not sure if it is working correctly as it is now.
- `Nx.Constants.max_finite(to_float_type(tensor))` could be replaced with `Nx.Constants.infinity(to_float_type(tensor))`.
- Why are keys `s64` at https://github.com/msluszniak/scholar/blob/e28d50ac81e96e82c4c7d23f24690ba67e0ee01f/lib/scholar/neighbors/nn_descent.ex#L598 and https://github.com/msluszniak/scholar/blob/e28d50ac81e96e82c4c7d23f24690ba67e0ee01f/lib/scholar/neighbors/nn_descent.ex#L603?

msluszniak:
* Yes, this is a typo
* It is initialized because using a random projection forest is not a necessary step and the number of leaves might be different than the number of neighbors. So I guess that all functions are needed. Neighbours might be initialized by the results of the RPF `predict/2` but this is using one ANN inside another ANN algorithm. The point is to  initialize graph of NNDescent using leaves from RPF https://github.com/lmcinnes/pynndescent/blob/0181c0cb836eb160eabf75386b553c31f527e3ef/pynndescent/pynndescent_.py#L346. 
* Yes, you're right, I'll change that
* There is no need to use floats here since all values in keys are sampled so we faster s64 that f64 ;)
*

krstopro:
> Neighbours might be initialized by the results of the RPF predict/2 but this is using one ANN inside another ANN algorithm.

Yes, that is exactly what I thought: initialize the k-NN graph using `RandomProjectionForest.predict/2`, then refine it using NN-descent. Something like that is already being done in LargeVis
https://github.com/elixir-nx/scholar/blob/cd64e154687f2462b7b228b430d1a0fb8c279ed6/lib/scholar/neighbors/large_vis.ex#L119-L120
except that NN-expansion is used instead of NN-descent.

I don't understand fully how the trees in PyNNDescent work; they are not implemented in the same way as in `RandomProjectionForest`. One thing that bothers me is that leaves are not computed using data instead of routing the points down the trees. After that leaves are used to update the graph, but the tree structure is ignored here.

Another issue might be in `get_leaves_from_forest/1`. It seems to me that it is assumed that the leaves are arranged as blocks of indices of size `leaf_size`, with the possible exception of the last leaf which might be smaller. This is not true in the general case. Consider the following example.
Let sample size be 10 and minimum leaf size be 2. When fitting the tree we first split the data into two nodes of size 5. Then both of the nodes will be split into another two nodes with sizes 3 and 2, respectively. Since we reached the minimum leaf size we stop splitting and set the leaf size to 3 (the larger of 3 and 2). We end up with 4 leaves with sizes 3, 2, 3, 2. When querying the leaves of size 2 (these are right subtrees of of trees of size 5) we will also query the last point of the leaf of size 3 left of it. In other words, there can be between a single point of overlap between two leaves. I might be wrong, but I don't think that is assumed currently.

lib/scholar/neighbors/nn_descent.ex
```diff
@@ -0,0 +1,838 @@
...
+  @moduledoc """
+  Nearest Neighbors Descent (NND) is an algorithm that calculates Approximated Nearest Neighbors (ANN)
+  for a given set of points[1].
+
+  It is implemented using Random Projection Forest[2].
+
+  ## References
+
+    [1] [Efficient K-Nearest Neighbor Graph Construction for Generic Similarity Measures](https://www.cs.princeton.edu/cass/papers/www11.pdf).
+    [2] Random projection trees and low dimensional manifolds
```
krstopro, line 11: Maybe replace this one with "Randomized partition trees for nearest neighbor search" (https://cseweb.ucsd.edu/~dasgupta/papers/exactnn-algo.pdf). Both papers are dealing with random projection trees, but the latter is focusing on k-NN.

lib/scholar/neighbors/large_vis.ex
```diff
@@ -104,7 +104,13 @@ defmodule Scholar.Neighbors.LargeVis do
     num_trees = opts[:num_trees] || 5 + round(:math.pow(size, 0.25))
     key = Keyword.get_lazy(opts, :key, fn -> Nx.Random.key(System.system_time()) end)
 
-    fit_n(tensor, num_neighbors: k, min_leaf_size: min_leaf_size, num_trees: num_trees, key: key)
+    fit_n(tensor,
+      num_neighbors: k,
+      min_leaf_size: min_leaf_size,
+      num_trees: num_trees,
+      key: key,
```
krstopro, line 111: I would set the key as the last argument, but this is just a personal preference.

krstopro (commented):
Minor comments.

msluszniak:
> > Neighbours might be initialized by the results of the RPF predict/2 but this is using one ANN inside another ANN algorithm.
> 
> Yes, that is exactly what I thought: initialize the k-NN graph using `RandomProjectionForest.predict/2`, then refine it using NN-descent. Something like that is already being done in LargeVis
> 
> https://github.com/elixir-nx/scholar/blob/cd64e154687f2462b7b228b430d1a0fb8c279ed6/lib/scholar/neighbors/large_vis.ex#L119-L120
> 
> except that NN-expansion is used instead of NN-descent.
> 
> I don't understand fully how the trees in PyNNDescent work; they are not implemented in the same way as in `RandomProjectionForest`. One thing that bothers me is that leaves are not computed using data instead of routing the points down the trees. After that leaves are used to update the graph, but the tree structure is ignored here.
> 
> Another issue might be in `get_leaves_from_forest/1`. It seems to me that it is assumed that the leaves are arranged as blocks of indices of size `leaf_size`, with the possible exception of the last leaf which might be smaller. This is not true in the general case. Consider the following example. Let sample size be 10 and minimum leaf size be 2. When fitting the tree we first split the data into two nodes of size 5. Then both of the nodes will be split into another two nodes with sizes 3 and 2, respectively. Since we reached the minimum leaf size we stop splitting and set the leaf size to 3 (the larger of 3 and 2). We end up with 4 leaves with sizes 3, 2, 3, 2. When querying the leaves of size 2 (these are right subtrees of of trees of size 5) we will also query the last point of the leaf of size 3 left of it. In other words, there can be between a single point of overlap between two leaves. I might be wrong, but I don't think that is assumed currently.

Yeah, you're right. They have an implementation that enables them to compute leaves as you describe. https://github.com/lmcinnes/pynndescent/blob/0181c0cb836eb160eabf75386b553c31f527e3ef/pynndescent/rp_trees.py#L1101. I'll change that to initialize the graph with the predicted results from `RPF`.

lib/scholar/neighbors/large_vis.ex
```diff
@@ -104,7 +104,13 @@ defmodule Scholar.Neighbors.LargeVis do
...
+      num_iters: opts[:num_iters]
+    )
   end
 
   defnp fit_n(tensor, opts) do
```
krstopro, line 116: `key` should be the second argument here, not an option. Whoever wrote `LargeVis` should have done it in the first place. :)

krstopro (commented):
Pass `key` as argument to `fit_n`.

msluszniak:
@krstopro I updated the code with Random Projection Forest as you suggested. I'm also wondering how to implement something similar to stack/queue.  There are some places in the code where I need to aggregate updates and the number of updates depends on tensor content. Currently, I allocate more space than needed by multiplying it by a fixed constant. However, for large data, this approach will be very ineffective. I'm thinking that computing prediction using RFP might be time-consuming in comparison to what is done in Python implementation. But, I think that it will be hard to go around so it's not the biggest problem.

krstopro:
@msluszniak Sorry for not responding before, I was rather busy. Looks good to me now. One thing I forgot to mention: the code of Random Projection Forest can be refactored to return the points of the leaf corresponding to the query point (leaf size should be at least `num_neighbors`). I think the following several lines are accomplishing this:
https://github.com/elixir-nx/scholar/blob/cd64e154687f2462b7b228b430d1a0fb8c279ed6/lib/scholar/neighbors/random_projection_forest.ex#L330-L344

> I'm thinking that computing prediction using RFP might be time-consuming in comparison to what is done in Python implementation.

That's a fair point. Do you know what are the major differences?

josevalim:
Btw, a new Nx version is out. Once this is merged we can ship a new release with all of these features you have been cooking :)

msluszniak:
> @msluszniak Sorry for not responding before, I was rather busy. Looks good to me now. One thing I forgot to mention: the code of Random Projection Forest can be refactored to return the points of the leaf corresponding to the query point (leaf size should be at least `num_neighbors`). I think the following several lines are accomplishing this:
> 
> https://github.com/elixir-nx/scholar/blob/cd64e154687f2462b7b228b430d1a0fb8c279ed6/lib/scholar/neighbors/random_projection_forest.ex#L330-L344
> 
> > I'm thinking that computing prediction using RFP might be time-consuming in comparison to what is done in Python implementation.
> 
> That's a fair point. Do you know what are the major differences?

They use leaf points as initial values, so I guess it's exactly what you pointed right here ;))

josevalim (approved):
Beautiful job. I will go ahead and merge it. :) It seems there is still one separate performance issue, so please open up a separate issue with details for further discussion.

josevalim:

PR #235, opened 2024-02-27
Add Mean Pinball Loss function.

JoaquinIglesiasTurina:
The Mean Pinball Loss function for evaluation of prediction performance of regression models.
Listed in the functions to add[ in this issue](https://github.com/elixir-nx/scholar/issues/122).

I've adapted the implementation from `sklearn` and the relevant of that library are linked in the code, to give credit and serve as reference.

### DIFF
--- a/lib/scholar/metrics/regression.ex
+++ b/lib/scholar/metrics/regression.ex
@@ -534,7 +534,7 @@ defmodule Scholar.Metrics.Regression do
         {:or,
          [
            {:custom, Scholar.Options, :weights, []},
-           {:custom, Scholar.Options, :multi_weights, []}           
+           {:custom, Scholar.Options, :multi_weights, []}
          ]},
       doc: """
       The weights for each observation. If not provided,
@@ -542,12 +542,12 @@ defmodule Scholar.Metrics.Regression do
       """
     ],
     multioutput: [
-      type: {:or,
-             [
-               {:custom, Scholar.Options, :weights, []},
-               {:in, [:raw_values, :uniform_average]}
-             ]
-      },
+      type:
+        {:or,
+         [
+           {:custom, Scholar.Options, :weights, []},
+           {:in, [:raw_values, :uniform_average]}
+         ]},
       default: :uniform_average,
       doc: """
       Defines aggregating of multiple output values.
@@ -559,14 +559,15 @@ defmodule Scholar.Metrics.Regression do
 
         `:uniform_average` :
             Errors of all outputs are averaged with uniform weight.
-      
+
       The weights for each observation. If not provided,
       all observations are assigned equal weight.
-      """      
+      """
     ]
   ]
+
   @mean_pinball_loss_schema NimbleOptions.new!(mean_pinball_loss_opts)
-  
+
   @doc ~S"""
   Calculates the mean pinball loss to evaluate predictive performance of quantile regression models.
 
@@ -598,6 +599,7 @@ defmodule Scholar.Metrics.Regression do
   deftransform mean_pinball_loss(y_true, y_pred, opts \\ []) do
     mean_pinball_loss_n(y_true, y_pred, NimbleOptions.validate!(opts, @mean_pinball_loss_schema))
   end
+
   defnp mean_pinball_loss_n(y_true, y_pred, opts) do
     assert_same_shape!(y_true, y_pred)
     alpha = opts[:alpha]
@@ -612,20 +614,25 @@ defmodule Scholar.Metrics.Regression do
     # mimics the sklearn behavior
     case opts[:multioutput] do
       # raw_values returns plain output errors. One value per channel.
-      :raw_values -> output_errors
+      :raw_values ->
+        output_errors
+
       # uniform_average returns the mean of the above. Note how they are averaged.
       :uniform_average ->
         output_errors
         |> Nx.mean()
+
       # pass `:multioutput` as sample weights to average the error of each output
       multi_output_weights ->
-        handle_sample_weights(output_errors, [sample_weights: multi_output_weights])
+        handle_sample_weights(output_errors, sample_weights: multi_output_weights)
     end
   end
 
   defnp handle_sample_weights(loss, opts, mean_opts \\ []) do
     case opts[:sample_weights] do
-      nil -> Nx.mean(loss, mean_opts)
+      nil ->
+        Nx.mean(loss, mean_opts)
+
       weights ->
         Nx.weighted_mean(loss, weights, mean_opts)
     end
--- a/lib/scholar/metrics/regression.ex
+++ b/lib/scholar/metrics/regression.ex
@@ -559,10 +559,13 @@ defmodule Scholar.Metrics.Regression do
     output_errors = handle_sample_weights(loss, opts, axes: [0])
     # mimics the sklearn behavior
     case opts[:multioutput] do
+      # raw_values returns plain output errors. One value per channel.
       :raw_values -> output_errors
+      # uniform_average returns the mean of the above. Note how they are averaged.
       :uniform_average ->
         output_errors
         |> Nx.mean()
+      # every other case, just take the mean without any axes nor weight.
       _ -> handle_sample_weights(loss, opts)
     end
   end
--- a/lib/scholar/metrics/regression.ex
+++ b/lib/scholar/metrics/regression.ex
@@ -548,7 +548,15 @@ defmodule Scholar.Metrics.Regression do
     diff = y_true - y_pred
     sign = diff >= 0
     loss = alpha * sign * diff - (1 - alpha) * (1 - sign) * diff
-    Nx.mean(loss)
+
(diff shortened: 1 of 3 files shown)


lib/scholar/metrics/regression.ex
```diff
@@ -519,6 +519,43 @@ defmodule Scholar.Metrics.Regression do
...
+    diff = Nx.subtract(y_true, y_pred)
+    sign = Nx.greater_equal(diff, 0)
+    subtracted_sign = Nx.subtract(1, sign)
+    
+    Nx.subtract(
+      Nx.multiply(alpha, sign)
+      |> Nx.multiply(diff),
+      Nx.multiply(1 - alpha, subtracted_sign)
+      |> Nx.multiply(diff))
+    |> Nx.mean()
```
msluszniak, line 556: You can simplify it to the following:
```suggestion
    diff = y_true - y_pred
    sign = diff >= 0
    loss = alpha * sign * diff - (1 - alpha) * (1 - sign) * diff
    Nx.mean(loss)
```

lib/scholar/metrics/regression.ex
```diff
@@ -519,6 +519,43 @@ defmodule Scholar.Metrics.Regression do
...
+      iex> y_pred = Nx.tensor([2, 3, 4])
+      iex> Scholar.Metrics.Regression.mean_pinball_loss(y_true, y_pred, 0.5)
+      #Nx.Tensor<
+        f32
+        0.5
+      >
+  """
+  defn mean_pinball_loss(y_true, y_pred, alpha \\ 0.5) do
+    check_shape(y_true, y_pred)
+
```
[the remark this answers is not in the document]
msluszniak: Optional arguments should be passed using `opts`

msluszniak:
Thank you for your contribution. I think that will be nice if we also support `sample_weights` and `multioutput`. I can help you with the implementation of these or can implement it by myself. Let me know what option do you prefer. :))

JoaquinIglesiasTurina:
I did not work with ` multioutput` nor `sample_weights` since I did not see them. Implemented in other loss functions.
I would be happy to add them.

lib/scholar/metrics/regression.ex
```diff
@@ -519,6 +519,38 @@ defmodule Scholar.Metrics.Regression do
...
+    check_shape(y_true, y_pred)
+    alpha = opts[:alpha]
+
+    # Formula adapted from sklearn:
+    # https://github.com/scikit-learn/scikit-learn/blob/128e40ed593c57e8b9e57a4109928d58fa8bf359/sklearn/metrics/_regression.py#L299
+    diff = y_true - y_pred
+    sign = diff >= 0
+    loss = alpha * sign * diff - (1 - alpha) * (1 - sign) * diff
+    Nx.mean(loss)
+  end
```
billylanchantin, line 552: Hey this is a total drive by comment, but I actually had to write a mean pinball error a few months ago:

* https://hexdocs.pm/exgboost/quantile_prediction_interval.html#discussion-pinball-error

I mention it because I also looked up sklearn's implementation and found it a bit confusing. I was able to simplify it to this:

```elixir
defmodule Pinball do
  import Nx.Defn

  defn mean_error(y_pred, y_true, alpha) do
    dy = y_true - y_pred
    # if dy < 0, do: alpha, else: alpha - 1
    coeff = alpha + (dy >= 0) - 1
    error = coeff * dy
    Nx.mean(error)
  end
end
```

Feel free to use or disregard :)

Also, I'll be able to update that livebook to use Scholar's version after this PR, so thanks!

lib/scholar/metrics/regression.ex
```diff
@@ -519,6 +519,65 @@ defmodule Scholar.Metrics.Regression do
...
+    # mimics the sklearn behavior
+    case opts[:multioutput] do
+      # raw_values returns plain output errors. One value per channel.
+      :raw_values -> output_errors
+      # uniform_average returns the mean of the above. Note how they are averaged.
+      :uniform_average ->
+        output_errors
+        |> Nx.mean()
+      # every other case, just take the mean without any axes nor weight.
+      _ -> handle_sample_weights(loss, opts)
```
msluszniak, line 569: I think that for averaging errors we should have a separate tensor with weights, look: https://github.com/scikit-learn/scikit-learn/blob/128e40ed593c57e8b9e57a4109928d58fa8bf359/sklearn/metrics/_regression.py#L302-L309

lib/scholar/metrics/regression.ex
```diff
@@ -519,6 +519,65 @@ defmodule Scholar.Metrics.Regression do
...
+      iex> y_true = Nx.tensor([[1, 0, 0, 1], [0, 1, 1, 1], [1, 1, 0, 1]])
+      iex> y_pred = Nx.tensor([[0, 0, 0, 1], [1, 0, 1, 1], [0, 0, 0, 1]])
+      iex> Scholar.Metrics.Regression.mean_pinball_loss(y_true, y_pred, alpha: 0.5, multioutput: :raw_values)
+      #Nx.Tensor<
+        f32[4]
+        [0.5, 0.3333333432674408, 0.0, 0.0]
+      >
+  """
+  defn mean_pinball_loss(y_true, y_pred, opts \\ [alpha: 0.5]) do
+    assert_same_shape!(y_true, y_pred)
```
msluszniak, line 550: It would be great if you also support NimbleOptions. Examples of how they are used are in almost all modules.

msluszniak:
I added two small suggestions. When they will be applied, we will merge this PR :))

JoaquinIglesiasTurina:
While adding the noted suggestions, I have introduced two pieces of code that I believe to be technical debt:
- I was getting some test failures due floating point precision errors. I've added[ this private function](https://github.com/JoaquinIglesiasTurina/scholar/blob/4588806793c26fb32ac029e4c38f4e7e7c928fa8/test/scholar/metrics/regression_test.exs#L132) to bypass those. I believe this function should be provided as a testing utility, either by Nx or Scholar. But I don't think I have sufficient knowledge of the project to make that decision.
- I needed multi dimensional weights options for `NimbleOptions`, so I [added that](https://github.com/JoaquinIglesiasTurina/scholar/blob/4588806793c26fb32ac029e4c38f4e7e7c928fa8/lib/scholar/options.ex#L86). It's copy-pasted and lightly modified `weights` option and it feels ugly. I am unsure of the implications of just increasing the allowed rank on the `weight` option, and `weight` takes a list, which I do not think would cut it for my use case. So this felt like the safer option. But maybe someone else can think of a better approach.

Other than that, I've renamed the `sample_weight:` option to `sample_weights:`. I've noticed that scikit-learn has `sample_weight` but Scholar has `sample_weights:` on `linear/regression.ex`. So I've made that change to be consistent.

lib/scholar/metrics/regression.ex
```diff
@@ -519,6 +519,118 @@ defmodule Scholar.Metrics.Regression do
...
+      iex> Scholar.Metrics.Regression.mean_pinball_loss(y_true, y_pred, alpha: 0.5, multioutput: :raw_values)
+      #Nx.Tensor<
+        f32[4]
+        [0.5, 0.3333333432674408, 0.0, 0.0]
+      >
+  """
+  deftransform mean_pinball_loss(y_true, y_pred, opts \\ []) do
+    mean_pinball_loss_n(y_true, y_pred, NimbleOptions.validate!(opts, @mean_pinball_loss_schema))
+  end
+  defnp mean_pinball_loss_n(y_true, y_pred, opts \\ []) do
```
msluszniak, line 601: In defnp, opts are always passed (otherwise there should not be opts ;) ) 
```suggestion
  defnp mean_pinball_loss_n(y_true, y_pred, opts) do
```

test/scholar/metrics/regression_test.exs
```diff
@@ -66,4 +66,70 @@ defmodule Scholar.Metrics.RegressionTest do
...
+      mpbl = Regression.mean_pinball_loss(y_true, y_pred,
+        alpha: 0.5, multioutput: Nx.tensor([1, 2, 3, 4]))
+      assert almost_equal(mpbl, Nx.tensor(0.1166666))
+      mpbl = Regression.mean_pinball_loss(y_true, y_pred,
+        alpha: 0.5, multioutput: nil)
+      assert almost_equal(mpbl, expected_error)
+    end
+  end
+
+  defp almost_equal(tensor1, tensor2, tolerance \\ 0.0001) do
```
msluszniak, line 132: There is such functionality in Nx and it is used here https://github.com/elixir-nx/scholar/blob/cd64e154687f2462b7b228b430d1a0fb8c279ed6/test/scholar/linear/linear_regression_test.exs#L75. There atol is absolute tolerance and rtol is relative tolerance.

JoaquinIglesiasTurina: Thank you for the pointer. I should have better read the docs!

msluszniak (approved):
Really good work, well done! :))

josevalim:

PR #236, opened 2024-03-04
Trimap

msluszniak:


lib/scholar/manifold/trimap.ex
```diff
@@ -0,0 +1,478 @@
...
+      :manhattan ->
+        Distance.manhattan(x, y, axes: [-1])
+
+      :cosine ->
+        Distance.cosine(x, y, axes: [-1])
+
+      :chebyshev ->
+        Distance.chebyshev(x, y, axes: [-1])
+    end
+  end
```
josevalim, line 137: If you prefer:

```suggestion
  deftransformp handle_dist(x, y, opts) do
    apply(Distance, opts[:metric], [x, y, [axes: [-1]])
  end
```

lib/scholar/manifold/trimap.ex
```diff
@@ -0,0 +1,478 @@
+defmodule Scholar.Manifold.Trimap do
+  @moduledoc """
+  TriMap: Large-scale Dimensionality Reduction Using Triplets.
```
josevalim, line 3: It you can add a bit more words, especially about how it compares with other techniques in Scholar (such as kNN), both in terms of performance and results, it would be excellent. :)

josevalim (approved):
Looks great. Just some minor comments.

I believe you can now merge/rebase this on main (or maybe it will be easier to start a new branch and cherry pick your trimap on top).

After merged, let's please writing a document comparing our manifolds, in performance and results. :)

PR #245, opened 2024-03-25
Update NNDescent

### LINKED ISSUE #241: Improve storing updates in NNDescent algorithm
In the original implementation of the NNDescent algorithm updates for the algorithm are stored in a stack. However, at this moment we cannot simulate this data structure easily so we need to think about the problem, because current implementation efficiency is impacted by the implementation using fixed-size big tensors. 

Look [here](https://github.com/elixir-nx/scholar/blob/ce39654907c0aad1c7dddb424cb3bb9c3a889528/lib/scholar/neighbors/nn_descent.ex#L420)

I also need to change the initialization to use leaves as follows: https://github.com/elixir-nx/scholar/pull/233#issuecomment-1964282253

msluszniak:
Closes #241


lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -278,9 +271,7 @@ defmodule Scholar.Neighbors.RandomProjectionForest do
   @doc """
   Computes approximate nearest neighbors of query tensor using random projection forest.
   Returns the neighbor indices and distances from query points.
-
   ## Examples
-
```
josevalim, line 283: I think those lines in this file were removed by accident? :D

lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -322,8 +313,7 @@ defmodule Scholar.Neighbors.RandomProjectionForest do
     predict_n(forest, query)
   end
 
-  defnp predict_n(forest, query) do
-    k = forest.num_neighbors
+  defn get_leaves(forest, query) do
```
josevalim, line 316: Should we add some docs here, now it is public? Or should we add `@doc false`?

msluszniak: `get_leaves` is a very specific functionality that probably wouldn't be used elsewhere. So I think we might add `@doc false`

lib/scholar/neighbors/random_projection_forest.ex
```diff
@@ -336,13 +326,16 @@ defmodule Scholar.Neighbors.RandomProjectionForest do
...
+    Nx.take(indices, pos)
+    |> Nx.devectorize()
+    |> Nx.rename(nil)
+    |> Nx.transpose(axes: [1, 0, 2])
+    |> Nx.reshape({query_size, num_trees * leaf_size})
+  end
 
+  defnp predict_n(forest, query) do
+    k = forest.num_neighbors
+    candidate_indices = get_leaves(forest, query)
```
krstopro, line 338: I would move `get_leaves` below `predict_n` as the code seems more understandable to me. This is just a personal preference; I am not aware of any conventions imposing this.

msluszniak: Sure

krstopro:
I should have a closer look, but looks great so far.

One thing that might need to be addressed (I forgot this when the first pull request was submitted) is the following. I don't think the current implementation of random projection forest works with any metric other than (squared) Euclidean. We might wanna raise an error if metric is set to something else and `tree_init?` is `true`.

msluszniak:
> I should have a closer look, but looks great so far.
> 
> One thing that might need to be addressed (I forgot this when the first pull request was submitted) is the following. I don't think the current implementation of random projection forest works with any metric other than (squared) Euclidean. We might wanna raise an error or warning if metric is set to something else and `tree_init?` is `true`.

Good call, I'll change this to prevent problematic cases :)

lib/scholar/neighbors/nn_descent.ex
```diff
@@ -127,6 +132,10 @@ defmodule Scholar.Neighbors.NNDescent do
 
     opts = NimbleOptions.validate!(opts, @opts_schema)
 
+    if opts[:tree_init?] and (opts[:metric] != :euclidean and opts[:metric] != :squared_euclidean) do
+      raise ArgumentError, "The tree initialization is only available for Euclidean metrics"
```
krstopro, line 136:
```suggestion
      raise ArgumentError, "The tree initialization is available only for Euclidean metric"
```

krstopro (commented):
Small suggestion regarding the error raised.

lib/scholar/neighbors/nn_descent.ex
```diff
@@ -127,6 +132,10 @@ defmodule Scholar.Neighbors.NNDescent do
 
     opts = NimbleOptions.validate!(opts, @opts_schema)
 
+    if opts[:tree_init?] and (opts[:metric] != :euclidean and opts[:metric] != :squared_euclidean) do
```
josevalim, line 135:
```suggestion
    if opts[:tree_init?] and opts[:metric] not in [:euclidean, :squared_euclidean] do
```

josevalim (approved):

PR #247, opened 2024-04-03
Feat/bayesian ridge

### LINKED ISSUE #244: Implement regularized Bayesian linear algorithms
In general, Bayesian methods trade in greater flexibility for computational expense.

Bayesian regularization algorithms extend the classical Ridge Regression with:

- [ ] a more data driven approach, estimating regularization parameters from data in the case of Bayesian Ridge Regression.
- [ ]  further sparsity in regularization in the case of Automatic Relevance Determination.

I've conducted a [quick](https://github.com/JoaquinIglesiasTurina/scholar/blob/feat/bayesian-ridge/test/scholar/linear/bayesian_ridge_regression_test.exs) and [very dirty](https://github.com/JoaquinIglesiasTurina/scholar/blob/feat/bayesian-ridge/lib/scholar/linear/bayesian_ridge_regression.ex)  proof of concept and concluded that these methods could be implemented  within `defn` , and therefore, fit the goals of Scholar. 

As such, I would like to keep working on this issue.

JoaquinIglesiasTurina:
While I still have to work do, namely:

-   documentation
-   handling the more features than samples case
-   fix internal variable naming of coefficients and whatnot.

I have [non passing](https://github.com/JoaquinIglesiasTurina/scholar/blob/60004c74de9b2b220f0ad3f0855054d592ea9566/test/scholar/linear/bayesian_ridge_regression_test.exs#L82) tests for each of the missing features.
However, the lion's share of the Bayesian Ridge is done and, I wanted to open this PR so code review can get started, and to ask a couple of questions.

# Scores

Given that the computation is done inside a `defn`, I am not sure it is possible to do have it be optional, as is the case in sklearn.

I am afraid the [testing code is not very idiomatic.](https://github.com/JoaquinIglesiasTurina/scholar/blob/cd3822c4f76b9fec9e5fa91ae6e41f761179b229/test/scholar/linear/bayesian_ridge_regression_test.exs#L76)
Also, I've had to make a precision compromise, as I could not get the exact same score with both methods, but the scores are within 5%.
Please compare with the [scikit-learn test.](https://github.com/scikit-learn/scikit-learn/blob/128e40ed593c57e8b9e57a4109928d58fa8bf359/sklearn/linear_model/tests/test_bayes.py#L35)

# Diabetes data

I've brought in [the diabetes data](https://github.com/JoaquinIglesiasTurina/scholar/blob/cd3822c4f76b9fec9e5fa91ae6e41f761179b229/test/support/scholar_case.ex#L198) to replicate scikit-learn tests.
Is this data open source?
Can I just include it?
Is having it as a `.csv` where it is a good idea?

# Matrix inversion times out

I've had some issues with diabetes data matrix inversion. Mainly, it times out the tests.
I've tried running a [linear regression with the diabetes data](https://github.com/JoaquinIglesiasTurina/scholar/blob/cd3822c4f76b9fec9e5fa91ae6e41f761179b229/test/scholar/linear/bayesian_ridge_regression_test.exs#L38), and that also times out, so I'm pretty sure the problem is not only with my implementation of this algorithm.
From what I've gathered this problem is related to [this issue.](https://github.com/elixir-nx/nx/issues/1027) Is that understanding correct?

To avoid the timeout [I've limited the data,](https://github.com/JoaquinIglesiasTurina/scholar/blob/cd3822c4f76b9fec9e5fa91ae6e41f761179b229/test/scholar/linear/bayesian_ridge_regression_test.exs#L50) and checked that I get similar results than with numpy and scikit-learn. However, that is no automated test.

# Module name
I've named the module `BayesianRidgeRegression`, maybe it should just be called `BayesianRidge`.

Closes #244.


lib/scholar/linear/bayesian_ridge_regression.ex
```diff
@@ -0,0 +1,320 @@
...
+    eps = Nx.Constants.smallest_positive_normal(x_type)
+    default_alpha = Nx.divide(1, Nx.add(Nx.variance(x), eps))
+    alpha = Keyword.get(opts, :alpha_init, default_alpha)
+    alpha = Nx.tensor(alpha, type: x_type)
+    opts = Keyword.put(opts, :alpha_init, alpha)
+
+    {lambda, opts} = Keyword.pop!(opts, :lambda_init)
+    lambda = Nx.tensor(lambda, type: x_type)
+    opts = Keyword.put(opts, :lambda_init, lambda)
+    zeros_list = for k <- 0..opts[:iterations], do: 0
```
josevalim, line 128:
```suggestion
    zeros_list = List.duplicate(0, opts[:iterations])
```

lib/scholar/linear/bayesian_ridge_regression.ex
```diff
@@ -0,0 +1,320 @@
...
+        do: Nx.as_type(sample_weights, x_type),
+        else: Nx.tensor(sample_weights, type: x_type)
+
+    # handle vector types
+    # handle default alpha value, add eps to avoid division by 0
+    eps = Nx.Constants.smallest_positive_normal(x_type)
+    default_alpha = Nx.divide(1, Nx.add(Nx.variance(x), eps))
+    alpha = Keyword.get(opts, :alpha_init, default_alpha)
+    alpha = Nx.tensor(alpha, type: x_type)
+    opts = Keyword.put(opts, :alpha_init, alpha)
```
josevalim, line 123: Tensors should never be passed as options. You should always pass all tensors as arguments to the `defn` function. Your job in this function is to:

1. If any option should become a tensor, you remove it from the options and pass argument
2. All other options can be passed directly to the `defn`

lib/scholar/linear/bayesian_ridge_regression.ex
```diff
@@ -0,0 +1,320 @@
...
+    {coefficients, intercept, alpha, lambda, rmse, iterations, has_converged, scores, sigma} =
+      fit_n(x, y, sample_weights, scores, opts)
+    iterations = Nx.to_number(iterations)
+    scores = scores
+    |> Nx.to_list()
+    |> Enum.take(iterations)
+
+    if Nx.to_number(has_converged) == 1 do
+      IO.puts("Convergence after #{Nx.to_number(iterations)} iterations")
+    end
```
josevalim, line 140: You really can't call this here. You have to assume you never can really read the tensor values, even inside `deftransform`. You can call: `Nx.Defn.jit(&BayesianRidgeRegression.fit/3).(x, y, opts)` and you will see these operations will fail. :) You can even add this as a test to check the operation is fully jittable. This means you can't do `Nx.to_list() |> Enum.take(iterations)` either. Whoever calls `fit`has to do that instead.

josevalim:
> Given that the computation is done inside a defn, I am not sure it is possible to do have it be optional, as is the case in sklearn.

It can be optional. You can pass compute_scope as a boolean and, inside the defn, check for this boolean and return nil if you are not computing the scores.

josevalim:
Hi @JoaquinIglesiasTurina! I have dropped some comments, this is looking great and I think we are almost there! :)

lib/scholar/linear/bayesian_ridge_regression.ex
```diff
@@ -0,0 +1,320 @@
...
+    default_alpha = Nx.divide(1, Nx.add(Nx.variance(x), eps))
+    alpha = Keyword.get(opts, :alpha_init, default_alpha)
+    alpha = Nx.tensor(alpha, type: x_type)
+    opts = Keyword.put(opts, :alpha_init, alpha)
+
+    {lambda, opts} = Keyword.pop!(opts, :lambda_init)
+    lambda = Nx.tensor(lambda, type: x_type)
+    opts = Keyword.put(opts, :lambda_init, lambda)
+    zeros_list = for k <- 0..opts[:iterations], do: 0
+    scores = Nx.tensor(zeros_list, type: x_type)
```
krstopro, line 129:
```suggestion
scores = Nx.broadcast(Nx.as_type(0.0, x_type), {opts[:iterations]})
```

krstopro (commented):
Indeed, this looks great. A bit of polishing and we merge it.

krstopro:
I think `sample_weights_flag` can be removed by setting default `sample_weights` to tensor of ones, i.e.
```
sample_weights = Nx.broadcast(Nx.as_type(1.0, x_type), {num_samples})
```
(this holds for few other modules as well)

I am still not sure what is the default way to pass weights as an option. In several modules it is assumed that they are a list of `num_samples` elements. Yet in `Scholar.Options` they are allowed to be `Nx.Tensor` as well, see
https://github.com/JoaquinIglesiasTurina/scholar/blob/feat/bayesian-ridge/lib/scholar/options.ex#L78.

@msluszniak @josevalim Any comments on this one?

lib/scholar/linear/bayesian_ridge_regression.ex
```diff
@@ -0,0 +1,320 @@
...
+      type: :pos_integer,
+      default: 300,
+      doc: """
+      Maximum number of iterations before stopping the fitting algorithm.
+      The number of iterations may be lower is parameters converge.
+      """
+    ],
+    sample_weights: [
+      type:
+        {:or,
```
krstopro, line 21: I would allow `sample_weights` to be only `type: {:custom, Scholar.Options, :weights, []}`.

lib/scholar/linear/bayesian_ridge_regression.ex
```diff
@@ -0,0 +1,320 @@
...
+  deftransform fit(x, y, opts \\ []) do
+    opts = NimbleOptions.validate!(opts, @opts_schema)
+
+    opts =
+      [
+        sample_weights_flag: opts[:sample_weights] != nil
+      ] ++
+        opts
+
+    {sample_weights, opts} = Keyword.pop(opts, :sample_weights, 1.0)
```
krstopro, line 109: I would set default `sample_weights` to `Nx.broadcast(Nx.as_type(1.0, x_type), {num_samples})`.

krstopro (commented):
Few suggestions regarding `sample_weights`.

lib/scholar/linear/bayesian_ridge_regression.ex
```diff
@@ -0,0 +1,320 @@
...
+    lambda = opts[:lambda_init]
+
+    alpha_1 = opts[:alpha_1]
+    alpha_2 = opts[:alpha_2]
+    lambda_1 = opts[:lambda_1]
+    lambda_2 = opts[:lambda_2]
+
+    iterations = opts[:iterations]
+
+    xt_y = Nx.dot(Nx.transpose(x), y)
```
krstopro, line 189:
```suggestion
xt_y = Nx.dot(x, [0], y, [0])
```

krstopro (commented):
You can use `Nx.dot/4` to avoid transposing `x`. For example if
```
key = Nx.Random.key(12)
{a, key} = Nx.Random.normal(key, shape: {3, 3})
{b, key} = Nx.Random.normal(key, shape: {3, 3})
```
then `Nx.dot(Nx.transpose(a), b)` and `Nx.dot(a, [0], b, [0])` both give
```
#Nx.Tensor<
  f32[3][3]
  [
    [0.14595268666744232, -0.05974636226892471, 0.2835732400417328],
    [-1.4989330768585205, -0.24490420520305634, -1.0999470949172974],
    [-0.4686814546585083, 0.25890597701072693, -1.2601317167282104]
  ]
>
```

msluszniak:
> I think `sample_weights_flag` can be removed by setting default `sample_weights` to tensor of ones, i.e.
> 
> ```
> sample_weights = Nx.broadcast(Nx.as_type(1.0, x_type), {num_samples})
> ```
> 
> (this holds for few other modules as well)
> 
> I am still not sure what is the default way to pass weights as an option. In several modules it is assumed that they are a list of `num_samples` elements. Yet in `Scholar.Options` they are allowed to be `Nx.Tensor` as well, see https://github.com/elixir-nx/scholar/blob/main/lib/scholar/options.ex#L78.
> 
> @msluszniak @josevalim Any comments on this one?

I think it's a good idea to try to unify this across all modules. I think we might create a separate issue and point to your comments. Do you want me to create this issue?

krstopro:
> > I think `sample_weights_flag` can be removed by setting default `sample_weights` to tensor of ones, i.e.
> > ```
> > sample_weights = Nx.broadcast(Nx.as_type(1.0, x_type), {num_samples})
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (this holds for few other modules as well)
> > I am still not sure what is the default way to pass weights as an option. In several modules it is assumed that they are a list of `num_samples` elements. Yet in `Scholar.Options` they are allowed to be `Nx.Tensor` as well, see https://github.com/elixir-nx/scholar/blob/main/lib/scholar/options.ex#L78.
> > @msluszniak @josevalim Any comments on this one?
> 
> I think it's a good idea to try to unify this across all modules. I think we might create a separate issue and point to your comments. Do you want me to create this issue?

I would move this in a separate issue. Same for input validation, e.g. checking if `x` is of rank 2, if `y` is of rank 1, if first dimension of `x` and `y` match, etc.

test/scholar/linear/bayesian_ridge_regression_test.exs
```diff
@@ -0,0 +1,120 @@
...
+  
+  test "compute scores" do
+    {x, y} = diabetes_data()
+    n_samples = 50 - 1
+    x = x[[0..n_samples, ..]]
+    IO.inspect(x)
+    y = y[[0..n_samples, ..]]
+    eps = Nx.Constants.smallest_positive_normal({:f, 64})
+    alpha = Nx.divide(1, Nx.add(Nx.variance(x), eps))
+    lambda = Nx.tensor(1.0)
```
msluszniak, line 55:
```suggestion
    eps = Nx.Constants.smallest_positive_normal(:f64)
    alpha = 1 / (Nx.variance(x) + eps)
    lambda = 1.0
```

test/scholar/linear/bayesian_ridge_regression_test.exs
```diff
@@ -0,0 +1,120 @@
...
+    compare_scores = Nx.divide(Nx.subtract(score, first_score), score)
+    check = Nx.less(compare_scores, 0.05) |> Nx.flatten()
+    assert check == Nx.tensor([1], type: {:u, 8})
+  end
+
+  defnp compute_score(x, y, alpha, lambda, alpha_1, alpha_2, lambda_1, lambda_2) do
+    {n_samples, _} = Nx.shape(x)
+    lambda_score = lambda_1 * Nx.log(lambda) - lambda_2 * lambda
+    alpha_score = alpha_1 * Nx.log(alpha) - alpha_2 * alpha
+    m = 1.0 / alpha * Nx.eye(n_samples) + 1.0 / lambda * Nx.dot(x, Nx.transpose(x))
```
msluszniak, line 85: If I remember correctly
```suggestion
    m = 1.0 / alpha * Nx.eye(n_samples) + 1.0 / lambda * Nx.dot(x, [-1], x, [-1])
```

msluszniak:
Thank you for the PR, I also dropped some minor suggestions :))

lib/scholar/linear/bayesian_ridge_regression.ex
JoaquinIglesiasTurina: My understanding is that model [hyper parameters](https://github.com/JoaquinIglesiasTurina/scholar/blob/0006e71592812485cfd48808559d0f01259bc277/lib/scholar/linear/bayesian_ridge_regression.ex#L59), which are plain scalars, are [ok to be passed as opts](https://github.com/JoaquinIglesiasTurina/scholar/blob/0006e71592812485cfd48808559d0f01259bc277/lib/scholar/linear/bayesian_ridge_regression.ex#L182)  to `defn`.
Is that understanding correct? 

Please note that those hyperparameters need to be pased to the `while` loop. So, they get converted to tensors, there. Right?

JoaquinIglesiasTurina:
Thank you all for the code review. I think I've addressed all the points brought up.

Right now, I'm missing documentation. Once that is done, I will push again, to avoid triggering the CI pipeline.

lib/scholar/linear/bayesian_ridge_regression.ex
josevalim: It depends. Everything that is passed as an option to a defn results in a different compilation. So, if you can convert something to a tensor and pass it as a tensor, that's ideally better. But that's not always possible. Sometimes the value is used as a shape, and different shapes always lead to different compilations.

> Please note that those hyperparameters need to be pased to the while loop. So, they get converted to tensors, there. Right?

Correct.

krstopro:
> > > I think `sample_weights_flag` can be removed by setting default `sample_weights` to tensor of ones, i.e.
> > > ```
> > > sample_weights = Nx.broadcast(Nx.as_type(1.0, x_type), {num_samples})
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > (this holds for few other modules as well)
> > > I am still not sure what is the default way to pass weights as an option. In several modules it is assumed that they are a list of `num_samples` elements. Yet in `Scholar.Options` they are allowed to be `Nx.Tensor` as well, see https://github.com/elixir-nx/scholar/blob/main/lib/scholar/options.ex#L78.
> > > @msluszniak @josevalim Any comments on this one?
> > 
> > 
> > I think it's a good idea to try to unify this across all modules. I think we might create a separate issue and point to your comments. Do you want me to create this issue?
> 
> I would move this in a separate issue. Same for input validation, e.g. checking if `x` is of rank 2, if `y` is of rank 1 or 2, if first dimension of `x` and `y` match, etc.

@msluszniak By "I would move this in a separate issue" I meant "you are welcome to open a separate issue for this". Sorry, I now realise I wasn't clear enough. 😅

JoaquinIglesiasTurina:
From my end, I don´t think there is anything else left to do. Please let me know if you think otherwise.

krstopro:
> From my end, I don´t think there is anything else left to do. Please let me know if you think otherwise.

@JoaquinIglesiasTurina Sure, will have a look later today.

lib/scholar/linear/bayesian_ridge_regression.ex
```diff
@@ -0,0 +1,480 @@
...
+    iterations = opts[:iterations]
+
+    xt_y = Nx.dot(x, [0], y, [0])
+    {u, s, vh} = Nx.LinAlg.svd(x, full_matrices?: false)
+    eigenvals = s ** 2
+    {n_samples, n_features} = Nx.shape(x)
+    {coef, rmse} = update_coef(x, y, n_samples, n_features, xt_y, u, vh, eigenvals, alpha, lambda)
+
+    {{coef, alpha, lambda, _rmse, iter, has_converged, scores}, _} =
+      while {{coef, rmse, alpha, lambda, iter = 0, has_converged = Nx.u8(0), scores = scores},
```
krstopro, line 323:
```suggestion
while {{coef, rmse, alpha, lambda, iter = Nx.u64(0), has_converged = Nx.u8(0), scores = scores},
```

lib/scholar/linear/bayesian_ridge_regression.ex
```diff
@@ -0,0 +1,480 @@
...
+  # Implements sample weighting by rescaling inputs and
+  # targets by sqrt(sample_weight).
+  defnp rescale(x, y, sample_weights) do
+    case Nx.shape(sample_weights) do
+      {} = scalar ->
+        scalar = Nx.sqrt(scalar)
+        {scalar * x, scalar * y}
+
+      _ ->
+        scale = sample_weights |> Nx.sqrt() |> Nx.make_diagonal()
```
krstopro, line 460: This can be simplified by doing
```suggestion
scale = sample_weights |> Nx.sqrt() |> Nx.new_axis(1)
```
and using `*` instead of `Nx.dot`.

krstopro (commented):
Few minor comments.

lib/scholar/linear/bayesian_ridge_regression.ex
JoaquinIglesiasTurina: I do not believe this change is possible.
Please note that `a` is a `{n_samples}` sized vector, `b` is an `{n_samples, n_features}` matrix and `sample_weights` is an `{n_samples}` sized vector.

`sample_weights |> Nx.sqrt() |> Nx.new_axis(1)` yields an `{n_samples, 1}` sized matrix, that cannot be multiplied with `a`.

As far as I can tell, `sample_weights |> Nx.sqrt() |> Nx.make_diagonal()` yields the only matrix that is dottable with both `a` and `b`.

An alternative would be keeping 2 `scale` tensors, one for `a` and one for `b`. I personally do not like this option as it would unbalance the function. You would have a different tensor for each piece of data, and the operations would look different than the other branch of the `case` statement.

Please, let me know if you think of a better way.

krstopro: You are correct, but we should be able to use `*` instead of `Nx.dot`. This should work:
```
defnp rescale(x, y, sample_weights) do
  factor = Nx.sqrt(sample_weights)
  x_scaled = case Nx.shape(factor) do
    {} -> factor * x
    _ -> Nx.new_axis(factor, 1) * x
  end
  y_scaled = factor * y
  {x_scaled, y_scaled}
end
```

JoaquinIglesiasTurina: This works and I find it's a pretty clean solution. Thank you for your comments.

krstopro (approved):

josevalim:

PR #248, opened 2024-04-05
Multinomial Naive Bayes Improvements

### LINKED ISSUE #242: Multinomial Naive Bayes bug
I have been doing some text classification in python and have been trying to get similar results in Elixir using scholar, but I am having issues. The multinomial NB Elixir implementation looks like it is swapping the class count and class log priors for the two classes. I have attached two zip files containing .npy files to help replicate the issue.

[x-train-data.zip](https://github.com/elixir-nx/scholar/files/14547172/x-train-data.zip)
[y-train-data.zip](https://github.com/elixir-nx/scholar/files/14547173/y-train-data.zip)

Here is an example python script to load the unzipped files and fit the multinomial NB classifier.
```python
from pprint import pprint
import numpy as np
from sklearn.naive_bayes import MultinomialNB

x_train = np.load('< path to x-train-data.npy>')
y_train = np.load('< path to y-train-data.npy> ')

classifier = MultinomialNB()
classifier.fit(x_train, y_train)
pprint(vars(classifier))
```

Here is an Elixir example to load these files and fit the multinomial NB scholar classifier.
```elixir
Mix.install([
  {:scholar, "~> 0.2.1"},
  {:exla, ">= 0.0.0"}
])

Nx.global_default_backend(EXLA.Backend)

x_train =
  "< path to x-train-data.npy>"
  |> File.read!()
  |> Nx.load_numpy!()

y_train =
  "<path to y-train-data.npy>"
  |> File.read!()
  |> Nx.load_numpy!()

model = Scholar.NaiveBayes.Multinomial.fit(x_train, y_train, num_classes: 2)
```

If you log the Elixir multinomial NB model it will show class 0 as being in the data 596 times and class 1 being in the data 604 times. These numbers are actually backwards class 0 is in the data 604 times and class 1 is in the data 596 times.

If you look at the logged python classifier it shows the correct counts.

krstopro:
Changelog:
- Removed `classes` property and `ordinal_encode` from `fit`.
- Updated docstring to specify that targets should be between 0 and `num_classes` - 1.
- Some refactoring, e.g. used `Nx.logsumexp`, moved argument validation inside `fit`, etc.
- Fixes #242.

### DIFF
--- a/lib/scholar/naive_bayes/multinomial.ex
+++ b/lib/scholar/naive_bayes/multinomial.ex
@@ -179,10 +179,10 @@ defmodule Scholar.NaiveBayes.Multinomial do
 
     opts = NimbleOptions.validate!(opts, @opts_schema)
 
-    x_type = to_float_type(x)
+    type = to_float_type(x)
 
     {alpha, opts} = Keyword.pop!(opts, :alpha)
-    alpha = Nx.tensor(alpha, type: x_type)
+    alpha = Nx.tensor(alpha, type: type)
 
     if Nx.shape(alpha) not in [{}, {num_features}] do
       raise ArgumentError,
@@ -210,7 +210,7 @@ defmodule Scholar.NaiveBayes.Multinomial do
     sample_weights_flag = opts[:sample_weights] != nil
 
     {sample_weights, opts} = Keyword.pop(opts, :sample_weights, :nan)
-    sample_weights = Nx.tensor(sample_weights, type: x_type)
+    sample_weights = Nx.tensor(sample_weights, type: type)
 
     if sample_weights_flag and Nx.shape(sample_weights) != {num_samples} do
       raise ArgumentError,
@@ -223,6 +223,7 @@ defmodule Scholar.NaiveBayes.Multinomial do
     opts =
       opts ++
         [
+          type: type,
           priors_flag: priors_flag,
           sample_weights_flag: sample_weights_flag
         ]
@@ -231,8 +232,7 @@ defmodule Scholar.NaiveBayes.Multinomial do
   end
 
   defnp fit_n(x, y, class_priors, sample_weights, alpha, opts) do
-    # TODO: Why not just to_float_type?
-    x_type = Nx.Type.merge(to_float_type(x), {:f, 32})
+    type = opts[:type]
     num_samples = Nx.axis_size(x, 0)
     num_classes = opts[:num_classes]
 
@@ -240,14 +240,14 @@ defmodule Scholar.NaiveBayes.Multinomial do
       y
       |> Nx.new_axis(1)
       |> Nx.equal(Nx.iota({num_samples, num_classes}, axis: 1))
-      |> Nx.as_type(x_type)
+      |> Nx.as_type(type)
 
     y_weighted =
       if opts[:sample_weights_flag],
         do: Nx.reshape(sample_weights, {num_samples, 1}) * y_one_hot,
         else: y_one_hot
 
-    alpha_lower_bound = Nx.tensor(1.0e-10, type: x_type)
+    alpha_lower_bound = Nx.tensor(1.0e-10, type: type)
 
     alpha =
       if opts[:force_alpha], do: alpha, else: Nx.max(alpha, alpha_lower_bound)
--- a/lib/scholar/naive_bayes/multinomial.ex
+++ b/lib/scholar/naive_bayes/multinomial.ex
@@ -2,6 +2,9 @@ defmodule Scholar.NaiveBayes.Multinomial do
   @moduledoc """
   Naive Bayes classifier for multinomial models.
 
+  The multinomial Naive Bayes classifier is suitable for classification with
+  discrete features (e.g., word counts for text classification)
+
   Time complexity is $O(K * N * C)$ where $N$ is the number of samples and $K$ is the number of features,
   and $C$ is the number of classes.
   """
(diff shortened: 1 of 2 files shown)


lib/scholar/naive_bayes/multinomial.ex
```diff
@@ -81,87 +84,201 @@ defmodule Scholar.NaiveBayes.Multinomial do
...
+          priors_flag: priors_flag,
+          sample_weights_flag: sample_weights_flag
+        ]
+
+    fit_n(x, y, class_priors, sample_weights, alpha, opts)
+  end
+
+  defnp fit_n(x, y, class_priors, sample_weights, alpha, opts) do
+    # TODO: Why not just to_float_type?
+    x_type = Nx.Type.merge(to_float_type(x), {:f, 32})
```
krstopro, line 235: @msluszniak I am not sure why `Nx.Type.merge/2` is used here. Do you maybe remember?

krstopro:
Another thing is that `Nx.logsumexp/2` should be pulled out of [nx.ex](https://github.com/elixir-nx/nx/blob/main/nx/lib/nx.ex) file. I might create a separate issue for that.

polvalente:
> Another thing is that `Nx.logsumexp/2` should be pulled out of [nx.ex](https://github.com/elixir-nx/nx/blob/main/nx/lib/nx.ex) file. I might create a separate issue for that.

Why?

krstopro:
> > Another thing is that `Nx.logsumexp/2` should be pulled out of [nx.ex](https://github.com/elixir-nx/nx/blob/main/nx/lib/nx.ex) file. I might create a separate issue for that.
> 
> Why?

@polvalente Yeah, I should have written the reason. :)
It's because [the maximum value that is subtracted from the exponentials](https://github.com/elixir-nx/nx/blob/bb056ce6aa49a114a0a248624609b336b1219f7e/nx/lib/nx.ex#L17334-L17337) should be treated as constant during differentiation [as done in Axon and JAX](https://github.com/elixir-nx/nx/blob/bb056ce6aa49a114a0a248624609b336b1219f7e/nx/lib/nx.ex#L17334-L17337). Correct me if I'm wrong - this cannot be done for functions in `nx.ex` file?
Alternative would be to have separate implementations in `Scholar` and `Axon`, but I would prefer to have it in `Nx` because I consider the function fundamental.

polvalente:
> > > Another thing is that `Nx.logsumexp/2` should be pulled out of [nx.ex](https://github.com/elixir-nx/nx/blob/main/nx/lib/nx.ex) file. I might create a separate issue for that.
> > 
> > Why?
> 
> @polvalente Yeah, I should have written the reason. :)
> It's because [the maximum value that is subtracted from the exponentials](https://github.com/elixir-nx/nx/blob/bb056ce6aa49a114a0a248624609b336b1219f7e/nx/lib/nx.ex#L17334-L17337) should be treated as constant during differentiation [as done in Axon and JAX](https://github.com/elixir-nx/nx/blob/bb056ce6aa49a114a0a248624609b336b1219f7e/nx/lib/nx.ex#L17334-L17337). Correct me if I'm wrong - this cannot be done for functions in `nx.ex` file?
> Alternative would be to have separate implementations in `Scholar` and `Axon`, but I would prefer to have it in `Nx` because I consider the function fundamental.

Ah, we can probably fix this by setting stop_grad there :)

krstopro:
> > > > Another thing is that `Nx.logsumexp/2` should be pulled out of [nx.ex](https://github.com/elixir-nx/nx/blob/main/nx/lib/nx.ex) file. I might create a separate issue for that.
> > > 
> > > 
> > > Why?
> > 
> > 
> > @polvalente Yeah, I should have written the reason. :)
> > It's because [the maximum value that is subtracted from the exponentials](https://github.com/elixir-nx/nx/blob/bb056ce6aa49a114a0a248624609b336b1219f7e/nx/lib/nx.ex#L17334-L17337) should be treated as constant during differentiation [as done in Axon and JAX](https://github.com/elixir-nx/nx/blob/bb056ce6aa49a114a0a248624609b336b1219f7e/nx/lib/nx.ex#L17334-L17337). Correct me if I'm wrong - this cannot be done for functions in `nx.ex` file?
> > Alternative would be to have separate implementations in `Scholar` and `Axon`, but I would prefer to have it in `Nx` because I consider the function fundamental.
> 
> Ah, we can probably fix this by setting stop_grad there :)

I thought I tried it but it didn't work. Probably a mistake by me. :)

polvalente:
> > > > > Another thing is that `Nx.logsumexp/2` should be pulled out of [nx.ex](https://github.com/elixir-nx/nx/blob/main/nx/lib/nx.ex) file. I might create a separate issue for that.
> > > > 
> > > > 
> > > > Why?
> > > 
> > > 
> > > @polvalente Yeah, I should have written the reason. :)
> > > It's because [the maximum value that is subtracted from the exponentials](https://github.com/elixir-nx/nx/blob/bb056ce6aa49a114a0a248624609b336b1219f7e/nx/lib/nx.ex#L17334-L17337) should be treated as constant during differentiation [as done in Axon and JAX](https://github.com/elixir-nx/nx/blob/bb056ce6aa49a114a0a248624609b336b1219f7e/nx/lib/nx.ex#L17334-L17337). Correct me if I'm wrong - this cannot be done for functions in `nx.ex` file?
> > > Alternative would be to have separate implementations in `Scholar` and `Axon`, but I would prefer to have it in `Nx` because I consider the function fundamental.
> > 
> > Ah, we can probably fix this by setting stop_grad there :)
> 
> I thought I tried it but it didn't work. Probably a mistake by me. :)

I'll check this out in a few min!

polvalente:
https://github.com/elixir-nx/nx/pull/1470 @krstopro

lib/scholar/naive_bayes/multinomial.ex
polvalente: I can see this being the case if the algorithm just doesn't work with precision less than f32. My SVD impl in Nx for instance only works properly for f32, not even f64.

msluszniak (approved):

lib/scholar/naive_bayes/multinomial.ex
msluszniak: Yes, that was exactly my intention.

krstopro: @msluszniak So, the algorithm doesn't work with `f16`? Any specific part?

msluszniak: Actually, I tested the implementation and it works for `:bf16` and `:f16`, so it's rather over-zeal 😅. Feel free to delete these merges.

krstopro: Great, thanks!

Issue #246, opened 2024-03-29
Implement PCA extensions


krstopro:
We already have PCA implemented. Two extensions that might be worth having:
- [ ] Incremental PCA. This was already requested [here](https://elixirforum.com/t/how-to-execute-incremental-pca-or-similar-batch-tasks-in-nx-scholar/59039).
- [ ] Kernel PCA.

RicardoSantos-99:
Both items here are done now: Incremental PCA via `PCA.incremental_fit/2` and Kernel PCA from [#337](https://github.com/elixir-nx/scholar/pull/337). I think this can be closed. @josevalim

PR #253, opened 2024-04-08
Batch k-NN

### LINKED ISSUE #239: Implement batch version of brute-force k-NN search
Current implementation of brute-force k-NN search takes $O(N^2 \times D)$ memory for sample size $N$ and dimension $D$. This quickly becomes intractable and can be reduced to $O(B \times N \times D)$ for batch size $B$. This could be important for benchmarking the approximate k-NN methods; in order to calculate the recall one needs to compute the exact nearest neighbors and the only way to do this for large $D$ is via brute-force search.

krstopro:
I am honestly not sure where should this be implemented. Right now I added `linear_search` (a better name might be needed, e.g. `brute_force_search`) inside `Scholar.Neighbors.Utils`. Like this it can be used inside other modules such as `Scholar.Neighbors.KNearestNeighbors` or dimensionality reduction algorithms (t-SNE, Trimap, PacMAP).

The function itself should be documented and tested. More distances are needed. I am just submitting a draft to get some feedback.

Closes #239.

### DIFF
--- a/lib/scholar/neighbors/utils.ex
+++ b/lib/scholar/neighbors/utils.ex
@@ -2,11 +2,89 @@ defmodule Scholar.Neighbors.Utils do
   @moduledoc false
   import Nx.Defn
   require Nx
+  import Scholar.Shared
+
+  defn linear_search(data, query, opts) do
+    k = opts[:num_neighbors]
+    metric = opts[:metric]
+    batch_size = opts[:batch_size]
+
+    type = Nx.Type.merge(to_float_type(data), to_float_type(query))
+
+    distance_fn =
+      case metric do
+        {:minkowski, p} ->
+          fn data, query ->
+            query
+            |> Nx.new_axis(1)
+            |> Nx.subtract(data)
+            |> Nx.pow(p)
+            |> Nx.sum(axes: [2])
+          end
+
+        {:cosine} ->
+          &Scholar.Metrics.Distance.pairwise_cosine/2
+      end
+
+    {query_size, dim} = Nx.shape(query)
+    num_batches = div(query_size, batch_size)
+    leftover_size = rem(query_size, batch_size)
+
+    batches =
+      query
+      |> Nx.slice_along_axis(0, num_batches * batch_size, axis: 0)
+      |> Nx.reshape({num_batches, batch_size, dim})
+
+    {neighbor_indices, neighbor_distances, _} =
+      while {
+              neighbor_indices = Nx.broadcast(Nx.u64(0), {num_batches, batch_size, k}),
+              neighbor_distances =
+                Nx.broadcast(Nx.as_type(:nan, type), {num_batches, batch_size, k}),
+              {
+                data,
+                batches,
+                i = 0
+              }
+            },
+            i < num_batches do
+        batch = batches[i]
+        distances = distance_fn.(data, batch)
+
+        indices = Nx.argsort(distances, axis: 1, type: :u64) |> Nx.slice_along_axis(0, k, axis: 1)
+        distances = Nx.take_along_axis(distances, indices, axis: 1)
+
+        neighbor_indices = Nx.put_slice(neighbor_indices, [i, 0, 0], Nx.new_axis(indices, 0))
+
+        neighbor_distances =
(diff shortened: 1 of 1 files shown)


josevalim:
It looks good to me. For now it is a private module, so we don't have to sweat too much about the name. Does this function have any relationship with find_neighbour? If so, maybe we call one `linear_search` and the other `batch_linear_search`?

krstopro:
I would rename `find_neighbors` to `linear_search_with_candidates` as that is what it really is: it performs linear search, but only on indices specified with `candidates` tensor.
https://github.com/elixir-nx/scholar/blob/0d1bcc106b8a1486d08eb67faba2f4326c3d5a7e/lib/scholar/neighbors/utils.ex#L13

msluszniak:
There is a function for handling multiple distances that are in `t-SNE`, `NNDescent`, and `Trimap` modules. I guess it might be worth to move it to Scholar.Shared module.

lib/scholar/neighbors/utils.ex
```diff
@@ -2,11 +2,89 @@ defmodule Scholar.Neighbors.Utils do
...
+
+    neighbor_indices = Nx.reshape(neighbor_indices, {num_batches * batch_size, k})
+    neighbor_distances = Nx.reshape(neighbor_distances, {num_batches * batch_size, k})
+
+    {neighbor_indices, neighbor_distances} =
+      if leftover_size > 0 do
+        leftover = Nx.slice_along_axis(query, query_size - leftover_size, leftover_size, axis: 0)
+        distances = distance_fn.(data, leftover)
+        indices = Nx.argsort(distances, axis: 1, type: :u64) |> Nx.slice_along_axis(0, k, axis: 1)
+        distances = Nx.take_along_axis(distances, indices, axis: 1)
```
msluszniak, line 72: These lines are common for the main part of the search in the while loop, maybe it is worth moving this into a separate function, but it's up to you since it's just 3 lines

lib/scholar/neighbors/utils.ex
```diff
@@ -2,11 +2,89 @@ defmodule Scholar.Neighbors.Utils do
...
+
+    {neighbor_indices, neighbor_distances, _} =
+      while {
+              neighbor_indices = Nx.broadcast(Nx.u64(0), {num_batches, batch_size, k}),
+              neighbor_distances =
+                Nx.broadcast(Nx.as_type(:nan, type), {num_batches, batch_size, k}),
+              {
+                data,
+                batches,
+                i = 0
```
msluszniak, line 46:
```suggestion
                i = Nx.u64(0)
```

lib/scholar/neighbors/utils.ex
krstopro: I'm considering it. The code still needs some polishing.
I am also thinking of abstracting the whole thing as kind of map over batches. It might be useful for Task 1 of #246 as well.

lib/scholar/neighbors/utils.ex
```diff
@@ -2,11 +2,89 @@ defmodule Scholar.Neighbors.Utils do
...
+          end
+
+        {:cosine} ->
+          &Scholar.Metrics.Distance.pairwise_cosine/2
+      end
+
+    {query_size, dim} = Nx.shape(query)
+    num_batches = div(query_size, batch_size)
+    leftover_size = rem(query_size, batch_size)
+
```
krstopro, line 32: This part might also need to be moved inside `Scholar.Shared` as a function that takes a tensor and batch size and returns `{batches, leftover}` where `batches` is a tensor of shape `{num_batches, batch_size, dim}` and `leftover` is a tensor of shape `{leftover_size, dim}`.
Might be relevant for #246, task 1.

msluszniak: Yes, definitely. In Python, there is even a function divmod that do that

josevalim:
`linear_search` and `linear_candidate_search` (or `linear_neighbour_search`) are fine to me! As I said, it is your call, the whole API is private, so focus on what makes the code cleaner :)

krstopro:
Closing this in favor of #257.

Issue #255, opened 2024-04-11
Refactor Scholar.Neighbors


krstopro:
As argued in #254, we need to update `Scholar.Neighbors` in the following way:
- [x] Add `BruteKNN` module that implements brute-force k-NN search. Its `predict` function has to return `{neighbor_indices, neighbor_distances}`. Make `batch_size` an option (as in #253).
- [x] Implement `KNNClassifier`.
- [x] Implement `KNNRegressor`.
- [x] Deprecate `Scholar.Neighbors.KNearestNeighbors`.

`KNNClassifier` and `KNNRegressor` have to take k-NN algorithm and metric options as atoms, but also allow for custom k-NN algorithms and metrics to be passed as modules and functions, respectively.

josevalim:
We don't need to deprecate, we can just rename the module. It should be easy for anyone to just rename it accordingly. :)

krstopro:
> KNNClassifier and KNNRegressor have to take k-NN algorithm and metric options as atoms, but also allow for custom k-NN algorithms and metrics to be passed as modules and functions, respectively.

Is it possible to pass an anonymous function as an option to function defined with `defn`?

josevalim:
Hi @krstopro, I want to release a new version soon. If we plant to remove `Scholar.Neighbors.KNearestNeighbors`, it would be better to do it before, so we update the docs accordingly?

josevalim:
We would also need to mirror those decisions on top of `RadiusNearestNeighbors`.

krstopro:
> Hi @krstopro, I want to release a new version soon. If we plant to remove `Scholar.Neighbors.KNearestNeighbors`, it would be better to do it before, so we update the docs accordingly?

@josevalim Agreed. I will try to implement the stuff from this issue by the end of the week if that's fine.

josevalim:
Perfect. KNN is the most important, Radius can be done later. :)

josevalim:
Thank you @krstopro! I will work on #265 and @msluszniak will update the notebooks. :) Once KNNRegression is ready, we can release a new version!

PR #257, opened 2024-04-13
Brute k-NN

### LINKED ISSUE #239: Implement batch version of brute-force k-NN search

krstopro:
Implements brute-force k-NN search algorithm.

Closes #239.

### DIFF
--- a/lib/scholar/neighbors/brute_knn.ex
+++ b/lib/scholar/neighbors/brute_knn.ex
@@ -212,6 +212,27 @@ defmodule Scholar.Neighbors.BruteKNN do
     {neighbor_indices, neighbor_distances}
   end
 
+  defn get_batches(tensor, opts) do
+    {size, dim} = Nx.shape(tensor)
+    batch_size = opts[:batch_size]
+    num_batches = div(size, batch_size)
+    leftover_size = rem(size, batch_size)
+
+    batches =
+      tensor
+      |> Nx.slice_along_axis(0, num_batches * batch_size, axis: 0)
+      |> Nx.reshape({num_batches, batch_size, dim})
+
+    leftover =
+      if leftover_size > 0 do
+        Nx.slice_along_axis(tensor, num_batches * batch_size, leftover_size, axis: 0)
+      else
+        nil
+      end
+
+    {batches, leftover}
+  end
+
   defnp brute_force_search(data, query, opts) do
     k = opts[:num_neighbors]
     metric = opts[:metric]
--- a/lib/scholar/neighbors/brute_knn.ex
+++ b/lib/scholar/neighbors/brute_knn.ex
@@ -3,10 +3,10 @@ defmodule Scholar.Neighbors.BruteKNN do
   Brute-Force k-Nearest Neighbor Search Algorithm.
 
   In order to find the k-nearest neighbors the algorithm calculates
-  the distance between query point and each of data samples.
+  the distance between the query point and each of the data samples.
   Therefore, its time complexity is $O(MN)$ for $N$ samples and $M$ query points.
-  It uses $O(BN)$ memory for batch size B.
-  Larger batch sizes will be faster, but will consume more memory.
+  It uses $O(BN)$ memory for batch size $B$.
+  Larger batch sizes will lead to faster predictions, but will consume more memory.
   """
   import Nx.Defn
   import Scholar.Shared
@@ -98,7 +98,8 @@ defmodule Scholar.Neighbors.BruteKNN do
   end
 
   @doc """
-  ...
+  Computes nearest neighbors of query tensor using brute-force search.
+  Returns the neighbors indices and distances from query points.
 
   ## Examples
 
--- a/lib/scholar/neighbors/brute_knn.ex
+++ b/lib/scholar/neighbors/brute_knn.ex
@@ -1,0 +1,237 @@
+defmodule Scholar.Neighbors.BruteKNN do
+  @moduledoc """
+  Brute-Force k-Nearest Neighbor Search Algorithm.
+
+  <how it works> .. i.e. ...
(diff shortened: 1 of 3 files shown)


lib/scholar/neighbors/brute_knn.ex
```diff
@@ -0,0 +1,230 @@
...
+  defstruct [:num_neighbors, :metric, :data, :batch_size]
+
+  opts = [
+    num_neighbors: [
+      required: true,
+      type: :pos_integer,
+      doc: "The number of nearest neighbors."
+    ],
+    metric: [
+      type: {:or, [{:custom, Scholar.Options, :metric, []}, {:fun, 2}]},
```
krstopro, line 25: Perhaps `Scholar.Options.metric` should be edited to support functions of arity 2?

lib/scholar/neighbors/brute_knn.ex
```diff
@@ -0,0 +1,230 @@
...
+    metric =
+      case opts[:metric] do
+        {:minkowski, p} ->
+          &Scholar.Metrics.Distance.minkowski(&1, &2, p: p)
+
+        :cosine ->
+          &Scholar.Metrics.Distance.cosine/2
+
+        fun when is_function(fun, 2) ->
+          fun
```
krstopro, line 89: I am not 100% sure about this. Could there be issues with `Nx` backends?

lib/scholar/shared.ex
```diff
@@ -89,4 +89,25 @@ defmodule Scholar.Shared do
 
     valid_broadcast(to_parse - 1, n_dims, shape1, shape2)
   end
+
+  defn get_batches(tensor, opts) do
```
krstopro, line 93: I cannot think of a better name here.

lib/scholar/neighbors/brute_knn.ex
josevalim: If everywhere we accept a metric we also accept functions, then yes. The easiest is probably to make it so it always returns a 2-arity function and then we call it. :)

lib/scholar/neighbors/brute_knn.ex
```diff
@@ -0,0 +1,230 @@
...
+    if Nx.rank(data) != 2 do
+      raise ArgumentError,
+            "expected input tensor to have shape {n_samples, n_features},
+             got tensor with shape: #{inspect(Nx.shape(data))}"
+    end
+
+    opts = NimbleOptions.validate!(opts, @opts_schema)
+
+    metric =
+      case opts[:metric] do
```
josevalim, line 81: Following up above: I would move this normalization to Scholar.Options.metric then. This way everywhere can rely on it being a 2-arity function.

lib/scholar/shared.ex
josevalim: Is it going to be used in other places? If so, where? Maybe keep it on BruteKNN until something else needs it? Btw, should this PR remove the current KNearestNeighbours module?

lib/scholar/neighbors/brute_knn.ex
krstopro: You mean instead of returning the atom to return the anonymous function of arity 2?
I agree, but some other modules need to be edited as well. For example k-d tree:
https://github.com/elixir-nx/scholar/blob/0d1bcc106b8a1486d08eb67faba2f4326c3d5a7e/lib/scholar/neighbors/kd_tree.ex#L33-L43
And here is how it is used there:
https://github.com/elixir-nx/scholar/blob/0d1bcc106b8a1486d08eb67faba2f4326c3d5a7e/lib/scholar/neighbors/kd_tree.ex#L283-L286
I would honestly prefer metric to be stored as a function inside a field.

krstopro: The main problem is that different algorithms support different metrics. Brute-force search works with literally any metric, while current implementation of random projection forest works only with Euclidean distance. I am not sure about k-d tree; I think it works only with the three metrics specified in the docs.

josevalim: It is fine then to keep this logic only in this module then :) IMO

lib/scholar/shared.ex
josevalim: > Not 100% sure, but I strongly believe we could use it for incremental PCA (https://github.com/elixir-nx/scholar/issues/246, task 1).

So I would keep it in the BruteKNN module. We extract it when there is a use case.

> KNearestNeighbors is a supervised model; it performs classification or regression. I would keep it until we implement KNNClassifier and KNNRegressor (task 2 and 3 of https://github.com/elixir-nx/scholar/issues/255).

Ok!

lib/scholar/neighbors/brute_knn.ex
josevalim: No, this is fine!

lib/scholar/neighbors/brute_knn.ex
krstopro: Agreed. I would still suggest converting atoms to functions and storing them as fields. Seems cleaner and doesn't require choosing the right function to call each time `fit` is used.
Also, it might be worth considering removing `Scholar.Options.metric` given the differences in metrics supported between modules.

josevalim (approved):

krstopro:
I guess this is ready to merge. ^_^

PR #262, opened 2024-05-07
Add livebook

### LINKED ISSUE #237: Add a notebook comparing the outputs of different neighbors algorithms
[Example](https://www.researchgate.net/figure/Application-of-t-SNE-LargeVis-UMAP-TriMAP-and-PaCMAP-to-the-Mammoth-dataset-In_fig1_346766201).

msluszniak:
Closes #237

### DIFF
--- a/notebooks/nearest_neighbors.livemd
+++ b/notebooks/nearest_neighbors.livemd
@@ -59,7 +59,7 @@ Now let's move to a more abstract example. You're listening to your favorite roc
 
 One type of [recommendation system](https://en.wikipedia.org/wiki/Recommender_system) relies on collaborative filtering: it recommends songs based on what other users with similar musical tastes listen to. Another approach is to treat songs as points, like in our earlier definition, and then compute the closest songs to your favorites. This is another example of calculating nearest neighbors. While it might not be immediately obvious how to represent songs as points and measure the "distance" between them, it turns out that it can be done.
 
-Another use of nearest neighbors is in building classifiers and regressors. This approach was covered in a [previous notebook](https://github.com/elixir-nx/scholar/blob/main/notebooks/k_nearest_neighbors.livemd), which we strongly recommend you review before proceeding with this one.
+Another use of nearest neighbors is in building classifiers and regressors. This approach was covered in a [previous notebook](k_nearest_neighbors.livemd), which we strongly recommend you review before proceeding with this one.
 
 ## How to compute K-Nearest Neighbors - Brute Force Approach
 
--- a/notebooks/nearest_neighbors.livemd
+++ b/notebooks/nearest_neighbors.livemd
@@ -1,0 +1,255 @@
+<!-- livebook:{"file_entries":[{"name":"knn_gyms.png","type":"attachment"}]} -->
+
+# Nearest Neighbors
+
+```elixir
+Mix.install([
+  {:scholar, path: "/home/mateusz/Dashbit/scholar", override: true},
+  {:explorer, "~> 0.8.1"},
+  {:exla, "~> 0.7.2"},
+  {:nx, "~> 0.7.2"},
+  {:req, "~> 0.4.14"},
+  {:kino_vega_lite, "~> 0.1.11"},
+  {:kino, "~> 0.12.3"},
+  {:kino_explorer, "~> 0.1.18"}
+])
+```
+
+## Setup
+
+```elixir
+require Explorer.DataFrame, as: DF
+require Explorer.Series, as: S
+Nx.global_default_backend(EXLA.Backend)
+```
+
+## Introduction to Nearest Neighbors
+
+In this notebook, we introduce the concept of "**nearest neighbors**." We start by defining what nearest neighbors are and why they are useful in real-life scenarios. Next, we present various methods to compute them, highlighting the advantages and disadvantages of each approach. To make this notebook easier to follow, we will demonstrate these algorithms on a few examples and analyze the results.
+
+## Definition of K-Nearest Neighbors
+
+Let's first define what exactly "nearest neighbors" are. Consider a set of points ${p_1, p_2, \ldots, p_N}$, where each point $p_i$ is represented as a vector of size $M$, with real-number coordinates: $p_i = (p_{i1}, p_{i2}, \ldots, p_{iM})$. We say that the dimension of the space is $M$. For convenience, instead of listing the set of points as ${p_1, p_2, \ldots, p_N}$, we can treat it as a matrix $P$ of size $N \times M$, where $P_{ij}$ represents the $j$th coordinate of the $i$th point.
+
+A metric between two points $a$ and $b$ is defined as a function d that takes two points and returns a non-negative real number, subject to the following conditions for arbitrary points $a$, $b$, and $c$:
+
+* $d(a,b) = 0 \iff a = b$ (identity)
+* $d(a,b) = d(b, a)$ (symmetry)
+* $d(a,b) \leq d(a, c) + d(b, c)$ (triangle inequality)
+
+Examples of such metrics include the Euclidean metric and the Manhattan metric.
+
+Given a set of points $P$, a point $q$, and a metric $d$, the $k$ nearest neighbors from $P$ to point $q$ are defined as a set $S \subseteq P$, with $\lvert S \rvert = k$, where the distance $d$ between the points in $S$ and $q$ is minimized. In simpler terms, $S$ is a set of $k$ points from $P$ with the smallest distance to point $q$.
+
+It's especially interesting to consider the case when point $q$ is part of the set $P$. In the remainder of the notebook, we will assume that $q$ is indeed in $P$, and if this is not the case, we will note it explicitly.
+
+## Why Do We Want to Use K-Nearest Neighbors
+
+Now, after the formal section with definitions, let's move on to the more practical part: why would we even want to calculate something like K-nearest Neighbors? Consider the following example. Let's say you've just moved to a new city, and you're here for the first time. Since you're an active person, you'd like to find a nearby gym with good facilities. What would you do? You'd probably start by searching for gyms on online maps. The search results might look something like this.
+
+<!-- livebook:{"break_markdown":true} -->
+
+![](files/knn_gyms.png)
+
+<!-- livebook:{"break_markdown":true} -->
+
+Now you can check out the gyms and eventually decide which one will be your regular spot. What did the search engine do? It basically calculated the nearest gyms (nearest neighbors) from your current location, which is the essence of finding nearest neighbors.
+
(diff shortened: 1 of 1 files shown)


notebooks/nearest_neighbors.livemd
```diff
@@ -0,0 +1,255 @@
+<!-- livebook:{"file_entries":[{"name":"knn_gyms.png","type":"attachment"}]} -->
+
+# Nearest Neighbors
+
+```elixir
+Mix.install([
+  {:scholar, path: "/home/mateusz/Dashbit/scholar", override: true},
```
josevalim, line 7:
```suggestion
  {:scholar, path: "../..", override: true},
```

notebooks/nearest_neighbors.livemd
```diff
@@ -0,0 +1,255 @@
...
+{x, y}
+```
+
+Once the data is preprocessed, we can move on to the KDTree model. First, we need to set up the model using `Scholar.Neighbors.KDTree.fit`. This method initializes the tree structure for further processing.
+
+```elixir
+num_neighbors = 6
+
+knn_model =
+  EXLA.jit_apply(&Scholar.Neighbors.KDTree.fit(&1, k: num_neighbors, metric: :euclidean), [x])
```
josevalim, line 144: I tend to prefer jit, a bit less confusing. :)

```suggestion
  EXLA.jit(&Scholar.Neighbors.KDTree.fit(&1, k: num_neighbors, metric: :euclidean).(x)
```

Can we change it accordingly?

notebooks/nearest_neighbors.livemd
```diff
@@ -0,0 +1,255 @@
...
+raw_data =
+  Req.get!(digits_url).body
+```
+
+Next, we load the data into a dataframe.
+
+```elixir
+{:ok, [_, {_, data}]} = :zip.extract(raw_data, [:memory])
+data = String.replace(data, ", ", ",")
+df_data = DF.load_csv!(data, delimiter: ",")
```
josevalim, line 227: Could we do this instead, without the replace? Or not really?

```suggestion
df_data = DF.load_csv!(data, delimiter: ", ")
```

josevalim:
Looks great! Should we also talk about NNDescent?

notebooks/nearest_neighbors.livemd
msluszniak: Ups, I'll correct this

notebooks/nearest_neighbors.livemd
msluszniak: Ok, sure

notebooks/nearest_neighbors.livemd
```diff
@@ -0,0 +1,255 @@
+<!-- livebook:{"file_entries":[{"name":"knn_gyms.png","type":"attachment"}]} -->
+
+# Nearest Neighbors
```
josevalim, line 3: Since we already have a notebook on kNN, what if we call this "Approximate nearest neighbours"?

notebooks/nearest_neighbors.livemd
```diff
@@ -0,0 +1,255 @@
...
+
+One type of [recommendation system](https://en.wikipedia.org/wiki/Recommender_system) relies on collaborative filtering: it recommends songs based on what other users with similar musical tastes listen to. Another approach is to treat songs as points, like in our earlier definition, and then compute the closest songs to your favorites. This is another example of calculating nearest neighbors. While it might not be immediately obvious how to represent songs as points and measure the "distance" between them, it turns out that it can be done.
+
+Another use of nearest neighbors is in building classifiers and regressors. This approach was covered in a [previous notebook](https://github.com/elixir-nx/scholar/blob/main/notebooks/k_nearest_neighbors.livemd), which we strongly recommend you review before proceeding with this one.
```
josevalim, line 62:
```suggestion
Another use of nearest neighbors is in building classifiers and regressors. This approach was covered in a [previous notebook](k_nearest_neighbors.livemd), which we strongly recommend you review before proceeding with this one.
```

notebooks/nearest_neighbors.livemd
```diff
@@ -0,0 +1,255 @@
...
+
+Given this, for large values of $N$, we need to consider other approaches. However, the brute-force method can still be useful in certain situations. If $N$ is small, the distance matrix $D$ can be computed in a reasonable amount of time. Additionally, if $k \approx N$, this approach is optimal since you always need to compute $O(Nk)$, which leads to $O(N^2)$ operations. Computing the distance matrix is optimized in many frameworks, giving it an advantage over other methods when $k$ is on the order of magnitude of $N$.
+
+In Scholar, the brute-force approach is implemented [here](https://github.com/elixir-nx/scholar/blob/main/lib/scholar/neighbors/k_nearest_neighbors.ex). In a previous section, we mentioned a notebook that uses this approach to calculate $k$-nearest neighbors.
```
josevalim, line 72:
```suggestion
In Scholar, the brute-force approach is implemented with `Scholar.Neighbors.KNearestNeighbors`. In a previous section, we mentioned a notebook that uses this approach to calculate $k$-nearest neighbors.
```

Or should we link to the new `BruteKNN` instead?

notebooks/nearest_neighbors.livemd
josevalim: Hrm, we also talk about KDTree here, which is not approximate. But KDTree is `k=1` right? So I think this notebook is good as is. :)

josevalim, line 3:
```suggestion
# Nearest neighbors
```

notebooks/nearest_neighbors.livemd
msluszniak: The version without `replace` won't work. I spent a lot of time figuring out what was wrong because initially dataframe looked perfectly fine but when I passed it to `Nx.stack` I got an error "There is no tensor passed to `Nx.concatenate`". It turned out that all columns were loaded as stings with leading space and the dataframe couldn't correctly convert to tensor.

notebooks/nearest_neighbors.livemd
```diff
@@ -0,0 +1,255 @@
+<!-- livebook:{"file_entries":[{"name":"knn_gyms.png","type":"attachment"}]} -->
```
josevalim, line 1: Oh, you should also run the notebook with persist outputs enabled, so we have the outputs in the guides. :)

notebooks/nearest_neighbors.livemd
msluszniak: This notebook also explains KDTrees which aren't approximate NN. I know the current name somehow collides with the notebook about brute-force implementation. I'll think about a new name.

notebooks/nearest_neighbors.livemd
msluszniak: I think the proposed version is better :)

notebooks/nearest_neighbors.livemd
msluszniak: $K$ in KD stand for the number of dimensions of data. In the example $K = 7$.

notebooks/nearest_neighbors.livemd
```diff
@@ -0,0 +1,255 @@
...
+LargeVis is an approximated nearest neighbors algorithm that uses the concept of a Random Projection Forest. A Random Projection Forest consists of multiple trees. Each tree is built using a divide-and-conquer approach. The entire dataset is used to start the construction, and at each node, the data is projected onto a random hyperplane and split as follows: points with a projection smaller than or equal to the median go into the left subtree, while those with a projection greater than the median go into the right subtree. This process is then recursively applied to the left and right subtrees.
+
+This structure allows us to quickly create an approximated nearest neighbors graph, which can then be used to find the nearest neighbors efficiently. To better understand this, below is a figure illustrating the process of creating a tree in a random projection forest.
+
+| ![Random Projection Forest](http://www.math.umassd.edu/~dyan/rpKNN.png?attredirects=0) |
+| :------------------------------------------------------------------------------------: |
+
+Figure 3: A process of creating a tree in Random Projection Forest. Source: http://www.math.umassd.edu/~dyan/rpKNN.png?attredirects=0
```
krstopro, line 204: Not sure about this figure. The reason is that random projection trees in `Scholar` always split the data in half. Perhaps the one from https://cseweb.ucsd.edu/~dasgupta/papers/rptree-stoc.pdf or https://cseweb.ucsd.edu/~dasgupta/papers/exactnn-algo.pdf might be a better idea?

notebooks/nearest_neighbors.livemd
```diff
@@ -0,0 +1,255 @@
...
+
+<!-- livebook:{"break_markdown":true} -->
+
+Here's a brief explanation of LargeVis. First, we compute an approximated nearest neighbors graph using a Random Projection Forest. Then, we refine the graph by examining the neighbors of each point's neighbors for a fixed number of iterations. Now, let's use LargeVis in practice.
```
krstopro, line 208:
```suggestion
Here's a brief explanation of LargeVis. First, we compute the approximate nearest neighbors graph using a Random Projection Forest. Then, we refine the graph by examining the neighbors of each point's neighbors for a fixed number of iterations. Now, let's use LargeVis in practice.
```

notebooks/nearest_neighbors.livemd
```diff
@@ -0,0 +1,255 @@
...
+```elixir
+num_neighbors = 6
+{indices, _dist} = Scholar.Neighbors.LargeVis.fit(x, num_neighbors: num_neighbors)
+```
+
+Finally, we compute the nearest neighbors and sort them in descending order based on the sum of shares.
+
+```elixir
+shares_of_neighbors = Nx.take(y, indices) |> Nx.squeeze(axes: [-1]) |> Nx.sum(axes: [-1])
+ord = Nx.argsort(shares_of_neighbors, direction: :desc)
```
krstopro, line 251: Is sort needed here? I think they are already sorted in descending order with respect to the distance:
https://github.com/elixir-nx/scholar/blob/main/lib/scholar/neighbors/utils.ex#L38

krstopro (commented):
Several comments.
I will probably have another look later even though it looks great so far.

notebooks/nearest_neighbors.livemd
msluszniak: Sort here is by the sum of shares

krstopro: Oh I see, sorry. :)

notebooks/nearest_neighbors.livemd
msluszniak: Are there any photos of these on the internet that I could link to?

msluszniak: I will include the image from the first paper that shows how splitting looks like in the case of kdtree and rpf, but I will leave the original image and mention that actual splitting in Scholar is implemented differently.

krstopro: That might be a better idea.

josevalim (approved):

krstopro (approved):

notebooks/nearest_neighbors.livemd
krstopro: I don't think it's that important, so feel free to leave the existing figure if you want. :)

PR #263, opened 2024-05-12
K-NN Classifier

krstopro:
There are several dilemmas I had while implementing this:
- ~~How to provide k-NN algorithm and algorithm-specific options? The way it's currently done is either as `:algorithm_name` or `{:algorithm_name, algorithm_opts}` (someone correct me if I'm wrong, but the latter should be idiomatic Elixir way of specifying both the module and options to be passed to the initialization function). Another way would be having a separate option for `algorithm_name` and algorithm-specific options `algorithm_opts`.~~
- ~~Literally every k-NN algorithm in Scholar takes `num_neighbors` as an option. I would however prefer passing it as separate option to `KNNClassifier` instead of nesting it inside algorithm-specific options. What I mean is doing `Scholar.Neighbors.KNNClassifier.fit(x, y, num_neighbors: 3, num_classes: 2)` instead of `Scholar.Neighbors.KNNClassifier.fit(x, y, {:brute, [num_neighbors: 3]}, num_classes: 2)`. Similarly for the metric option. Also, currently it is possible to do `Scholar.Neighbors.KNNClassifier.fit(x, y, {:brute, [num_neighbors: 5]}, num_neighbors: 3, num_classes: 2)` and `num_neighbors: 5` will override `num_neighbors: 3`. Perhaps an error should be raised to prevent this?~~

TODO:
- ~~I think `KDTree.predict/2` should be updated to return `{neighbors, distances}` (currently it returns just `neighbors`; an unit-test is failing because of this). I might need help with this one.~~
- ~~Implement `KNNClassifier.predict_proba/2`.~~
- Add more metrics, e.g. `:euclidean`.
- Not sure, but `Scholar.Options.metric` might also need to be edited. Alternative is removing it from k-NN modules and specifying metrics as atoms in the docs.
- ~~Maybe few more unit-tests.~~

Last, I am sorry it took me slightly longer to implement this than I said. I suffered a horrible bike crash this week. I am fine, but still recovering, both physically and mentally. Started implementing `KNNRegressor` in parallel with this one - shouldn't take long.

### DIFF
--- a/lib/scholar/neighbors/kd_tree.ex
+++ b/lib/scholar/neighbors/kd_tree.ex
@@ -271,10 +271,6 @@ defmodule Scholar.Neighbors.KDTree do
     predict_n(tree, data, NimbleOptions.validate!(opts, @predict_schema))
   end
 
-  defnp predict_n(tree, data, opts) do
-    query_points(data, tree, opts)
-  end
-
   defnp sort_by_distances(distances, point_indices) do
     indices = Nx.argsort(distances)
     {Nx.take(distances, indices), Nx.take(point_indices, indices)}
@@ -315,7 +311,7 @@ defmodule Scholar.Neighbors.KDTree do
     end
   end
 
-  defnp query_points(point, tree, opts) do
+  defnp predict_n(tree, point, opts) do
     k = opts[:k]
     node = Nx.as_type(root(), :s64)
 
@@ -458,12 +454,12 @@ defmodule Scholar.Neighbors.KDTree do
 
                 # Should be not reachable
                 true ->
-                  {node, i, visited, nearest_neighbors, distances, -1, down, up}
+                  {node, i, visited, nearest_neighbors, distances, Nx.u8(-1), down, up}
               end
 
             # Should be not reachable
             true ->
-              {node, i, visited, nearest_neighbors, distances, -1, down, up}
+              {node, i, visited, nearest_neighbors, distances, Nx.u8(-1), down, up}
           end
 
         {nearest_neighbors, {node, data, indices, point, distances, visited, i, mode, down, up}}
--- a/lib/scholar/neighbors/kd_tree.ex
+++ b/lib/scholar/neighbors/kd_tree.ex
@@ -331,9 +331,7 @@ defmodule Scholar.Neighbors.KDTree do
     indices = tree.indices |> Nx.as_type(:s64)
     data = tree.data
 
-    down = Nx.u8(0)
-    up = Nx.u8(1)
-    mode = down
+    mode = down()
     i = Nx.s64(0)
 
     [nearest_neighbors, node, distances, visited, i, mode, point] =
@@ -348,28 +346,27 @@ defmodule Scholar.Neighbors.KDTree do
       ])
 
     {nearest_neighbors, _} =
-      while {nearest_neighbors,
-             {node, data, indices, point, distances, visited, i, mode, down, up}},
+      while {nearest_neighbors, {node, data, indices, point, distances, visited, i, mode}},
             node != -1 and i >= 0 do
         coord_indicator = rem(i, dims)
 
-        {node, i, visited, nearest_neighbors, distances, mode, down, up} =
+        {node, i, visited, nearest_neighbors, distances, mode} =
           cond do
             node >= size ->
-              {parent(node), i - 1, visited, nearest_neighbors, distances, up, down, up}
+              {parent(node), i - 1, visited, nearest_neighbors, distances, up()}
 
-            mode == down and
+            mode == down() and
                 point[[coord_indicator]] < data[[indices[node], coord_indicator]] ->
-              {left_child(node), i + 1, visited, nearest_neighbors, distances, down, down, up}
+              {left_child(node), i + 1, visited, nearest_neighbors, distances, down()}
 
-            mode == down and
+            mode == down() and
                 point[[coord_indicator]] >= data[[indices[node], coord_indicator]] ->
-              {right_child(node), i + 1, visited, nearest_neighbors, distances, down, down, up}
+              {right_child(node), i + 1, visited, nearest_neighbors, distances, down()}
 
-            mode == up ->
+            mode == up() ->
               cond do
                 visited[indices[node]] ->
-                  {parent(node), i - 1, visited, nearest_neighbors, distances, up, down, up}
+                  {parent(node), i - 1, visited, nearest_neighbors, distances, up()}
 
                 (left_child(node) >= size and right_child(node) >= size) or
                   (left_child(node) < size and visited[indices[left_child(node)]] and
@@ -390,7 +387,7 @@ defmodule Scholar.Neighbors.KDTree do
                       opts
                     )
 
-                  {parent(node), i - 1, visited, nearest_neighbors, distances, up, down, up}
+                  {parent(node), i - 1, visited, nearest_neighbors, distances, up()}
 
                 left_child(node) < size and visited[indices[left_child(node)]] and
                   right_child(node) < size and
@@ -416,10 +413,9 @@ defmodule Scholar.Neighbors.KDTree do
                        ) <
                          distances
                      ) do
-                    {right_child(node), i + 1, visited, nearest_neighbors, distances, down, down,
-                     up}
+                    {right_child(node), i + 1, visited, nearest_neighbors, distances, down()}
                   else
-                    {parent(node), i - 1, visited, nearest_neighbors, distances, up, down, up}
+                    {parent(node), i - 1, visited, nearest_neighbors, distances, up()}
                   end
 
                 ((right_child(node) < size and visited[indices[right_child(node)]]) or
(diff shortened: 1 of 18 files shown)


lib/scholar/neighbors/knn_classifier.ex
```diff
@@ -0,0 +1,245 @@
...
+
+      iex> x_train = Nx.tensor([[1, 2], [2, 3], [3, 4], [4, 5], [5, 6]])
+      iex> y_train = Nx.tensor([0, 0, 0, 1, 1])
+      iex> model = Scholar.Neighbors.KNNClassifier.fit(x_train, y_train, num_neighbors: 3, num_classes: 2)
+      iex> x_test = Nx.tensor([[1, 3], [4, 2], [3, 6]])
+      iex> Scholar.Neighbors.KNNClassifier.predict(model, x_test)
+      Nx.tensor([0, 0, 1])
+  """
+  deftransform predict(model, x) do
+    knn_module = model.algorithm.__struct__
```
krstopro, line 160: This doesn't work inside `defn`. Not sure how to resolve it.

josevalim:
> How to provide k-NN algorithm and algorithm-specific options? 

One option is to pass all options altogether. Then KNNClassifier "splits" (via Keyword.split) the options it uses and passes all other options to the underlying algorithm, which will also use NimbleOptions to validate and raise in case of unknown/bad options.

FWIW, I'd also call it simply `:algorithm`.

lib/scholar/neighbors/knn_classifier.ex
josevalim: What is the issue with this code? It looks fine as is to me.

josevalim:
> Not sure, but Scholar.Options.metric might also need to be edited. Alternative is removing it from k-NN modules and specifying metrics as atoms in the docs.

What do you want to edit? We should probably make it consistent and  make it always return a two element function. Is this what you want?

josevalim:
> I think KDTree.predict/2 should be updated to return {neighbors, distances} (currently it returns just neighbors; an unit-test is failing because of this). I might need help with this 

@msluszniak could you please give a hand on this one? 🙌 

> Last, I am sorry it took me slightly longer to implement this than I said. I suffered a horrible bike crash this week. I am fine, but still recovering, both physically and mentally. 

Sorry to hear but also glad to you are fine! Have a speedy recovery!

krstopro:
> > Not sure, but Scholar.Options.metric might also need to be edited. Alternative is removing it from k-NN modules and specifying metrics as atoms in the docs.
> 
> What do you want to edit? We should probably make it consistent and make it always return a two element function. Is this what you want?

I am not sure if we want to specify the metric option as
https://github.com/elixir-nx/scholar/blob/e0e92d07ffd690c7a11b87daf66d9e8642f3f25a/lib/scholar/neighbors/kd_tree.ex#L35
or simply as `type: {:in, [:minkowski, :cosine]}`. Especially if we want to add more metrics, it might become an issue which of those are supported by different k-NN algorithms. Another thing is, as you say, whether normalization should be performed inside `Scholar.Options.metric` or inside the modules where metric can be specified (as mentioned [here](https://github.com/elixir-nx/scholar/pull/257#discussion_r1564119231)).

krstopro:
> > How to provide k-NN algorithm and algorithm-specific options?
> 
> One option is to pass all options altogether. Then KNNClassifier "splits" (via Keyword.split) the options it uses and passes all other options to the underlying algorithm, which will also use NimbleOptions to validate and raise in case of unknown/bad options.
> 
> FWIW, I'd also call it simply `:algorithm`.

Yeah, this is probably the way. :)

lib/scholar/neighbors/knn_classifier.ex
krstopro: I wanted to put `knn_module.predict(model.algorithm, x)` inside `defn` (e.g. `predict_n`), but the line above won't work (invalid numerical expression error as far as I remember).

josevalim: I see. In any case, this code looks fine to me, I can play with it later on and see if I can improve the defn error.

josevalim:
> I am not sure if we want to specify the metric option as

Let's open up a separate issue to normalize how metric is handled. My suggestion would be to use ` {:custom, Scholar.Options, :metric, []}` everywhere and, if you cannot handle any metric, then you explicitly opt-in to the only ones you can.

msluszniak:
> 
> @msluszniak could you please give a hand on this one? 🙌

Sure, I'll work on that. 

>
> Last, I am sorry it took me slightly longer to implement this than I said. I suffered a horrible bike crash this week. I am fine, but still recovering, both physically and mentally.

I'm sorry, I wish you a quick recovery.

lib/scholar/neighbors/kd_tree.ex
```diff
@@ -311,8 +333,8 @@ defmodule Scholar.Neighbors.KDTree do
     end
   end
 
-  defnp predict_n(tree, point, opts) do
-    k = opts[:k]
+  defnp predict_n(tree, point) do
+    k = tree.num_neighbors
```
msluszniak, line 337: I we now pass `num_neighbors` in fit, I think we may add a note that there is no need to compute all KDTree from scratch for a different number of nearest neighbors.

lib/scholar/neighbors/knn_classifier.ex
krstopro: Apparently, it's not this line that is causing the problem. The following line is giving me `invalid numerical expression` error
```
{neighbors, distances} = algorithm_module.predict(model.algorithm, x)
```
I am not sure why, but it probably has to do something with how `defn` works. Could you maybe have a look?

josevalim: Ok, it makes sense it doesn't work, because `defn` then cannot guarantee `algorithm_module.predict` is a `defn` at compile-time. We can likely relax it though but that shouldn't block it here. My suggestion is to make `predict` a `defn predict` and move only the invocation to transform:

```
defn predict(model, x) do
  {neighbors, distances} = predict_apply(model.algorithm, x)
  ...
end

deftransformp predict_apply(algo, x) do
  algo.__struct__.predict(algo, x)
end
```

We can probably do something similar for calling `fit` on init.

josevalim: FWIW, it is not strictly necessary for calling `fit`, because in `fit` the call to `algo.fit` is literally the only `defn` that we invoke, but it probably makes sense to do it as a best practice anyway.

msluszniak:
> I think KDTree.predict/2 should be updated to return {neighbors, distances} (currently it returns just neighbors; an unit-test is failing because of this). I might need help with this one.

I've sent a PR with improvement

krstopro:
> > I think KDTree.predict/2 should be updated to return {neighbors, distances} (currently it returns just neighbors; an unit-test is failing because of this). I might need help with this one.
> 
> I've sent a PR with improvement

Very quick, thanks. :)

lib/scholar/neighbors/knn_classifier.ex
krstopro: > FWIW, it is not strictly necessary for calling `fit`, because in `fit` the call to `algo.fit` is literally the only `defn` that we invoke, but it probably makes sense to do it as a best practice anyway.

I didn't use `defn` for `fit` exactly for the reason you wrote above. Are you suggesting to do it anyway? I.e. to move `algorithm_module.fit(x, algorithm_opts)` inside `defn`?

josevalim: It is up to you. I am fine either way. :)

krstopro: I think it just makes the code slightly harder to read, so I would avoid doing it for `fit`. :)

lib/scholar/neighbors/kd_tree.ex
krstopro: Perhaps, but then we do the same in `BruteKNN` and `RandomProjectionForest`. They all now take `num_neighbors` as an option to `fit`.

lib/scholar/neighbors/knn_classifier.ex
```diff
@@ -0,0 +1,287 @@
...
+      iex> Scholar.Neighbors.KNNClassifier.predict_proba(model, x)
+      Nx.tensor(
+        [
+          [1.0, 0.0],
+          [1.0, 0.0],
+          [1.0, 0.0]
+        ]
+      )
+  """
+  defn predict_proba(model, x) do
```
josevalim, line 182: Let's rename it to `predict_probability`, because that's what we call these functions everywhere else! However, if they are incorrect and we can't figure out why, we can remove this for now and add it in future PRs. :) Your call!

krstopro: I would rather investigate it now. I don't expect it to take long, but then, you never know. :)

lib/scholar/neighbors/knn_classifier.ex
```diff
@@ -0,0 +1,287 @@
...
+
+    weights =
+      case model.weights do
+        :uniform -> Nx.broadcast(1.0, neighbors)
+        :distance -> check_weights(distances)
+      end
+
+    indices =
+      Nx.stack(
+        [Nx.iota(Nx.shape(labels_pred), axis: 0), Nx.take(model.labels, labels_pred)],
```
josevalim, line 197: I think this is what we want?

```suggestion
        [Nx.iota(Nx.shape(labels_pred), axis: 0), labels_pred],
```

krstopro: Yes, I think so. Let me rename `labels_pred` to `neighbor_labels`, I think it is more suitable name.

lib/scholar/neighbors/knn_classifier.ex
```diff
@@ -0,0 +1,287 @@
...
+        [
+          [1.0, 0.0],
+          [1.0, 0.0],
+          [0.3333333432674408, 0.6666666865348816]
+        ]
+      )
+  """
+  defn predict_probability(model, x) do
+    num_samples = Nx.axis_size(x, 0)
+    type = Nx.Type.merge(to_float_type(x), {:f, 32})
```
krstopro, line 184: I don't see why this is necessary here.
@msluszniak Do you maybe remember?

josevalim: Actually, I think this is a legacy from before we had to_flaot_type. You can remove it.

```suggestion
    type = to_float_type(x)
```

krstopro: Yes, the reason I asked is because I don't see why merging is necessary here. I know there are some issues with `:f64` and `:f16`, but here only elementary operations are used. I cannot spot a problem here.

krstopro: > Actually, I think this is a legacy from before we had to_flaot_type. You can remove it.

Quite possible, I think there was a similar issue in `MultinomialNaiveBayes`.

josevalim (approved):

msluszniak (approved):

PR #264, opened 2024-05-13
Output distances in kdtree

msluszniak:


josevalim (approved):

lib/scholar/neighbors/kd_tree.ex
```diff
@@ -266,26 +306,28 @@ defmodule Scholar.Neighbors.KDTree do
...
+        [
+          [2.0, 2.0, 6.0],
+          [7.0, 7.0, 7.0],
+          [3.0, 3.0, 5.0]
+        ]
+      >
   """
-  deftransform predict(tree, data, opts \\ []) do
-    predict_n(tree, data, NimbleOptions.validate!(opts, @predict_schema))
+  deftransform predict(tree, data) do
```
krstopro, line 319: `deftransform` is redundant here, isn't it? I would add data validation inside this function though.

krstopro: Oh, my bad, I now see that the options are validated. I guess the options to `predict` should be removed.

msluszniak: Yeah, you're right. As we now validate options in `fit` we can make it `defn`

krstopro: I would validate `data` in predict as well, maybe rename it to `query` (up to you).
Just the usual stuff: that rank is equal to 2 and that the 2nd dimension matches the one of the data used to fit the model.

Issue #249, opened 2024-04-06
Unification of handling weights passed as arguments to functions 


msluszniak:
As mentioned [here](https://github.com/elixir-nx/scholar/pull/247#issuecomment-2038130186) we want to unify the way we pass the weights in Scholar to the algorithms. Currently, there are a few ways that we do it. For example, we pass them as tensors or as a list via `opts`.

JoaquinIglesiasTurina:
I would like to take a stab at this one, particularly regarding linear methods.
I think this would also be a great opportunity to DRY up some of the duplication of private functions that's spreading over there.
So far, the better approach I've found is to move all these functions to a helper module, make them public,
but make sure to set `@moduledoc false` and `@doc false`. My proof of concept is [here](https://github.com/JoaquinIglesiasTurina/scholar/blob/main/lib/scholar/linear/linear_helpers.ex).
Is this the best approach to handle this? Or should I take an alternative approach?

msluszniak:
Yes, this looks great! You can send a PR with your improvements :)

PR #267, opened 2024-05-15
Unify weight handling and refactor linear models' helper functions

JoaquinIglesiasTurina:
Implements the suggestion from [here](https://github.com/elixir-nx/scholar/issues/249).

I've had to revert the change suggested by @krstopro [here](https://github.com/elixir-nx/scholar/pull/247#discussion_r1573718068), as that change made some tests from `LinearRegression` and `RidgeRegression` fail.

When it comes to ` LinearRegression`  I believe those failures are related to the shape of the target data: 
- The tests yield failures when the data is 2 dimensional, [like a column vector](https://github.com/elixir-nx/scholar/blob/322687a04ea0f701efdf42fd6360627a0ed30518/test/scholar/linear/linear_regression_test.exs#L132)
- Flattening those data yield a passing test.

`RidgeRegression` tests fail with multi-output target data. These cannot be flattened.

### DIFF
--- a/lib/scholar/linear/linear_helpers.ex
+++ b/lib/scholar/linear/linear_helpers.ex
@@ -43,14 +43,20 @@ defmodule Scholar.Linear.LinearHelpers do
   # targets by sqrt(sample_weight).
   @doc false
   defn rescale(x, y, sample_weights) do
-    case Nx.shape(sample_weights) do
-      {} = scalar ->
-        scalar = Nx.sqrt(scalar)
-        {scalar * x, scalar * y}
+    factor = Nx.sqrt(sample_weights)
 
-      _ ->
-        scale = sample_weights |> Nx.sqrt() |> Nx.make_diagonal()
-        {Nx.dot(scale, x), Nx.dot(scale, y)}
-    end
+    x_scaled =
+      case Nx.shape(factor) do
+        {} -> factor * x
+        _ -> x * Nx.new_axis(factor, -1)
+      end
+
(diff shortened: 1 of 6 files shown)


lib/scholar/linear/linear_helpers.ex
```diff
@@ -0,0 +1,56 @@
...
+    case Nx.shape(sample_weights) do
+      {} = scalar ->
+        scalar = Nx.sqrt(scalar)
+        {scalar * x, scalar * y}
+
+      _ ->
+        scale = sample_weights |> Nx.sqrt() |> Nx.make_diagonal()
+        {Nx.dot(scale, x), Nx.dot(scale, y)}
+    end
+  end
```
msluszniak, line 55: Based on this https://github.com/scikit-learn/scikit-learn/blob/ 8721245511de2f225ff5f9aa5f5fadce663cd4a3/sklearn/linear_model/_base.py#L203
```suggestion
  defn rescale(x, y, sample_weights) do
    factor = Nx.sqrt(sample_weights)

    x_scaled =
      case Nx.shape(factor) do
        {} -> factor * x
        _ -> x * Nx.new_axis(factor, -1)
      end

    y_scaled =
      case Nx.rank(y) do
        1 -> factor * y
        _ -> y * Nx.new_axis(factor, -1)
      end

    {x_scaled, y_scaled}
  end
```

msluszniak (approved):
I think it's now good to go ;)

PR #268, opened 2024-05-15
K-NN Regressor

### LINKED ISSUE #255: Refactor Scholar.Neighbors

krstopro:
Implements k-NN regressor model and removes previous implementation of k-NN.
As usual there are few things I am not sure about:
- There is a lot of code repeated from `KNNClassifier` (e.g.` deftransformp compute_knn`). Was thinking pulling it out of both modules and putting it somewhere else, but that might make the code harder to understand.
- Labels `y` can be both rank-1 and rank-2. I think this is slightly different than usual, where tensors that can contain more than one feature are always rank-2. I would suggest allowing `y` to be only of rank 2.

Closes #255.


josevalim (approved):

krstopro:
> * Labels `y` can be both rank-1 and rank-2. I think this is slightly different than usual, where tensors that can contain more than one feature are always rank-2. I would suggest allowing `y` to be only of rank 2.

I would prefer that we all agree on this before merging this pull request.

josevalim:
I don't have an opinion. But it is always easier to relax the constraints later (i.e. allow only rank 2 for now and potentially also rank 1 later). @msluszniak do you have any thoughts?

msluszniak:
Yes, I agree that for now we can accept only 2-dim labels

msluszniak (approved):

PR #274, opened 2024-06-01
use Nx.BinaryBackend for cv notebook

santiago-imelio:
Removes setup cell that sets EXLA as default backend and compiler. This is not supported for EXGBoost at the time.


josevalim:
@santiago-imelio we should also remove `exla` from Mix.install and perhaps we could update the `nx` dependency too?

josevalim:
If we can remove the override: true, that would be even better!

josevalim (approved):

josevalim:

Issue #272, opened 2024-05-30
Add manifold version of LargeVis


josevalim:
We already have the building blocks for ANN. @krstopro, how much work do you believe it is to also have it for manifold/dimensionality reduction?

krstopro:
Probably not much, though I am worried about memory cost of current implementations. `LargeVis` dimensionality reduction uses large `num_neighbors` (150 in the paper) and `RandomProjectionForest.predict/2` takes $O(QTK)$ memory for number of queries $Q$, number of trees $T$, and number of neighbors $K$. In any case, leave it to me.

By the way, thanks for the shout out on the blog. :)

krstopro:
Apparently, this is much harder than I thought. One of the issues is sampling the negative edges of the k-NN graph, i.e. edges that are not in the graph. To do this, [the authors allocate and iterate over a rather large table](https://github.com/lferry007/LargeVis/blob/feb8121e8eb9652477f7f564903d189ee663796f/Linux/LargeVis.cpp#L544-L567). I don't like this approach and I don't think I can come up with something efficiently implementable using `Nx`.

josevalim:
Why do you think it would be hard to implement it? The parr that seems complex is the head traversal, but if we can assume a max depth (not sure if that’s the case), it should be parallelizable.

krstopro:
It's not really about head traversal, more about allocating a table of size `1e8`. I'm still checking if this can be avoided.

krstopro:
The problem above might be avoidable, but there is another one that might not be. Namely, the method involves symmetrising the k-NN graph and this requires the computation of reverse k-NN graph which I don't think can be done efficiently with `Nx`.
<img width="468" alt="Screenshot 2024-06-14 at 19 03 13" src="https://github.com/elixir-nx/scholar/assets/7234787/9b7a8988-97e9-41a6-be00-34500617eb1d">

josevalim:
The symmetrized seems to be doable with a while loop over i,j?

krstopro:
[...]

PR #278, opened 2024-06-07
Manifold learning notebooks

msluszniak:
I decided to replace NNDescent in Trimap with LargeVis. In the original implementation of Trimap, authors used Annoy, which is more similar to LargeVis and it's faster. In Google's implementation using Jax, they use NNDescent, which is slower and eventually makes the whole implementation slower than the not-optimised, original one.
original implementation: https://github.com/eamid/trimap
Google implementation: https://github.com/google-research/google-research/tree/master/trimap


msluszniak:
@krstopro do you know there is any way to add support in LargeVis or precisely in RPF for data that is in test cases here and currently fails?

krstopro:
@msluszniak Yeah, I should fix that. The main reason I didn't do it in the first place was because I wouldn't recommend using random projection forest for those cases, i.e. when the sample size is less than two times leaf size.

I would recommend making k-NN algorithm configurable for Trimap. Similarly as it was done in `KNNClassifier` and `KNNRegressor`.

msluszniak:
Yeah, that's a good idea

notebooks/manifold_learning.livemd
```diff
@@ -0,0 +1,155 @@
...
+```elixir
+tensor_data = Nx.stack(df, axis: 1)
+```
+
+Since there is almost 1 million data points and they are sorted, we shuffle dataset and then use only the part of the dataset.
+
+We start with Trimap. It's a manifold learning algorithm that is based of nearest neighbors. It preserves the global structure of dataset, but it doesn't handle in a poroper way the local structure. Let's look what will be the result of the Trimap on mammoth dataset.
+
+```elixir
+{tensor_data, key} = EXLA.jit(&Nx.Random.shuffle(&1, &2)).(Nx.Random.key(42), tensor_data)
```
josevalim, line 61: We no longer need to JIT if we set the default compiler to EXLA, as in the top of the notebook :)

josevalim:
> I would recommend making k-NN algorithm configurable for Trimap. Similarly as it was done in KNNClassifier and KNNRegressor.

Good idea indeed, I have created an issue for it here: #280.

#272 is also related here (it could be another example to show in the notebook).

notebooks/manifold_learning.livemd
```diff
@@ -0,0 +1,155 @@
...
+  {:kino_vega_lite, "~> 0.1.11"},
+  {:kino, "~> 0.12.3"},
+  {:kino_explorer, "~> 0.1.18"},
+  {:tucan, "~> 0.3.1"}
+])
+```
+
+## Setup
+
+We will use `Explorer` in this notebook, so let's define aliase for its main module DataFrame:
```
krstopro, line 21:
```suggestion
We will use `Explorer` in this notebook, so let's define an alias for its main module DataFrame:
```

notebooks/manifold_learning.livemd
```diff
@@ -0,0 +1,155 @@
...
+And let's configure `EXLA` as our default backend (where our tensors are stored) and compiler (which compiles Scholar code) across the notebook and all branched sections:
+
+```elixir
+Nx.global_default_backend(EXLA.Backend)
+Nx.Defn.global_default_options(compiler: EXLA)
+```
+
+## Testing Manifold Learning Functionalities
+
+In this notebook, we test how manifold learning algorithms works, and what results we can get from them.
```
krstopro, line 36:
```suggestion
In this notebook we test how manifold learning algorithms works and what results we can get from them.
```
I would remove the commas here.

notebooks/manifold_learning.livemd
```diff
@@ -0,0 +1,155 @@
...
+
+Firstly, let's fetch the dataset that we experiment on. The data represents 3D coordinates of a mammoth. Below we include a figure of original dataset.
```
krstopro, line 38:
```suggestion
First, let's fetch the dataset that we experiment on. The data represents 3D coordinates of a mammoth. Below we include a figure of original dataset.
```

notebooks/manifold_learning.livemd
```diff
@@ -0,0 +1,155 @@
...
+  Tucan.scatter(coords, "x", "y", point_size: 1)
+])
+|> Tucan.set_size(300, 300)
+|> Tucan.set_title(
+  "Mammoth dataset with reduced dimensionality using Trimap",
+  offset: 25
+)
+```
+
+As we see, t-SNE have completely different results than trimap. This is because t-SNE has a completely different mathematical background of computation. Also t-SNE is slower algorithm, so it can't be used to such big datasets as trimap. However, t-SNE still preserve some features of mammoth like small tusks, feets, and corp. You can experiment with parameter *perplexity* which can significantly change the output of the algorithm.
```
krstopro, line 153:
```suggestion
As we see, t-SNE gives completely different results than trimap. This is because t-SNE has a completely different mathematical background of computation. Also t-SNE is slower algorithm, so it can't be used on such big datasets as trimap. However, t-SNE preserves some features of mammoth like small tusks, feets, and corp. You can experiment with parameter *perplexity* which can substantially change the output of the algorithm.
```

krstopro (commented):
Looks great, few comments on the writing from my side.

notebooks/manifold_learning.livemd
```diff
@@ -0,0 +1,155 @@
...
+
+Now, convert the dataframe into tensor, so we can manipulate the data using `Scholar`.
+
+```elixir
+tensor_data = Nx.stack(df, axis: 1)
+```
+
+Since there is almost 1 million data points and they are sorted, we shuffle dataset and then use only the part of the dataset.
+
+We start with Trimap. It's a manifold learning algorithm that is based of nearest neighbors. It preserves the global structure of dataset, but it doesn't handle in a poroper way the local structure. Let's look what will be the result of the Trimap on mammoth dataset.
```
josevalim, line 58: Let's start with a new section for Trimap. Perhaps even a branched section?

notebooks/manifold_learning.livemd
```diff
@@ -0,0 +1,155 @@
...
+|> Tucan.set_size(300, 300)
+|> Tucan.set_title(
+  "Mammoth data set with reduced dimensionality using trimap",
+  offset: 25
+)
+```
+
+These two plots are similiar but there are some important differences. Even if the second figure seems "prettier" it is less informative than the result of trimap. On the first figure, we can spot two tusks while one the second one they overlap and we see only one. Similarly, legs overlay on the first plot and one the second one they are spread and don't intersect with each other.
+
+Now, lets try different algorithm: t-SNE
```
josevalim, line 118: Let's start a  new section for this one too. Perhaps it could even be a branched section?

josevalim (approved):
Ship it once CI passes!

notebooks/manifold_learning.livemd
```diff
@@ -0,0 +1,149 @@
...
+
+## Trimap
+
+We start with Trimap. It's a manifold learning algorithm that is based of nearest neighbors. It preserves the global structure of dataset, but it doesn't handle in a poroper way the local structure. Let's look what will be the result of the Trimap on mammoth dataset.
+
+```elixir
+{tensor_data, key} = Nx.Random.shuffle(Nx.Random.key(42), tensor_data)
+
+trimap_res =
+  Scholar.Manifold.Trimap.embed(tensor_data[[0..10000, ..]],
```
msluszniak, line 68:
```suggestion
  Scholar.Manifold.Trimap.transform(tensor_data[[0..10000, ..]],
```

PR #281, opened 2024-06-10
Make nn algorithm configurable

### LINKED ISSUE #280: Make kNN configurable in Trimap

msluszniak:
Closes #280

### DIFF
--- a/lib/scholar/manifold/trimap.ex
+++ b/lib/scholar/manifold/trimap.ex
@@ -301,26 +301,37 @@ defmodule Scholar.Manifold.Trimap do
     num_extra = min(num_inliners + 50, num_points)
 
     neighbors =
-      case opts[:algorithm] do
-        :nndescent ->
-          nndescent =
-            Scholar.Neighbors.NNDescent.fit(inputs,
-              num_neighbors: num_extra,
-              tree_init?: false,
-              metric: opts[:metric],
-              tol: 1.0e-5
-            )
+      if Nx.axis_size(inputs, 0) <= 500 do
+        model =
+          Scholar.Neighbors.BruteKNN.fit(inputs,
+            num_neighbors: num_extra,
+            metric: opts[:metric]
+          )
 
-          nndescent.nearest_neighbors
+        {neighbors, _distances} = Scholar.Neighbors.BruteKNN.predict(model, inputs)
+        neighbors
+      else
+        case opts[:algorithm] do
+          :nndescent ->
+            nndescent =
+              Scholar.Neighbors.NNDescent.fit(inputs,
+                num_neighbors: num_extra,
+                tree_init?: false,
+                metric: opts[:metric],
+                tol: 1.0e-5
+              )
 
-        :large_vis ->
-          {neighbors, _distances} =
-            Scholar.Neighbors.LargeVis.fit(inputs,
-              num_neighbors: num_extra,
-              metric: opts[:metric]
-            )
+            nndescent.nearest_neighbors
 
-          neighbors
+          :large_vis ->
+            {neighbors, _distances} =
+              Scholar.Neighbors.LargeVis.fit(inputs,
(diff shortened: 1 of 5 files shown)


lib/scholar/manifold/trimap.ex
```diff
@@ -113,6 +113,16 @@ defmodule Scholar.Manifold.Trimap do
...
       """
+    ],
+    algorithm: [
+      type: {:in, [:nndescent, :large_vis]},
+      default: :large_vis,
+      doc: ~S"""
+      Algorithm used to compute the nearest neighbors. Possible values:
+      * `:nndescent` - Nearest Neighbors Descent. See `Scholar.Neighbors.NNDescent` for more details.
+
+      * `:large_vis` - LargeVis algorithm. See `Scholar.Neighbors.LaregVis` for more details.
```
josevalim, line 124:
```suggestion
      * `:large_vis` - LargeVis algorithm. See `Scholar.Neighbors.LargeVis` for more details.
```

josevalim (approved):

lib/scholar/manifold/trimap.ex
```diff
@@ -113,6 +113,16 @@ defmodule Scholar.Manifold.Trimap do
       doc: ~S"""
       Metric used to compute the distances.
       """
+    ],
+    algorithm: [
+      type: {:in, [:nndescent, :large_vis]},
```
krstopro, line 118: I would add at least `:brute`. Maybe custom k-NN graph construction algorithm to be passed as a module as well.

krstopro (commented):
Just a comment regarding allowed algorithms.

krstopro:
Function that computes the Trimap is called `embed` (not `fit` or `transform`). Does this deviate from the rest?

lib/scholar/manifold/trimap.ex
josevalim: Is there any benefit in using brute other than nndescent or large_vis?

krstopro: I'd say it's the best one to use for smaller datasets.

krstopro: Perhaps making the selection automatic depending on dataset size (sample size and number of features) would be ideal.

msluszniak: I will force use brute for `n < 100` and otherwise one of these two approximated algorithms, is it ok or do we want to add `:brute` anyway?

msluszniak:
> Function that computes the Trimap is called `embed` (not `fit` or `transform`). Does this deviate from the rest?

Yeah, I will change the name to `transform`.

lib/scholar/manifold/trimap.ex
krstopro: Maybe use brute-force for $N \times D^2 \leq T$ for some constant $T$ (e.g. $10^5$ or $10^6$ or so).
I would add `:brute` anyway.

Looking at it now, we might wanna change `predict` to `transform` in these algorithms as well.

msluszniak: for $N = 20000$ and $D = 2$, we rather won't use brute, generally, it's more effective to have this condition only on N I guess, I can increase this to 500

krstopro: Sounds good to me.

krstopro:
Looks good to me! If you want, I can have a more detailed look tomorrow.

Otherwise, feel free to merge it. I hope I am not annoying with these comments. ^_^

krstopro (approved):

msluszniak:
> Looks good to me! If you want, I can have a more detailed look tomorrow.
> 
> Otherwise, feel free to merge it. I hope I am not annoying with these comments. ^_^

Your comments is a valueable feedback for me, so don't stop 😀

josevalim (approved):

PR #284, opened 2024-06-21
Fix typo in manifold learning notebook

krstopro:
Just changed Trimap to t-SNE next to the plot of t-SNE embeddings.


josevalim:
@krstopro you probably executed the notebook again, which changed the results. Can you please change just the word directly in the source, outside of Livebook? Then we avoid polluting the codebase. :) (we should probably give a seed so it is deterministic, but we can do it later).

krstopro:
> @krstopro you probably executed the notebook again, which changed the results. Can you please change just the word directly in the source, outside of Livebook?

Sure, but should the result (in this case the plot) be updated as well?

krstopro:
It's this line that I changed, which also changes the plot
https://github.com/elixir-nx/scholar/blob/main/notebooks/manifold_learning.livemd?plain=1#L288

josevalim:
You should be able to change the plot directly too, or is that what you did and GH is not showing it? :)

krstopro:
> You should be able to change the plot directly too, or is that what you did and GH is not showing it? :)

I ran the entire notebook, just in case. :)

krstopro:
Closing this in favor of #285.

Issue #250, opened 2024-04-06
Unification of shape checks


msluszniak:
As mentioned [here](https://github.com/elixir-nx/scholar/pull/247#issuecomment-2038130186).

krstopro:
Is this really completed?

josevalim:
Errr, wrong issue.

josevalim:
At the same time, I am not sure if there is something to unify here? I think checking for shapes should be done per module, as they can provide better error messages within their context?

krstopro:
> At the same time, I am not sure if there is something to unify here? I think checking for shapes should be done per module, as they can provide better error messages within their context?

Could be the case. My idea was to have several functions for the cases that keep repeating, e.g. if `x` is of rank 2, if `x` and `y` agree on the first axis, etc.

josevalim:
The problem is that the error message ends up being generic: "expected tensors to have matching leading dimensions" while today we can say  "expected the data to have the same dimension as the query" etc. We actually had those helpers inside Nx itself at the beginning but we removed them because of that once we introduced `case`. :)

msluszniak:
Sure, that is really strange behaviour and not supposed to happen, so feel free to work on it. I'll appreciate that :)

JoaquinIglesiasTurina:
I've taken a further look at the issue with linear models. I think there is a decision to be made on how to handle the situation:
- Should the output of an `{n_samples, 1}` and `{n_samples}` shaped target be the same? Or should it be different?

Meaning, `RidgeRegression` returns `{1, n_samples}` shaped coefficients for `{n_samples, 1}` shaped targets. While for `{n_samples}` targets, it returns `{n_samples}` shaped coefficients. 
This is the behaviour of scikit's ordinary least squares. This behaviour can be achieved by fixing some `Nx.dot/2` to `Nx.dot/4`.

A different approach would be, shape-check the target, and make sure to flatten it prior to fitting the model. This would ensure that `{n_samples, 1}` and `{n_samples}` shaped targets yield equally shaped coefficients. 
This approach has some points of significance:
- Inconsistency with scikit's api (I don't know if this is a problem).
- It raises the question: How do we handle linear models with multioutput options? 
- It's potentially a breaking change of `RidgeRegression`'s api.

I mathematically favor the second option, when linear models are described mathematically, the target is a column vector and actually fitting a column vector and a vector should yield the same results.
[...]

PR #288, opened 2024-07-11
Fix/linear shapes

JoaquinIglesiasTurina:
Implemented the changes proposed [here](https://github.com/elixir-nx/scholar/issues/250#issuecomment-2184157918)
Please note the changes to `test/scholar/linear/bayesian_ridge_regression_test.exs` to fully understand the scope of the changes.

On the above comment, I defended models being equal when fitting single column vectors and one dimensional vectors.
However, I've just thought a drawback to this approach.
If the output of `.predict` is the same on both input cases, this means that prediction and target shapes may be different. This can cause some inconveniences.
I still think this is the better approach, but I want to make clear that the change is significant.

### DIFF
--- a/lib/scholar/linear/bayesian_ridge_regression.ex
+++ b/lib/scholar/linear/bayesian_ridge_regression.ex
@@ -428,7 +428,8 @@ defmodule Scholar.Linear.BayesianRidgeRegression do
 
   @doc """
   Makes predictions with the given `model` on input `x`.
-  Output predictions have shape {n_samples} when train target is shaped either {n_samples} or {n_samples, 1}.
+  
+  Output predictions have shape `{n_samples}` when train target is shaped either `{n_samples}` or `{n_samples, 1}`.
 
   ## Examples
 
--- a/lib/scholar/linear/bayesian_ridge_regression.ex
+++ b/lib/scholar/linear/bayesian_ridge_regression.ex
@@ -428,6 +428,8 @@ defmodule Scholar.Linear.BayesianRidgeRegression do
 
   @doc """
   Makes predictions with the given `model` on input `x`.
+  Output predictions have shape {n_samples} when train target is shaped either {n_samples} or {n_samples, 1}.
+
   ## Examples
 
       iex> x = Nx.tensor([[1], [2], [6], [8], [10]])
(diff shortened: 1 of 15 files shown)


msluszniak (approved):
I think it looks great :). Additionally, I would indicate somewhere information that the different input and target shapes may be different in specific case.

lib/scholar/linear/linear_helpers.ex
```diff
@@ -5,6 +5,37 @@ defmodule Scholar.Linear.LinearHelpers do
 
   @moduledoc false
 
+  @doc false
+  def valid_column_vector(y, n_samples) do
```
josevalim, line 9:
```suggestion
  defp valid_column_vector?(y, n_samples) do
```

lib/scholar/linear/linear_helpers.ex
```diff
@@ -5,6 +5,37 @@ defmodule Scholar.Linear.LinearHelpers do
...
+  end
+
+  @doc false
+  def validate_y_shape(y, n_samples, module_name) do
+    y = flatten_column_vector(y, n_samples)
+    is_valid_target? = Nx.rank(y) == 1
+
+    if not is_valid_target? do
+      message =
+        "#{module_name} expected y to have shape {n_samples}, got tensor with shape: #{inspect(Nx.shape(y))}"
```
josevalim, line 31:
```suggestion
        "#{inspect(module_name)} expected y to have shape {n_samples}, got tensor with shape: #{inspect(Nx.shape(y))}"
```

JoaquinIglesiasTurina:
> I think it looks great :). Additionally, I would indicate somewhere information that the different input and target shapes may be different in specific case.

Great, I'll update the docs

lib/scholar/linear/bayesian_ridge_regression.ex
```diff
@@ -428,6 +428,8 @@ defmodule Scholar.Linear.BayesianRidgeRegression do
 
   @doc """
   Makes predictions with the given `model` on input `x`.
+  Output predictions have shape {n_samples} when train target is shaped either {n_samples} or {n_samples, 1}.
```
josevalim, line 431: We should start a new paragraph, because the first paragraph should always be a short summary. And use backticks around code, can you please apply similar changes below? :)


```suggestion

  Output predictions have shape `{n_samples}` when train target
  is shaped either `{n_samples}` or `{n_samples, 1}`.
```

JoaquinIglesiasTurina: Sure thing. Thank you for your comments.

josevalim (approved):

PR #289, opened 2024-07-18
Incremental PCA

krstopro:
Adds Incremental PCA. As usual, there are several things I am not sure about:

- Basically the struct is almost the same as for regular PCA. These should probably be a single module, but it isn't that trivial to merge them (they are separate classes in `scikit-learn`).
- ~~There are two `fit/2` clauses: one that takes a tensor and another that takes a stream as a first argument. The one that takes a tensor is defined with `deftransform` while the one that takes a stream is defined with `def`. Is this fine?~~
- ~~Is it possible to add a popular dataset in `incremental_pca_test.exs` without adding `explorer` or `scidata` dependency to `mix.exs`?~~
- `transform/2` is the same as for regular PCA.


lib/scholar/shared.ex
```diff
@@ -89,4 +89,26 @@ defmodule Scholar.Shared do
...
+    batches =
+      tensor
+      |> Nx.slice_along_axis(0, num_batches * batch_size, axis: 0)
+      |> Nx.reshape({num_batches, batch_size, dim})
+
+    leftover =
+      if leftover_size > min_batch_size do
+        Nx.slice_along_axis(tensor, num_batches * batch_size, leftover_size, axis: 0)
+      else
+        nil
```
krstopro, line 109: I guess this won't work outside `defn`. Is it fine though?

lib/scholar/decomposition/incremental_pca.ex
```diff
@@ -0,0 +1,355 @@
...
+    opts = NimbleOptions.validate!(opts, @stream_schema)
+    first_batch = Enum.at(batches, 0)
+    model = fit_first_n(first_batch, opts)
+    batches = Stream.drop(batches, 1)
+
+    Enum.reduce(
+      batches,
+      model,
+      # TODO: JIT
+      fn batch, model -> partial_fit(model, batch) end
```
krstopro, line 175: I guess this needs to be jitted?

krstopro (commented):
Two doubts.

josevalim:
> There are two fit/2 clauses: one that takes a tensor and another that takes a stream as a first argument. The one that takes a tensor is defined with deftransform while the one that takes a stream is defined with def. Is this fine?

This is currently undefined behaviour. I would maybe have a function name specific to streams... if I have more ideas I will drop as comment when I review the PR.

> Is it possible to add a popular dataset in incremental_pca_test.exs without adding explorer or scidata dependency to mix.exs?

It is fine to add scidata as a test dependency only.

lib/scholar/decomposition/incremental_pca.ex
```diff
@@ -0,0 +1,355 @@
...
+
+  ## Options
+
+  #{NimbleOptions.docs(@stream_schema)}
+  """
+  def fit(batches = %Stream{}, opts) do
+    opts = NimbleOptions.validate!(opts, @stream_schema)
+    first_batch = Enum.at(batches, 0)
+    model = fit_first_n(first_batch, opts)
+    batches = Stream.drop(batches, 1)
```
josevalim, line 169: This will iterate some lazy streams (such as file) twice. For eample, we will open it once to get the first batch and then again to traverse it. The correct way is to have a single reduce, which it starts with an accumulator of `{:head, ...}` and then you swap it to `{:tail, ...}` for the remaining iterations.

Although I wonder if we should have this function here... perhaps it is best to have `fit_head` and `fit_tail` as public functions and then later on we provide facilities for streaming on any model that has `fit_head` and `fit_tail`?

lib/scholar/decomposition/incremental_pca.ex
josevalim: It will be automatically JITted once it calls `partial_fit_n` since it is a defn.

krstopro: Right, I should have thought about that. :)

josevalim: It is also completely fine to forget those things!

lib/scholar/decomposition/incremental_pca.ex
krstopro: > One problem with doing this using a single function is that there needs to be some starting model as initial accumulator.

This can probably be achieved by setting `num_samples_seen` to `0` and filling tensor fields with `nan`.

josevalim: There is no need to have an initial accumulator, you could have the acc be `:first` and then `model`, and pattern matching distinguighes between them.

But you could also have a model with a single `:initialized` field as a boolean (or even a u8) and then we could easily fold it all into a single `defn`.

krstopro: > There is no need to have an initial accumulator, you could have the acc be :first and then model, and pattern matching distinguighes between them.

I see, that would work.

> But you could also have a model with a single :initialized field as a boolean (or even a u8) and then we could easily fold it all into a single defn.

Checking if `num_samples_seen` is equal to `0` can tell us if the model has been initialised or not. The only issue would be the floating point type of tensor fields, but then we can ask the user to provide it.

krstopro: Alright, it is possible to have a single reduce, but then the problem is that the user needs to specify the type and the number of features as those cannot be inferred without running the stream.

I suppose it is a cleaner approach to have a reducer with two clauses: one for head and one for tail.

krstopro:
I've decided to remove the `fit/2` clause that takes tensor as the first argument as I don't think it's needed. :)

The code should be much cleaner now, but still some problems remain.

lib/scholar/decomposition/incremental_pca.ex
```diff
@@ -0,0 +1,294 @@
...
+    IO.puts("num_components: #{opts[:num_components]}")
+
+    Enum.reduce(
+      batches,
+      nil,
+      fn batch, model -> fit_batch(model, batch, opts) end
+    )
+  end
+
+  defp fit_batch(nil, batch, opts), do: fit_head_n(batch, opts)
```
krstopro, line 109: Using `fit_head/2` instead of `fit_head_n/2` over here won't work. I think it has to do with the compilation. For some reason the options won't be passed to `fit_head/2` and `opts[:num_components]` will be `nil`. I don't know how to fix this.

lib/scholar/decomposition/incremental_pca.ex
```diff
@@ -0,0 +1,311 @@
...
+      Nx.tensor(
+        [
+          [-0.33354005217552185, 0.1048964187502861, -0.8618107080105579, -0.3674643635749817],
+          [-0.5862125754356384, -0.7916879057884216, 0.15874788165092468, -0.06621300429105759]
+        ]
+      )
+      iex> ipca.singular_values
+      Nx.tensor([77.05782028025969, 10.137848854064941])
+  """
+  def fit(batches = %Stream{}, opts) do
```
josevalim, line 106: We should not match on a stream because `%Stream{}` is only one of the possible streams. Instead we should just treat it as an enumerable:

```suggestion
  def fit(batches, opts) do
```

lib/scholar/decomposition/incremental_pca.ex
josevalim: There is a typo in the option name down below, that's probably why: num_components vs num_componenets.

lib/scholar/decomposition/incremental_pca.ex
josevalim: I wonder though if we should instead merge this into PCA and add a new function called `fit_stream`.

krstopro: Yes, I was about to recommend that! Do we need to call it `fit_stream` or `fit_incremental` or we can just add another clause? Two clauses were working over here; one was defined with `deftransform`, the other with `def`.

lib/scholar/decomposition/incremental_pca.ex
krstopro: Damn, good catch. :)

lib/scholar/decomposition/incremental_pca.ex
josevalim: I would do a separate function. Either fit_stream or fit_incremental.

krstopro:
Closing in favor of #291.

PR #295, opened 2024-08-09
Added OPTICS clustering algorithm

norm4nn:
Added:
- OPTICS clustering algorithm
- Its description
- doctests

Currently, the only available clustering method is dbscan, but I plan to add the xi method soon.

### DIFF
--- a/lib/scholar/cluster/optics.ex
+++ b/lib/scholar/cluster/optics.ex
@@ -1,0 +1,201 @@
+defmodule Scholar.Cluster.OPTICS do
+  @moduledoc """
+  OPTICS (Ordering Points To Identify the Clustering Structure), closely related to DBSCAN, finds core sample of high density and expands clusters from them. Unlike DBSCAN, keeps cluster hierarchy for a variable neighborhood radius. Clusters are then extracted using a DBSCAN-like method.
+  """
+  import Nx.Defn
+  require Nx
+
+  opts = [
+    min_samples: [
+      default: 5,
+      type: :pos_integer,
+      doc: "The number of samples in a neighborhood for a point to be considered as a core point."
+    ],
+    max_eps: [
+      default: Nx.Constants.infinity(),
+      type: {:custom, Scholar.Options, :beta, []},
+      doc:
+        "The maximum distance between two samples for one to be considered as in the neighborhood of the other. Default value of Nx.Constants.infinity() will identify clusters across all scales "
+    ],
+    eps: [
+      default: Nx.Constants.nan(),
+      type: {:custom, Scholar.Options, :beta, []},
+      doc:
+        "The maximum distance between two samples for one to be considered as in the neighborhood of the other. By default it assumes the same value as max_eps."
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(opts)
+
+  @doc """
+
+  Perform OPTICS clustering for `x` which is tensor of {n_samples, n_features} shape.
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+  The function returns a labels tensor of shape {n_samples}
+        Cluster labels for each point in the dataset given to fit().
+        Noisy samples are labeled as -1.
+
+  ## Examples
+
+      iex> x = Nx.tensor([[1, 2], [2, 5], [3, 6], [8, 7], [8, 8], [7, 3]])
+      iex> Scholar.Cluster.OPTICS.fit(x, min_samples: 2)
+      #Nx.Tensor<
+        s64[6]
+        [-1, -1, -1, -1, -1, -1]
+      >
+      iex> Scholar.Cluster.OPTICS.fit(x, eps: 4.5, min_samples: 2)
+      #Nx.Tensor<
+        s64[6]
+        [0, 0, 0, 1, 1, 1]
+      >
+      iex> Scholar.Cluster.OPTICS.fit(x, eps: 2, min_samples: 2)
+      #Nx.Tensor<
+        s64[6]
+        [-1, 0, 0, 1, 1, -1]
+      >
+      iex> Scholar.Cluster.OPTICS.fit(x, eps: 1, min_samples: 2)
+      #Nx.Tensor<
+        s64[6]
+        [-1, -1, -1, 0, 0, -1]
+      >
+      iex> Scholar.Cluster.OPTICS.fit(x, eps: 4.5, min_samples: 3)
+      #Nx.Tensor<
+        s64[6]
+        [0, 0, 0, 1, 1, -1]
+      >
+  """
+
+  deftransform fit(x, opts \\ []) do
+    fit_p(x, NimbleOptions.validate!(opts, @opts_schema))
+  end
+
+  defnp fit_p(x, opts \\ []) do
+    {core_distances, reachability, _predecessor, ordering} = compute_optics_graph(x, opts)
+    eps = opts[:eps]
+    cluster_optics_dbscan(reachability, core_distances, ordering, eps: eps)
+  end
+
+  defnp compute_optics_graph(x, opts \\ []) do
+    min_samples = opts[:min_samples]
+    max_eps = opts[:max_eps]
+    n_samples = Nx.axis_size(x, 0)
+    reachability = Nx.broadcast(Nx.Constants.max_finite({:f, 32}), {n_samples})
+    predecessor = Nx.broadcast(-1, {n_samples})
+    neighbors = Scholar.Neighbors.BruteKNN.fit(x, num_neighbors: min_samples)
+    core_distances = compute_core_distances(x, neighbors, min_samples: min_samples)
+
+    core_distances =
+      Nx.select(Nx.greater(core_distances, max_eps), Nx.Constants.infinity(), core_distances)
+
+    ordering = Nx.broadcast(0, {n_samples})
+    processed = Nx.broadcast(0, {n_samples})
+
+    {_order_idx, core_distances, reachability, predecessor, _processed, ordering, _x, _max_eps} =
+      while {order_idx = 0, core_distances, reachability, predecessor, processed, ordering, x,
+             max_eps},
+            order_idx < n_samples do
+        unprocessed_mask = Nx.equal(processed, 0)
+        point = Nx.argmin(Nx.select(unprocessed_mask, reachability, Nx.Constants.infinity()))
+        processed = Nx.put_slice(processed, [point], Nx.broadcast(1, {1}))
+        ordering = Nx.put_slice(ordering, [order_idx], Nx.broadcast(point, {1}))
+
+        {reachability, predecessor} =
+          set_reach_dist(core_distances, reachability, predecessor, point, processed, x,
+            max_eps: max_eps
+          )
+
+        {order_idx + 1, core_distances, reachability, predecessor, processed, ordering, x,
+         max_eps}
+      end
+
+    reachability =
+      Nx.select(
+        Nx.equal(reachability, Nx.Constants.max_finite({:f, 32})),
+        Nx.Constants.infinity(),
+        reachability
+      )
+
+    {core_distances, reachability, predecessor, ordering}
+  end
+
+  defnp compute_core_distances(x, neighbors, opts \\ []) do
+    {_neighbors, distances} = Scholar.Neighbors.BruteKNN.predict(neighbors, x)
+    Nx.slice_along_axis(distances, opts[:min_samples] - 1, 1, axis: 1)
+  end
+
+  defnp set_reach_dist(
+          core_distances,
+          reachability,
+          predecessor,
+          point_index,
+          processed,
+          x,
+          opts \\ []
+        ) do
+    max_eps = opts[:max_eps]
+    n_features = Nx.axis_size(x, 1)
+    n_samples = Nx.axis_size(x, 0)
+    nbrs = Scholar.Neighbors.BruteKNN.fit(x, num_neighbors: n_samples)
+    t = Nx.take(x, point_index, axis: 0)
+    p = Nx.broadcast(t, {1, n_features})
+    {neighbors, distances} = Scholar.Neighbors.BruteKNN.predict(nbrs, p)
+
+    neighbors = Nx.flatten(neighbors)
+    distances = Nx.flatten(distances)
+    indices_ngbrs = Nx.argsort(neighbors)
+    neighbors = Nx.take(neighbors, indices_ngbrs)
+    distances = Nx.take(distances, indices_ngbrs)
+    are_neighbors_processed = Nx.take(processed, neighbors)
+
+    filtered_neighbors =
+      Nx.select(
+        Nx.logical_or(are_neighbors_processed, Nx.greater(distances, max_eps)),
+        Nx.multiply(-1, neighbors),
+        neighbors
+      )
+
+    dists = Nx.flatten(Scholar.Metrics.Distance.pairwise_minkowski(p, x))
+    core_distance = Nx.take(core_distances, point_index)
+    rdists = Nx.max(dists, core_distance)
+    improved = Nx.less(rdists, reachability)
+    improved = Nx.select(improved, filtered_neighbors, -1)
+
+    improved =
+      Nx.select(
+        Nx.logical_and(Nx.equal(improved, -1), Nx.greater(filtered_neighbors, 0)),
+        Nx.multiply(filtered_neighbors, -1),
+        filtered_neighbors
+      )
+
+    rdists = Nx.select(Nx.greater_equal(improved, 0), rdists, 0)
+    reversed_improved = Nx.max(Nx.multiply(improved, -1), 0)
+
+    reachability =
+      Nx.select(Nx.less_equal(improved, 0), Nx.take(reachability, reversed_improved), rdists)
+
+    predecessor =
+      Nx.select(Nx.less_equal(improved, 0), Nx.take(predecessor, reversed_improved), point_index)
+
+    {reachability, predecessor}
+  end
+
+  defnp cluster_optics_dbscan(reachability, core_distances, ordering, opts \\ []) do
+    eps = opts[:eps]
+    far_reach = Nx.flatten(Nx.greater(reachability, eps))
+    near_core = Nx.flatten(Nx.less_equal(core_distances, eps))
+    far_and_not_near = Nx.multiply(far_reach, Nx.subtract(1, near_core))
+    far_reach = Nx.take(far_reach, ordering)
+    near_core = Nx.take(near_core, ordering)
+    far_and_near = Nx.multiply(far_reach, near_core)
+    Nx.as_type(Nx.cumulative_sum(far_and_near), :s8)
+    labels = Nx.subtract(Nx.as_type(Nx.cumulative_sum(far_and_near), :s8), 1)
+    labels = Nx.take(labels, Nx.argsort(ordering))
+    Nx.select(far_and_not_near, -1, labels)
+  end
+end
--- a/lib/scholar/cluster/optics.ex
+++ b/lib/scholar/cluster/optics.ex
@@ -77,7 +77,14 @@ defmodule Scholar.Cluster.OPTICS do
 
   defnp fit_p(x, opts \\ []) do
     {core_distances, reachability, _predecessor, ordering} = compute_optics_graph(x, opts)
-    eps = opts[:eps]
+
+    eps =
+      if Nx.equal(opts[:eps], Nx.Constants.nan()) do
+        opts[:max_eps]
+      else
+        opts[:eps]
+      end
+
     cluster_optics_dbscan(reachability, core_distances, ordering, eps: eps)
   end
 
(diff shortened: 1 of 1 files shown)


lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,208 @@
...
+
+  deftransform fit(x, opts \\ []) do
+    fit_p(x, NimbleOptions.validate!(opts, @opts_schema))
+  end
+
+  defnp fit_p(x, opts \\ []) do
+    {core_distances, reachability, _predecessor, ordering} = compute_optics_graph(x, opts)
+
+    eps =
+      if Nx.equal(opts[:eps], Nx.Constants.nan()) do
```
msluszniak, line 82: I think here this condition is always false as nan is not equal to anything, even itself

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,208 @@
...
+    min_samples = opts[:min_samples]
+    max_eps = opts[:max_eps]
+    n_samples = Nx.axis_size(x, 0)
+    reachability = Nx.broadcast(Nx.Constants.max_finite({:f, 32}), {n_samples})
+    predecessor = Nx.broadcast(-1, {n_samples})
+    neighbors = Scholar.Neighbors.BruteKNN.fit(x, num_neighbors: min_samples)
+    core_distances = compute_core_distances(x, neighbors, min_samples: min_samples)
+
+    core_distances =
+      Nx.select(Nx.greater(core_distances, max_eps), Nx.Constants.infinity(), core_distances)
```
msluszniak, line 101:
```suggestion
      Nx.select(core_distances > max_eps, Nx.Constants.infinity(), core_distances)
```

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,208 @@
...
+
+    ordering = Nx.broadcast(0, {n_samples})
+    processed = Nx.broadcast(0, {n_samples})
+
+    {_order_idx, core_distances, reachability, predecessor, _processed, ordering, _x, _max_eps} =
+      while {order_idx = 0, core_distances, reachability, predecessor, processed, ordering, x,
+             max_eps},
+            order_idx < n_samples do
+        unprocessed_mask = Nx.equal(processed, 0)
```
msluszniak, line 110:
```suggestion
        unprocessed_mask = processed == 0
```

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,208 @@
...
+        point = Nx.argmin(Nx.select(unprocessed_mask, reachability, Nx.Constants.infinity()))
+        processed = Nx.put_slice(processed, [point], Nx.broadcast(1, {1}))
+        ordering = Nx.put_slice(ordering, [order_idx], Nx.broadcast(point, {1}))
```
msluszniak, line 113: I'd rather use `Nx.new_axis` here

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,208 @@
...
+            max_eps: max_eps
+          )
+
+        {order_idx + 1, core_distances, reachability, predecessor, processed, ordering, x,
+         max_eps}
+      end
+
+    reachability =
+      Nx.select(
+        Nx.equal(reachability, Nx.Constants.max_finite({:f, 32})),
```
msluszniak, line 126:
```suggestion
        reachability == Nx.Constants.max_finite({:f, 32})),
```

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,208 @@
...
+      Nx.select(
+        Nx.logical_or(are_neighbors_processed, Nx.greater(distances, max_eps)),
+        Nx.multiply(-1, neighbors),
+        neighbors
+      )
+
+    dists = Nx.flatten(Scholar.Metrics.Distance.pairwise_minkowski(p, x))
+    core_distance = Nx.take(core_distances, point_index)
+    rdists = Nx.max(dists, core_distance)
+    improved = Nx.less(rdists, reachability)
```
msluszniak, line 173:
```suggestion
    improved = rdists < reachability
```

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,208 @@
...
+    improved = Nx.select(improved, filtered_neighbors, -1)
+
+    improved =
+      Nx.select(
+        Nx.logical_and(Nx.equal(improved, -1), Nx.greater(filtered_neighbors, 0)),
+        Nx.multiply(filtered_neighbors, -1),
```
msluszniak, line 179: You can simplify this as above ;)

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,208 @@
...
+        filtered_neighbors
+      )
+
+    rdists = Nx.select(Nx.greater_equal(improved, 0), rdists, 0)
+    reversed_improved = Nx.max(Nx.multiply(improved, -1), 0)
```
msluszniak, line 184: Same here

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,208 @@
...
+
+    reachability =
+      Nx.select(Nx.less_equal(improved, 0), Nx.take(reachability, reversed_improved), rdists)
+
+    predecessor =
+      Nx.select(Nx.less_equal(improved, 0), Nx.take(predecessor, reversed_improved), point_index)
```
msluszniak, line 190: Same here

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,208 @@
...
+    indices_ngbrs = Nx.argsort(neighbors)
+    neighbors = Nx.take(neighbors, indices_ngbrs)
+    distances = Nx.take(distances, indices_ngbrs)
+    are_neighbors_processed = Nx.take(processed, neighbors)
+
+    filtered_neighbors =
+      Nx.select(
+        Nx.logical_or(are_neighbors_processed, Nx.greater(distances, max_eps)),
+        Nx.multiply(-1, neighbors),
+        neighbors
```
msluszniak, line 167: `or` and `*` will also work here

msluszniak:
Thank you for the PR! I dropped some small comments for now. I'll try to make a more careful pass soon ;)

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,207 @@
+defmodule Scholar.Cluster.OPTICS do
+  @moduledoc """
+  OPTICS (Ordering Points To Identify the Clustering Structure), closely related to DBSCAN, finds core sample of high density and expands clusters from them. Unlike DBSCAN, keeps cluster hierarchy for a variable neighborhood radius. Clusters are then extracted using a DBSCAN-like method.
```
josevalim, line 3: The first sentence should be really short.

Also please add a new line so it wraps around 80-100 characters. :)

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,207 @@
...
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+  The function returns a labels tensor of shape {n_samples}
+        Cluster labels for each point in the dataset given to fit().
+        Noisy samples are labeled as -1.
```
josevalim, line 42: Two space indentation perhaps?

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,207 @@
...
+
+  Perform OPTICS clustering for `x` which is tensor of {n_samples, n_features} shape.
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+  The function returns a labels tensor of shape {n_samples}
```
josevalim, line 40:
```suggestion
  The function returns a labels tensor of shape `{n_samples}`.
```

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,207 @@
...
+      doc:
+        "The maximum distance between two samples for one to be considered as in the neighborhood of the other. By default it assumes the same value as max_eps."
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(opts)
+
+  @doc """
+
+  Perform OPTICS clustering for `x` which is tensor of {n_samples, n_features} shape.
```
josevalim, line 32:
```suggestion
  Perform OPTICS clustering for `x` which is tensor of `{n_samples, n_features}` shape.
```

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,207 @@
...
+        reachability == Nx.Constants.max_finite({:f, 32}),
+        Nx.Constants.infinity(),
+        reachability
+      )
+
+    {core_distances, reachability, predecessor, ordering}
+  end
+
+  defnp compute_core_distances(x, neighbors, opts \\ []) do
+    {_neighbors, distances} = Scholar.Neighbors.BruteKNN.predict(neighbors, x)
```
josevalim, line 135: Should we allow the KNN algorithm to be given, like we do for `algorithm` in `Scholar.Neighbors.KNNClassifier`?

lib/scholar/cluster/optics.ex
```diff
@@ -1,6 +1,10 @@
 defmodule Scholar.Cluster.OPTICS do
   @moduledoc """
-  OPTICS (Ordering Points To Identify the Clustering Structure), closely related to DBSCAN, finds core sample of high density and expands clusters from them. Unlike DBSCAN, keeps cluster hierarchy for a variable neighborhood radius. Clusters are then extracted using a DBSCAN-like method.
+  OPTICS (Ordering Points To Identify the Clustering Structure) is an algorithm
+  for finding density-based clusters in spatial data. It is closely related
+  to DBSCAN, finds core sample of high density and expands clusters from them. 
+  Unlike DBSCAN, keeps cluster hierarchy for a variable neighborhood radius. 
+  Clusters are then extracted using a DBSCAN-like method.
```
josevalim, line 7:
```suggestion
  OPTICS (Ordering Points To Identify the Clustering Structure) is an algorithm
  for finding density-based clusters in spatial data.

  It is closely related to DBSCAN, finds core sample of high density and expands
  clusters from them.  Unlike DBSCAN, keeps cluster hierarchy for a variable
  neighborhood radius.  Clusters are then extracted using a DBSCAN-like
  method.
```

lib/scholar/cluster/optics.ex
```diff
@@ -69,14 +95,46 @@ defmodule Scholar.Cluster.OPTICS do
...
+
+        :kd_tree ->
+          Scholar.Neighbors.KDTree
+
+        :random_projection_forest ->
+          Scholar.Neighbors.RandomProjectionForest
+
+        module when is_atom(module) ->
+          module
+      end
```
josevalim, line 130: Please remember to run `mix format` :)

lib/scholar/cluster/optics.ex
```diff
@@ -69,14 +95,46 @@ defmodule Scholar.Cluster.OPTICS do
...
+    model = algorithm_module.fit(x, algorithm_opts)
+    {_neighbors, distances} = algorithm_module.predict(model, x)
```
josevalim, line 132: Can this logic be moved to a `defn` or `defnp`? It would be nice to have it in there, because then all of the fit/predict code would folded and jitted into one too.

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,257 @@
...
+    predecessor = Nx.broadcast(-1, {n_samples})
+    core_distances = Nx.slice_along_axis(distances, opts[:min_samples] - 1, 1, axis: 1)
+    core_distances =
+      Nx.select(core_distances > max_eps, Nx.Constants.infinity(), core_distances)
+
+    ordering = Nx.broadcast(0, {n_samples})
+    processed = Nx.broadcast(0, {n_samples})
+
+    {_order_idx, core_distances, reachability, predecessor, _processed, ordering, _x, _max_eps} =
+      while {order_idx = 0, core_distances, reachability, predecessor, processed, ordering, x,
```
krstopro, line 162: I think this can be vectorized.

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,257 @@
...
+          predecessor,
+          point_index,
+          processed,
+          x,
+          opts \\ []
+        ) do
+    max_eps = opts[:max_eps]
+    n_features = Nx.axis_size(x, 1)
+    n_samples = Nx.axis_size(x, 0)
+    nbrs = Scholar.Neighbors.BruteKNN.fit(x, num_neighbors: n_samples)
```
krstopro, line 201: You are using `Scholar.Neighbors.BruteKNN` here even though the k-NN algorithm can be passed as an option to `fit/2`. Is this intended?

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,257 @@
...
+        opts[:eps]
+      end
+
+    cluster_optics_dbscan(reachability, core_distances, ordering, eps: eps)
+  end
+
+  defnp compute_optics_graph(x, distances, opts \\ []) do
+    max_eps = opts[:max_eps]
+    n_samples = Nx.axis_size(x, 0)
+    reachability = Nx.broadcast(Nx.Constants.max_finite({:f, 32}), {n_samples})
```
krstopro, line 152: Any specific reason to use `Nx.Constants.max_finite/1` instead of `Nx.Constants.infinity/1`? Also, I don't think it's a good idea to hardcode the types, in this case `:f32`. Better use `Nx.type(x)`.

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,257 @@
...
+  opts = [
+    min_samples: [
+      default: 5,
+      type: :pos_integer,
+      doc: """
+      The number of samples in a neighborhood for a point to be considered as a core point.
+      """
+    ],
+    max_eps: [
+      default: Nx.Constants.infinity(),
```
krstopro, line 21: I wouldn't set the default value here, rather in `fit/2`. I think it is better checking if `opts[:max_eps]` is `nil` and then setting it to `Nx.Constanst.infinity(Nx.type(x))`.

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,257 @@
...
+      type: {:custom, Scholar.Options, :beta, []},
+      doc: """
+        The maximum distance between two samples for one to be considered as in the neighborhood of the other. 
+        Default value of Nx.Constants.infinity() will identify clusters across all scales.
+      """
+    ],
+    eps: [
+      default: Nx.Constants.infinity(),
```
krstopro, line 29: Same here.

krstopro (commented):
Some comments from my side. There are several more potential problems:
- There is no struct within the module. This seems a bit strange.
- Metric might need to be passed as an option as well.
- Don't forget to write unit-tests.

lib/scholar/cluster/optics.ex
josevalim: Can you provide a more specific to hint on how to vectorize? Or maybe you can do a separate pass later and vectorize it? Given they are getting acquainted with the codebase + Nx, it may be a bit too complex to pull off. :)

krstopro: To be honest, I didn't think much about it, nor am I insisting on vectorization for this pull request. I just mentioned it as something to think of. Should have written that as well. 😅

That said, the fact that the condition in the loop is `order_idx < num_samples` and `order_idx` is incremented in every iteration kinda indicates that this could be vectorized. I need to have a closer look.

lib/scholar/cluster/optics.ex
norm4nn: I tried but encountered an issue at lines 131 and 132:
```
model = algorithm_module.fit(x, algorithm_opts).
{_neighbors, distances} = algorithm_module.predict(model, x)
```
These lines cannot be used within a numerical function, as it results in an error:

_Invalid numerical expression: algorithm_module.fit(x, algorithm_opts)._

It seems that a similar issue exists in the module you mentioned, `Scholar.Neighbors.KNNClassifier`, where this code is also placed within a deftransform function.

lib/scholar/cluster/optics.ex
norm4nn: You're absolutely right about the struct—it should be included to maintain consistency.

Also, I get a headache just thinking about vectorizing this entire section of code . Implementing it in its current state was already pretty complex.

lib/scholar/cluster/optics.ex
norm4nn: It's not intended, I missed that. I will repair in next commit.

lib/scholar/cluster/optics.ex
krstopro: Very well, you can leave vectorization for someone else :)

lib/scholar/cluster/optics.ex
norm4nn: There is a reason. In line 168:
point = Nx.argmin(Nx.select(unprocessed_mask, reachability, Nx.Constants.infinity()))
I want to ensure that a point already processed won’t be selected again as the next point to process. At the same time, the reachability tensor should initially hold the highest possible values (preferably infinity) because I’m searching for the smallest element in each step of the while loop.

That said, I'm not particularly proud of this approach and would appreciate any suggestions for improvement.

lib/scholar/cluster/optics.ex
josevalim: I see, we need to do this slightly different, I will write a comment above.

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,259 @@
...
+        [0, 0, 0, 1, 1, -1]
+      >
+      iex> Scholar.Cluster.OPTICS.fit(x, max_eps: 2, min_samples: 1, algorithm: :kd_tree, metric: {:minkowski, 1})
+      #Nx.Tensor<
+        s64[6]
+        [0, 1, 1, 2, 2, 3]
+      >
+  """
+
+  deftransform fit(x, opts \\ []) do
```
josevalim, line 107: So please make this one a `defn` and then call a `deftransform` that will split the options and invoke the algorithm. The important point is that you need to pass through at least one `defn` before you do any work, then everything from there onwards is jitted. :)

norm4nn: Ok, I will do that

lib/scholar/cluster/optics.ex
norm4nn: It can be problematic, in edge case where `x` tensor of type {:s, 64} the following code will fail: 
```
x = Nx.tensor([[1, 2], [2, 5], [3, 6], [8, 7], [8, 8], [7, 3]])
Nx.Constants.infinity(Nx.type(x))
```

with error
```
** (FunctionClauseError) no function clause matching in Nx.Type.infinity_binary/1    
    
    The following arguments were given to Nx.Type.infinity_binary/1:
    
        # 1
        {:s, 64}
    
    Attempted function clauses (showing 4 out of 4):
    
        def infinity_binary({:bf, 16})
        def infinity_binary({:f, 16})
        def infinity_binary({:f, 32})
        def infinity_binary({:f, 64})
    
    (nx 0.7.3) lib/nx/type.ex:120: Nx.Type.infinity_binary/1
    (nx 0.7.3) lib/nx/constants.ex:96: Nx.Constants.infinity/2
    #cell:y6u577gyyh6jwkmd:2: (file)
```
so it seems that it will only work if `x` is float type tensor. 

Similar problem with hardcoded float type `reachability`:
`reachability = Nx.broadcast(Nx.Constants.max_finite({:f, 32}), {n_samples})`
I can't set its type to `Nx.type(x)` like this:
`reachability = Nx.broadcast(Nx.Constants.max_finite(Nx.type(x), {n_samples})`
because it will break each time when `Nx.type(x)` returns something different than float type - any of `{:bf, 16}`, `{:f, 16}`, `{:f, 32}`, `{:f, 64}`. `reachability` tensor should always be float type.

krstopro: You can use `Scholar.Shared.to_float_type/1`. You should convert `x` to a floating-point tensor using `Scholar.Shared.to_float` anyway.

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,302 @@
+defmodule Scholar.Cluster.OPTICS do
+  @moduledoc """
+    OPTICS (Ordering Points To Identify the Clustering Structure) is an algorithm
+  for finding density-based clusters in spatial data.
```
josevalim, line 4:
```suggestion
  OPTICS (Ordering Points To Identify the Clustering Structure) is an algorithm
  for finding density-based clusters in spatial data.

```

josevalim (approved):

lib/scholar/cluster/optics.ex
```diff
@@ -0,0 +1,302 @@
...
+    min_samples = opts[:min_samples]
+
+    if min_samples < 2 do
+      raise ArgumentError,
+            """
+            min_samples must be an int in the range [2, inf), got min_samples = #{inspect(min_samples)}
+            """
+    end
+
+    algorithm_opts = Keyword.put(algorithm_opts, :num_neighbors, min_samples)
```
krstopro, line 137: I don't think this line is needed anymore.

krstopro (commented):
One more comment.

I will try to have a detailed look one of these days.

lib/scholar/cluster/optics.ex
norm4nn: It turns out that it is needed, in the same function, few lines below I call function `.fit()` from `algorithm_module` , which requires `:num_neighbors`. However the value I set here doesn't have any meaning so it can be any positive integer like 1 for example. It wouldn't be required if I could pass just `algorithm_module` to module struct, but it doesn't seem to be possible.

josevalim:

Issue #144, opened 2023-07-24
Support Batching in Classification and Regression Metrics


msluszniak:
We need to add the same support for batching as in the Distances module. Before doing that, we need to add basic support for

josevalim:
I think you forgot to complete the issue description. :)

norm4nn:
Could you provide more information about batching in classification metrics? I'm unsure how this concept is supposed to work. Additionally, I couldn't find anything related to batching in scikit-learn's classification metrics.

RicardoSantos-99:
Hi @msluszniak

I'd like to clarify what "batching" means in this context, as I see the issue description is brief.

Based on the recent commits that added batching support to **Regression** (`#297`) and **Distance** metrics, I understand "batching" refers to adding an `axes:` option parameter to the classification metric functions. This would allow computing the metric across specific axes instead of always returning a scalar.

For example:

**Current (no batching):**
```elixir
Scholar.Metrics.Classification.accuracy(y_true, y_pred)
# Returns: a scalar
```
Proposed (with batching):
```elixir
Scholar.Metrics.Classification.accuracy(y_true, y_pred, axes: [0])
# Returns: a tensor, computed per batch along axis 0
```

Is this the correct understanding? If so, the implementation pattern should be similar to what was done in regression metrics, ading deftransform + defn functions with NimbleOptions.validate! for the axes: parameter?

PR #297, opened 2024-09-02
Added batching to regression metrics

norm4nn:
Following up on #144 issue:

I've added batching to the `Regression` metrics, similar to the batching implemented in the `Distances` metrics. However, I'm not entirely sure if I fully understood the issue, so I would greatly appreciate any feedback.

I did not make changes to the following metrics:

- `mean_pinball_loss`
- `d2_absolute_error_score`
- `d2_pinball_score`

These metrics appear to already support batching through the `:multioutput` argument. If that's the case, perhaps we should consider standardizing the argument to `:axes` for consistency across similar functions.

Please let me know if I’m mistaken about anything.

### DIFF
--- a/lib/scholar/metrics/regression.ex
+++ b/lib/scholar/metrics/regression.ex
@@ -34,10 +34,10 @@ defmodule Scholar.Metrics.Regression do
       (1.0 if prediction is perfect, 0.0 otherwise)
       """
     ]
-  ]
+  ] ++ general_schema
 
   @general_schema NimbleOptions.new!(general_schema)
-  @r2_schema NimbleOptions.new!(general_schema ++ r2_schema)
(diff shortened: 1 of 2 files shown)


josevalim (approved):
LGTM!

@msluszniak, is this what you had in mind?

msluszniak:
Yes, that's exactly what I meant. Originally, I added option `multioutput` to mimic sklearn, but now maybe it'll be more beneficial to remove it and support it as you do it @norm4nn, what do you think?

norm4nn:
Sounds good, I’ll go with your suggestion!

josevalim (approved):

msluszniak (approved):
Superb! Just only `mix format` and we are good to go :)

josevalim:

PR #299, opened 2024-09-04
Change `Utils.flip_svd/2` to `flip_svd/2` inside `Scholar.Decomposition.PCA`

### LINKED ISSUE #298: Non-terminating PCA decomposition
Since we merged #289, PCA decomposition suite no longer finishes running for me. The doctest itself fails. @krstopro, can you please take a quick look when you have some time?

krstopro:
Fixes a bug inside `Scholar.Decomposition.PCA` where `Scholar.Decomposition.Utils.flip_svd/2` was used instead of `flip_svd/2` from the same module.

Closes #298


josevalim (approved):

josevalim:
CI was failing and now it no longer fails, so we are all good :)

krstopro:
Yeah, unit tests were passing originally on my machine, but that is because I had `Scholar.Decomposition.Utils` as my first approach was to have a separate module for incremental PCA. I didn't bother looking at CI as it was taking way too long to finish (probably had something to do with CrowdStrike accident). Apologies for that.

PR #303, opened 2024-10-20
Add KNNImputer

srzeszut:
I have added the KNNImputer and I am currently implementing tests to ensure that it behaves as expected across various scenarios, including edge cases.

### DIFF
--- a/lib/scholar/impute/knn_imputer.ex
+++ b/lib/scholar/impute/knn_imputer.ex
@@ -121,7 +121,7 @@ defmodule Scholar.Impute.KNNImputer do
   ## Examples
 
       iex> x = Nx.tensor([[40.0, 2.0],[4.0, 5.0],[7.0, :nan],[:nan, 8.0],[11.0, 11.0]])
-      iex> imputer = Scholar.Impute.KNNImputer.fit(x, strategy: :mean)
+      iex> imputer = Scholar.Impute.KNNImputer.fit(x, number_of_neighbors: 2)
       iex> Scholar.Impute.KNNImputer.transform(imputer, x)
       Nx.tensor(
         [
--- a/lib/scholar/impute/knn_imputer.ex
+++ b/lib/scholar/impute/knn_imputer.ex
@@ -103,7 +103,9 @@ defmodule Scholar.Impute.KNNImputer do
             "Number of neighbors rows must be less than number valid of rows - 1 (valid row is row with more than 1 non nan value)"
     end
 
-    statistics = knn_impute(x, num_neighbors: num_neighbors)
+    placeholder_value = Nx.Constants.nan() |> Nx.tensor()
+
+    statistics = knn_impute(x, placeholder_value, num_neighbors: num_neighbors)
     #     statistics = all_nan_rows_count
     missing_values = opts[:missing_values]
     %__MODULE__{statistics: statistics, missing_values: missing_values}
@@ -136,13 +138,12 @@ defmodule Scholar.Impute.KNNImputer do
     Nx.select(mask, statistics, x)
   end
 
-  defn knn_impute(x, opts \\ []) do
+  defnp knn_impute(x, placeholder_value, opts \\ []) do
     mask = Nx.is_nan(x)
     {num_rows, num_cols} = Nx.shape(x)
     num_neighbors = opts[:num_neighbors]
 
-    placeholder_value = Nx.Constants.nan()
-    values_to_impute = Nx.broadcast(Nx.tensor(placeholder_value), x)
+    values_to_impute = Nx.broadcast(placeholder_value, x)
 
     {_, values_to_impute} =
       while {{row = 0, mask, num_neighbors, num_rows, x}, values_to_impute},
--- a/lib/scholar/impute/knn_imputer.ex
+++ b/lib/scholar/impute/knn_imputer.ex
@@ -73,15 +73,15 @@ defmodule Scholar.Impute.KNNImputer do
     end
 
     if opts[:missing_values] != :nan and
-       Nx.any(Nx.is_nan(x)) == Nx.tensor(1, type: :u8) do
+         Nx.any(Nx.is_nan(x)) == Nx.tensor(1, type: :u8) do
       raise ArgumentError,
             ":missing_values other than :nan possible only if there is no Nx.Constant.nan() in the array"
     end
 
     x =
       if opts[:missing_values] != :nan,
-         do: Nx.select(Nx.equal(x, opts[:missing_values]), Nx.Constants.nan(), x),
-         else: x
+        do: Nx.select(Nx.equal(x, opts[:missing_values]), Nx.Constants.nan(), x),
+        else: x
 
     num_neighbors = opts[:number_of_neighbors]
 
--- a/lib/scholar/impute/knn_imputer.ex
+++ b/lib/scholar/impute/knn_imputer.ex
@@ -46,11 +46,10 @@ defmodule Scholar.Impute.KNNImputer do
     [`Nx.Constant.nan/0`](https://hexdocs.pm/nx/Nx.Constants.html#nan/0) values.
 
   ## Examples
-
-      iex> x = Nx.tensor([[40.0, 2.0],[4.0, 5.0],[7.0, :nan],[:nan, 8.0],[11.0, 11.0]])
-      iex> Scholar.Impute.KNNImputer.fit(x, number_of_neighbors: 2)
-      %Scholar.Impute.KNNImputer{
-      statistics: #Nx.Tensor<
+  iex> x = Nx.tensor([[40.0, 2.0],[4.0, 5.0],[7.0, :nan],[:nan, 8.0],[11.0, 11.0]])
+  iex> Scholar.Impute.KNNImputer.fit(x, number_of_neighbors: 2)
+  %Scholar.Impute.KNNImputer{
+  statistics: #Nx.Tensor<
         f32[5][2]
         [
             [NaN, NaN],
@@ -59,10 +58,9 @@ defmodule Scholar.Impute.KNNImputer do
             [7.5, NaN],
             [NaN, NaN]
           ]
-      >,
-      missing_values: :nan
-      }
-
+  >,
+  missing_values: :nan
+  }
   """
 
   deftransform fit(x, opts \\ []) do
@@ -93,7 +91,6 @@ defmodule Scholar.Impute.KNNImputer do
 
     {rows, cols} = Nx.shape(x)
 
-    # TODO calculate all nan rows
     row_nan_count = Nx.sum(Nx.is_nan(x), axes: [1])
     # row with only 1 non nan value is also considered as all nan row
     all_nan_rows =
@@ -125,7 +122,6 @@ defmodule Scholar.Impute.KNNImputer do
       iex> imputer = Scholar.Impute.KNNImputer.fit(x, strategy: :mean)
       iex> Scholar.Impute.KNNImputer.transform(imputer, x)
       Nx.tensor(
-         f32[5][2]
         [
           [40.0, 2.0],
           [4.0, 5.0],
--- a/lib/scholar/impute/knn_imputer.ex
+++ b/lib/scholar/impute/knn_imputer.ex
@@ -1,0 +1,259 @@
+defmodule Scholar.Impute.KNNImputer do
+  @moduledoc """
+  Imputer for completing missing values using k-Nearest Neighbors.
+
+  Each sample's missing values are imputed using the mean value from
+    `n_neighbors` nearest neighbors found in the training set. Two samples are
+    close if the features that neither is missing are close.
+  """
+  import Nx.Defn
+  import Scholar.Metrics.Distance
+
+  @derive {Nx.Container, keep: [:missing_values], containers: [:statistics]}
+  defstruct [:statistics, :missing_values]
+
+  opts_schema = [
+    missing_values: [
+      type: {:or, [:float, :integer, {:in, [:nan]}]},
+      default: :nan,
+      doc: ~S"""
+      The placeholder for the missing values. All occurrences of `:missing_values` will be imputed.
+      """
+    ],
+    number_of_neighbors: [
+      type: :pos_integer,
+      default: 2,
+      doc: "The number of nearest neighbors."
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(opts_schema)
+
+  @doc """
+  Imputer for completing missing values using k-Nearest Neighbors.
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+    The function returns a struct with the following parameters:
+
+    * `:missing_values` - the same value as in `:missing_values`
+
+    * `:statistics` - The imputation fill value for each feature. Computing statistics can result in
+    [`Nx.Constant.nan/0`](https://hexdocs.pm/nx/Nx.Constants.html#nan/0) values.
+
+  ## Examples
+
+      iex> x = Nx.tensor([[40.0, 2.0],[4.0, 5.0],[7.0, :nan],[:nan, 8.0],[11.0, 11.0]])
+      iex> Scholar.Impute.KNNImputer.fit(x, number_of_neighbors: 2)
+      %Scholar.Impute.KNNImputer{
+      statistics: #Nx.Tensor<
+        f32[5][2]
+        [
+            [NaN, NaN],
+            [NaN, NaN],
+            [NaN, 8.0],
+            [7.5, NaN],
+            [NaN, NaN]
+          ]
+      >,
+      missing_values: :nan
+      }
+
+  """
+
+  deftransform fit(x, opts \\ []) do
+    opts = NimbleOptions.validate!(opts, @opts_schema)
+
+    input_rank = Nx.rank(x)
+
+    if input_rank != 2 do
+      raise ArgumentError, "Wrong input rank. Expected: 2, got: #{inspect(input_rank)}"
+    end
+
+    if opts[:missing_values] != :nan and
+       Nx.any(Nx.is_nan(x)) == Nx.tensor(1, type: :u8) do
+      raise ArgumentError,
+            ":missing_values other than :nan possible only if there is no Nx.Constant.nan() in the array"
+    end
+
+    x =
+      if opts[:missing_values] != :nan,
+         do: Nx.select(Nx.equal(x, opts[:missing_values]), Nx.Constants.nan(), x),
+         else: x
+
+    num_neighbors = opts[:number_of_neighbors]
+
+    if num_neighbors < 1 do
+      raise ArgumentError, "Number of neighbors must be greater than 0"
+    end
+
+    {rows, cols} = Nx.shape(x)
+
+    # TODO calculate all nan rows
+    row_nan_count = Nx.sum(Nx.is_nan(x), axes: [1])
+    # row with only 1 non nan value is also considered as all nan row
+    all_nan_rows =
+      Nx.select(Nx.greater_equal(row_nan_count, cols - 1), Nx.tensor(1), Nx.tensor(0))
+
+    all_nan_rows_count = Nx.sum(all_nan_rows)
+
+    if num_neighbors > rows - 1 - Nx.to_number(all_nan_rows_count) do
+      raise ArgumentError,
+            "Number of neighbors rows must be less than number valid of rows - 1 (valid row is row with more than 1 non nan value)"
+    end
+
+    statistics = knn_impute(x, num_neighbors: num_neighbors)
+    #     statistics = all_nan_rows_count
+    missing_values = opts[:missing_values]
+    %__MODULE__{statistics: statistics, missing_values: missing_values}
+  end
+
+  @doc """
+  Impute all missing values in `x` using fitted imputer.
+
+  ## Return Values
+
+  The function returns input tensor with NaN replaced with values saved in fitted imputer.
+
+  ## Examples
+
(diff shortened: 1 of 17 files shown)


lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
...
+
+    The function returns a struct with the following parameters:
+
+    * `:missing_values` - the same value as in `:missing_values`
+
+    * `:statistics` - The imputation fill value for each feature. Computing statistics can result in
+    [`Nx.Constant.nan/0`](https://hexdocs.pm/nx/Nx.Constants.html#nan/0) values.
+
+  ## Examples
+  iex> x = Nx.tensor([[40.0, 2.0],[4.0, 5.0],[7.0, :nan],[:nan, 8.0],[11.0, 11.0]])
```
josevalim, line 49: Remember to indent docs example with 4 spaces from the heredoc. So you must add four spaces here :)

lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
...
+    if input_rank != 2 do
+      raise ArgumentError, "Wrong input rank. Expected: 2, got: #{inspect(input_rank)}"
+    end
+
+    if opts[:missing_values] != :nan and
+         Nx.any(Nx.is_nan(x)) == Nx.tensor(1, type: :u8) do
+      raise ArgumentError,
+            ":missing_values other than :nan possible only if there is no Nx.Constant.nan() in the array"
+    end
+
```
josevalim, line 80: This check does not really work in Nx. If you call `fit` inside `Nx.Defn.jit`, then `x` is an expression, and we can't read its values to find out if there is a nan or not. The best we can do is to remove this check and document it.

lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
...
+    {rows, cols} = Nx.shape(x)
+
+    row_nan_count = Nx.sum(Nx.is_nan(x), axes: [1])
+    # row with only 1 non nan value is also considered as all nan row
+    all_nan_rows =
+      Nx.select(Nx.greater_equal(row_nan_count, cols - 1), Nx.tensor(1), Nx.tensor(0))
+
+    all_nan_rows_count = Nx.sum(all_nan_rows)
+
+    if num_neighbors > rows - 1 - Nx.to_number(all_nan_rows_count) do
```
josevalim, line 101: Same here, this code won't work because, when you have an expression, you can't get a number from it. Can we remove this check? What happens if we don't check for this condition?

josevalim: You can test this by calling `fit` after jitting it with `Nx.Defn.fit`.

lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
...
+      raise ArgumentError,
+            "Number of neighbors rows must be less than number valid of rows - 1 (valid row is row with more than 1 non nan value)"
+    end
+
+    placeholder_value = Nx.Constants.nan() |> Nx.tensor()
+
+    statistics = knn_impute(x, placeholder_value, num_neighbors: num_neighbors)
```
josevalim, line 108: You should be fine to pass it as a regular argument, since it is always a number?

```suggestion
    statistics = knn_impute(x, placeholder_value, num_neighbors)
```

lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
...
+      while {{row = 0, mask, num_neighbors, num_rows, x}, values_to_impute},
+            Nx.less(row, num_rows) do
+        {_, values_to_impute} =
+          while {{col = 0, mask, num_neighbors, num_cols, row, x}, values_to_impute},
+                Nx.less(col, num_cols) do
+            if mask[row][col] > 0 do
+              {rows, cols} = Nx.shape(x)
+
+              neighbor_avg =
+                calculate_knn(x, row, col, rows: rows, num_neighbors: opts[:num_neighbors])
```
josevalim, line 158:
```suggestion
                calculate_knn(x, row, col, rows, num_neighbors)
```

lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
...
+
+    # minus nan column
+    coordinates = coordinates - 1
+
+    # inputes zeros in nan_col to calculate distance with squared_euclidean
+    new_row = Nx.indexed_put(row, Nx.new_axis(nan_col, 0), Nx.tensor(0))
+
+    # if potential neighbor has nan in nan_col, we don't want to calculate distance and the case if potential_neighbour is the row to impute
+    {potential_neighbor} =
+      if potential_neighbor[nan_col] == Nx.Constants.nan() do
```
josevalim, line 224: I am not sure if this check is guaranteed to work, given two NaNs are not guaranteed to be equal. Using `Nx.is_nan` would be more appropriate.

lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
...
+  """
+  import Nx.Defn
+  import Scholar.Metrics.Distance
+
+  @derive {Nx.Container, keep: [:missing_values], containers: [:statistics]}
+  defstruct [:statistics, :missing_values]
+
+  opts_schema = [
+    missing_values: [
+      type: {:or, [:float, :integer, {:in, [:nan]}]},
```
msluszniak, line 17: Why `:nan` here is wrapped inside `:in`?

lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
...
+
+    if opts[:missing_values] != :nan and
+         Nx.any(Nx.is_nan(x)) == Nx.tensor(1, type: :u8) do
+      raise ArgumentError,
+            ":missing_values other than :nan possible only if there is no Nx.Constant.nan() in the array"
+    end
+
+    x =
+      if opts[:missing_values] != :nan,
+        do: Nx.select(Nx.equal(x, opts[:missing_values]), Nx.Constants.nan(), x),
```
msluszniak, line 83: Use `Nx.is_nan` here NaN is not equal to itself

lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
...
+    if num_neighbors < 1 do
+      raise ArgumentError, "Number of neighbors must be greater than 0"
+    end
+
+    {rows, cols} = Nx.shape(x)
+
+    row_nan_count = Nx.sum(Nx.is_nan(x), axes: [1])
+    # row with only 1 non nan value is also considered as all nan row
+    all_nan_rows =
+      Nx.select(Nx.greater_equal(row_nan_count, cols - 1), Nx.tensor(1), Nx.tensor(0))
```
msluszniak, line 97: From the check in 88th line, everything might be moved to defn as you cannot effectively execute the check below. Then, this can be simplify to `all_nan_rows = row_nan_count <= cols - 1)`

lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
...
+
+        if i == nan_row do
+          distance = Nx.Constants.infinity({:f, 32})
+          row_distances = Nx.indexed_put(row_distances, Nx.new_axis(i, 0), distance)
+          {{i + 1, x, row_with_value_to_fill, rows, nan_row, nan_col}, row_distances}
+        else
+          distance = nan_euclidian(row_with_value_to_fill, nan_col, potential_donor)
+          row_distances = Nx.indexed_put(row_distances, Nx.new_axis(i, 0), distance)
+          {{i + 1, x, row_with_value_to_fill, rows, nan_row, nan_col}, row_distances}
+        end
```
msluszniak, line 202:
```suggestion
        distance = if i == nan_row do
          Nx.Constants.infinity({:f, 32})
        else
          nan_euclidian(row_with_value_to_fill, nan_col, potential_donor)
        end
        row_distances = Nx.indexed_put(row_distances, Nx.new_axis(i, 0), distance)
        {{i + 1, x, row_with_value_to_fill, rows, nan_row, nan_col}, row_distances}
```

lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
...
+
+        {potential_neighbor}
+      end
+
+    # calculates how many values are present in the row without nan_col to calculate weight for the distance
+    present_coordinates = Nx.sum(Nx.logical_not(Nx.is_nan(potential_neighbor))) - 1
+
+    # if row has all nans we skip it
+    {weight, potential_neighbor} =
+      if present_coordinates == 0 do
```
msluszniak, line 240: As mentioned in comment up, try to replace "bare" numbers with typed tensors

lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
+defmodule Scholar.Impute.KNNImputer do
```
msluszniak, line 1: I think it should be written with double t `KNNImputter` like formatter etc.

msluszniak (commented):
Thanks for the PR, I dropped some comments :))

krstopro:
Hi @srzeszut and thanks for the pull request. I’m traveling now and don’t have my laptop with me. Will be back this Sunday, so I will have a look probably next week.

lib/scholar/impute/knn_imputer.ex
srzeszut: I found this check in simple imputer 
https://github.com/elixir-nx/scholar/blob/main/lib/scholar/impute/simple_imputer.ex
Are you sure it won't work?

lib/scholar/impute/knn_imputer.ex
srzeszut: when i pass it like this  calculate_knn(x, row, col, rows, num_neighbors) the top_k function gets tensor instead of scalar and it does't work.

lib/scholar/impute/knn_imputer.ex
srzeszut: could you suggest how to change it please?
because this is not compiling 
type: {:or, [:float, :integer,:nan]},

lib/scholar/impute/knn_imputer.ex
```diff
@@ -0,0 +1,256 @@
...
+
+  # nan_col is the column of the value to impute
+  defnp nan_euclidian(row, nan_col, potential_neighbor) do
+    {coordinates} = Nx.shape(row)
+
+    # minus nan column
+    coordinates = coordinates - 1
+
+    # inputes zeros in nan_col to calculate distance with squared_euclidean
+    new_row = Nx.indexed_put(row, Nx.new_axis(nan_col, 0), Nx.tensor(0))
```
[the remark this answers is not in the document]
srzeszut: I changed it but I don't know how to change this line 
    row_distances = Nx.iota({rows}, type: {:f, 32})
because i don't know what the type calculated distance will be at this point

srzeszut:
thanks for the review I apply suggested changes and leave some comments

lib/scholar/impute/knn_imputer.ex
josevalim: It is also broken there. :)

lib/scholar/impute/knn_imputer.ex
josevalim: This is correct as is. float and integer are types, nan is a value. :)

lib/scholar/impute/knn_imputter.ex
```diff
@@ -0,0 +1,265 @@
...
+    row_nan_count = Nx.sum(Nx.is_nan(x), axes: [1])
+    # row with only 1 non nan value is also considered as all nan row
+    all_nan_rows =
+      Nx.select(Nx.greater_equal(row_nan_count, cols - 1), Nx.tensor(1), Nx.tensor(0))
+
+    all_nan_rows_count = Nx.sum(all_nan_rows)
+
+    if num_neighbors > rows - 1 - Nx.to_number(all_nan_rows_count) do
+      raise ArgumentError,
+            "Number of neighbors rows must be less than number valid of rows - 1 (valid row is row with more than 1 non nan value)"
```
josevalim, line 104: error messages start in lowercase. :)

```suggestion
            "number of neighbors rows must be less than number valid of rows - 1 (valid row is row with more than 1 non nan value)"
```

lib/scholar/impute/knn_imputter.ex
```diff
@@ -0,0 +1,265 @@
...
+    {rows, cols} = Nx.shape(x)
+
+    row_nan_count = Nx.sum(Nx.is_nan(x), axes: [1])
+    # row with only 1 non nan value is also considered as all nan row
+    all_nan_rows =
+      Nx.select(Nx.greater_equal(row_nan_count, cols - 1), Nx.tensor(1), Nx.tensor(0))
+
+    all_nan_rows_count = Nx.sum(all_nan_rows)
+
+    if num_neighbors > rows - 1 - Nx.to_number(all_nan_rows_count) do
```
josevalim, line 102: Can you please add some tests? In particular, please add a test where you call jit this function and then you call it: `Nx.Defn.jit(...).(arg1, arg2)`. It should reveal some errors around here. :)

lib/scholar/impute/knn_imputer.ex
josevalim: I have fixed it there: c024c5bc18dbbce4c1365adbbbb9cc26fa2738c5

lib/scholar/impute/knn_imputter.ex
srzeszut: I added tests and checked it. I removed those checks and added them in the description

lib/scholar/impute/knn_imputter.ex
```diff
@@ -0,0 +1,247 @@
...
+  ]
+
+  @opts_schema NimbleOptions.new!(opts_schema)
+
+  @doc """
+  Imputter for completing missing values using k-Nearest Neighbors.
+
+  Preconditions:
+    * `number_of_neighbors` is a positive integer.
+    *  number of neighbors must be less than number valid of rows - 1 (valid row is row with more than 1 non nan value) otherwise it is better to use simple imputter
```
josevalim, line 39: Please try to break this long line :)

lib/scholar/impute/knn_imputter.ex
```diff
@@ -0,0 +1,247 @@
...
+    end
+
+    x =
+      if opts[:missing_values] != :nan,
+        do: Nx.select(Nx.equal(x, opts[:missing_values]), Nx.Constants.nan(), x),
+        else: x
+
+    num_neighbors = opts[:number_of_neighbors]
+
+    placeholder_value = Nx.Constants.nan() |> Nx.tensor()
```
josevalim, line 90: Can we move these three lines inside `knn_impute`, so it is all in `defn`? Btw, you can create a nan tensor with `Nx.tensor(:nan)`. :)

test/scholar/impute/knn_imputter_test.exs
```diff
@@ -0,0 +1,128 @@
...
+      x = Nx.tensor([1, 2, 2, 3])
+
+      assert_raise ArgumentError,
+                   "Wrong input rank. Expected: 2, got: 1",
+                   fn ->
+                     KNNImputter.fit(x, missing_values: 1, number_of_neighbors: 2)
+                   end
+    end
+
+    test "Invalid n_neighbors value" do
```
josevalim, line 116: Test names start in lowercase :)

```suggestion
    test "invalid impute rank" do
      x = Nx.tensor([1, 2, 2, 3])

      assert_raise ArgumentError,
                   "Wrong input rank. Expected: 2, got: 1",
                   fn ->
                     KNNImputter.fit(x, missing_values: 1, number_of_neighbors: 2)
                   end
    end

    test "invalid n_neighbors value" do
```

lib/scholar/impute/knn_imputter.ex
```diff
@@ -0,0 +1,247 @@
...
+    missing_values: [
+      type: {:or, [:float, :integer, {:in, [:nan]}]},
+      default: :nan,
+      doc: ~S"""
+      The placeholder for the missing values. All occurrences of `:missing_values` will be imputed.
+
+      The default value expects there are no NaNs in the input tensor.
+      """
+    ],
+    number_of_neighbors: [
```
krstopro, line 25: I would suggest changing this to `num_neighbors` to be consistent with the rest of `Scholar`.

krstopro (commented):
First review. Some features we might wanna have:
- Make k-NN algorithm configurable.
- Make the metric configurable.

You can leave these for another pull request. Have a look at e.g. `KNNClassifier` how it is done over there.
I should have another look tonight.

lib/scholar/impute/knn_imputter.ex
srzeszut: when I moved  it it looks like Nx.equal(x, opts[:missing_values]) this is not working properly because
it changes everything to nan

josevalim: cool, thank you!

krstopro (commented):
Several minor comments for now. I have to go through the code at least once more as I don't exactly understand the logic here.

lib/scholar/impute/knn_imputter.ex
polvalente, line 39:
```suggestion
    *  The number of neighbors must be less than the number of valid rows - 1.
       A valid row is a row with more than 1 non-NaN values. Otherwise it is better to use a simpler imputer.
```

lib/scholar/impute/knn_imputter.ex
```diff
@@ -0,0 +1,247 @@
...
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+    The function returns a struct with the following parameters:
+
+    * `:missing_values` - the same value as in `:missing_values`
+
+    * `:statistics` - The imputation fill value for each feature. Computing statistics can result in
+    [`Nx.Constant.nan/0`](https://hexdocs.pm/nx/Nx.Constants.html#nan/0) values.
```
polvalente, line 53:
```suggestion
    [`Nx.Constants.nan/0`](https://hexdocs.pm/nx/Nx.Constants.html#nan/0) values.
```

polvalente: Do you need the explicit linking in hexdoc?

lib/scholar/impute/knn_imputter.ex
```diff
@@ -0,0 +1,247 @@
...
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+    The function returns a struct with the following parameters:
+
+    * `:missing_values` - the same value as in `:missing_values`
```
polvalente, line 50:
```suggestion
    * `:missing_values` - the same value as in the `:missing_values` option
```

lib/scholar/impute/knn_imputter.ex
```diff
@@ -0,0 +1,247 @@
...
+
+    input_rank = Nx.rank(x)
+
+    if input_rank != 2 do
+      raise ArgumentError, "Wrong input rank. Expected: 2, got: #{inspect(input_rank)}"
+    end
+
+    x =
+      if opts[:missing_values] != :nan,
+        do: Nx.select(Nx.equal(x, opts[:missing_values]), Nx.Constants.nan(), x),
```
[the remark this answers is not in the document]
polvalente: This is a deftransform, so `Nx.equal` is the proper function. `==` will be `Elixir.Kernel.==`

lib/scholar/impute/knn_imputter.ex
polvalente, line 90:
```suggestion
    placeholder_value = Nx.Constants.nan()
```

polvalente: you probably want to pass the input type here to avoid upcasts

lib/scholar/impute/knn_imputter.ex
```diff
@@ -0,0 +1,247 @@
...
+  """
+  import Nx.Defn
+  import Scholar.Metrics.Distance
+
+  @derive {Nx.Container, keep: [:missing_values], containers: [:statistics]}
+  defstruct [:statistics, :missing_values]
+
+  opts_schema = [
+    missing_values: [
+      type: {:or, [:float, :integer, {:in, [:nan]}]},
```
polvalente, line 17:
```suggestion
      type: {:or, [:float, :integer, {:in, [:nan]}]},
```

I believe this should allow :infinity and :neg_infinity too for completeness

lib/scholar/impute/knn_imputter.ex
```diff
@@ -0,0 +1,247 @@
...
+            if mask[row][col] > 0 do
+              {rows, cols} = Nx.shape(x)
+
+              neighbor_avg =
+                calculate_knn(x, row, col, rows: rows, num_neighbors: opts[:num_neighbors])
+
+              indices =
+                [Nx.stack(row), Nx.stack(col)]
+                |> Nx.concatenate()
+                |> Nx.stack()
```
polvalente, line 146:
```suggestion
              indices = Nx.stack([row, col]) |> Nx.reshape({1, 2})
```

If I read the code correctly, row and col are scalars and this should yield the same result

lib/scholar/impute/knn_imputter.ex
```diff
@@ -0,0 +1,247 @@
...
+
+              values_to_impute = Nx.indexed_put(values_to_impute, indices, Nx.stack(neighbor_avg))
```
polvalente, line 148:
```suggestion
              values_to_impute = Nx.put_slice(values_to_impute, [row, col], Nx.reshape(neighbor_avg, {1, 1}))
```

I think this is even simpler

srzeszut:
Thanks for all the comments, I applied your suggested changes to the code.

josevalim (approved):

msluszniak (approved):

josevalim:

PR #305, opened 2024-10-27
TruncatedSVD fix

norm4nn:
Hey, I found a bug in the recently added TSVD (I was the author 😬), which caused the entire TSVD not to work. The test I added is failing on main branch, but it should indeed pass. Sorry bout that!


josevalim:

PR #306, opened 2024-10-30
Add BernoulliNB and Binarizer

srzeszut:
I am adding BernoulliNB and tests. I needed a binarizer in it so I added it with tests to preprocessing.

### DIFF
--- a/lib/scholar/naive_bayes/bernoulli.ex
+++ b/lib/scholar/naive_bayes/bernoulli.ex
@@ -1,0 +1,407 @@
+defmodule Scholar.NaiveBayes.Bernoulli do
+  @moduledoc """
+  Naive Bayes classifier for multivariate Bernoulli models.
+  Like MultinomialNB, this classifier is suitable for discrete data. The
+  difference is that while MultinomialNB works with occurrence counts,
+  BernoulliNB is designed for binary/boolean features.
+  """
+  import Nx.Defn
+  import Scholar.Shared
+
+  @derive {Nx.Container,
+           containers: [
+             :feature_count,
+             :class_count,
+             :class_log_priors,
+             :feature_log_probability
+           ]}
+  defstruct [:feature_count, :class_count, :class_log_priors, :feature_log_probability]
+
+  opts_schema = [
+    num_classes: [
+      type: :pos_integer,
+      required: true,
+      doc: ~S"""
+      Number of different classes used in training.
+      """
+    ],
+    alpha: [
+      type: {:or, [:float, {:list, :float}]},
+      default: 1.0,
+      doc: ~S"""
+      Additive (Laplace/Lidstone) smoothing parameter
+      (set alpha to 0.0 and force_alpha to true, for no smoothing).
+      """
+    ],
+    force_alpha: [
+      type: :boolean,
+      default: true,
+      doc: ~S"""
+      If `false` and alpha is less than 1e-10, it will set alpha to
+      1e-10. If `true`, alpha will remain unchanged. This may cause
+      numerical errors if alpha is too close to 0.
+      """
+    ],
+    binarize: [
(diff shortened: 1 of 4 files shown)


lib/scholar/naive_bayes/bernoulli.ex
```diff
@@ -0,0 +1,407 @@
...
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(opts_schema)
+
+  @doc """
+  Fits a naive Bayes model. The function assumes that the targets `y` are integers
+  between 0 and `num_classes` - 1 (inclusive). Otherwise, those samples will not
+  contribute to `class_count`.
+  ## Options
```
josevalim, line 83: Please remember to add new lines before and after each section header. Before lists and so on. :)

lib/scholar/naive_bayes/bernoulli.ex
```diff
@@ -0,0 +1,407 @@
...
+            """
+    end
+
+    opts = NimbleOptions.validate!(opts, @opts_schema)
+    type = to_float_type(x)
+
+    x_binarize =
+      if opts[:binarize] != nil,
+        do: Scholar.Preprocessing.Binarizer.fit_transform(x, threshold: opts[:binarize]),
+        else: x
```
josevalim, line 180: Try to move as much code as possible inside `fit_n`. For example, could this be moved there?

josevalim: I believe `opts[:binarize] != nil` inside `defn` will want to treat those as tensors. Perhaps this may work:

```elixir
case opts[:binarize] do
  nil -> x
  binarize ->Scholar.Preprocessing.Binarizer.fit_transform(x, threshold: binarize)
end
```

If it doesn't work we can leave it as is :)

srzeszut: it worked, Thanks!

krstopro:
I would suggest removing the binarizer from `fit` and assuming that the features are already binarized. You should mention this in the docs.

I will performed a more detailed review somewhere this week.

lib/scholar/naive_bayes/bernoulli.ex
```diff
@@ -0,0 +1,432 @@
...
+  defn predict_log_probability(%__MODULE__{} = model, x) do
+    check_dim(x, Nx.axis_size(model.feature_count, 1))
+    jll = joint_log_likelihood(model, x)
+
+    log_proba_x =
+      jll
+      |> Nx.logsumexp(axes: [1])
+      |> Nx.new_axis(1)
+      |> Nx.broadcast(jll)
+
```
polvalente, line 360: Is the explicit broadcast necessary?

lib/scholar/preprocessing/binarizer.ex
```diff
@@ -0,0 +1,56 @@
...
+        ]
+      >
+  """
+  deftransform fit_transform(tensor, opts \\ []) do
+    binarize_n(tensor, NimbleOptions.validate!(opts, @binarize_schema))
+  end
+
+  defnp binarize_n(tensor, opts) do
+    threshold = opts[:threshold]
+    Nx.select(Nx.greater(tensor, threshold), 1, 0)
```
polvalente, line 54:
```suggestion
    tensor > threshold
```

msluszniak (approved):

lib/scholar/preprocessing/binarizer.ex
srzeszut: I will change it

lib/scholar/naive_bayes/bernoulli.ex
srzeszut: I changed it

josevalim:

PR #307, opened 2024-10-30
Partial Least Square SVD

norm4nn:
Added Partial Least Square SVD transformer and tests for it. 

[Scikit-learn context](https://scikit-learn.org/stable/modules/generated/sklearn.cross_decomposition.PLSSVD.html#sklearn.cross_decomposition.PLSSVD)

### DIFF
--- a/lib/scholar/cross_decomposition/pls_svd.ex
+++ b/lib/scholar/cross_decomposition/pls_svd.ex
@@ -1,0 +1,117 @@
+defmodule Scholar.CrossDecomposition.PLSSVD do
+  @moduledoc """
+
+  """
+  import Nx.Defn
+  import Scholar.Shared
+
+  opts_schema = [
+    num_components: [
+      default: 2,
+      type: :pos_integer,
+      doc: "The number of components to keep. Should be in `[1,
+        min(n_samples, n_features, n_targets)]`."
+    ],
+    scale: [
+      default: true,
+      type: :boolean,
+      doc: "Whether to scale `x` and `x`."
+    ],
+  ]
+
+  @opts_schema NimbleOptions.new!(opts_schema)
+
+  @doc """
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+
+  ## Examples
+
+  """
+
+  deftransform fit(x, y opts \\ []) do
+    fit_n(x, y,  NimbleOptions.validate!(opts, @opts_schema))
+  end
+
+  defnp fit_n(x, y, opts) do
+    {x, y} = check_x_y(x, y, opts)
+    num_components = opts[:num_components]
+    {x, y, x_mean, y_mean, x_std, y_std} = center_scale_xy(x, y, opts)
+  end
+
+  defnp check_x_y(x, y, opts) do
+    y =
+      case Nx.shape(y) do
+        {n} -> Nx.reshape(y, {n, 1})
+        _ -> y
+      end
+
+    num_components = opts[:num_components]
+    {num_samples, num_features} = Nx.shape(x)
+    {num_samples_y, num_targets} = Nx.axis_shape(y, 0)
+
+    cond do
+      num_samples != num_samples_y ->
+        raise ArgumentError,
+              """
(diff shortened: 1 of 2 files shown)


lib/scholar/cross_decomposition/pls_svd.ex
```diff
@@ -0,0 +1,357 @@
...
+    x_scores = Nx.dot(xr, x_weights)
+
+    yr = (y - y_mean) / y_std
+    y_scores = Nx.dot(yr, y_weights)
+    {x_scores, y_scores}
+  end
+
+  @doc """
+    Learn and apply the dimensionality reduction.
+    Takes as arguments:
```
josevalim, line 225:
```suggestion
  Learn and apply the dimensionality reduction.
  
  The arguments are:
```

lib/scholar/cross_decomposition/pls_svd.ex
```diff
@@ -0,0 +1,357 @@
...
+    
+    * `y` - targets, `{num_samples, num_targets}` shaped `y` tensor
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+    Returns tuple with transformed data `{x_transformed, y_transformed}` where:
```
josevalim, line 237:
```suggestion
  Returns tuple with transformed data `{x_transformed, y_transformed}` where:
```

josevalim (approved):

lib/scholar/cross_decomposition/pls_svd.ex
```diff
@@ -0,0 +1,356 @@
...
+  end
+
+  defnp fit_n(x, y, opts) do
+    {x, y} = check_x_y(x, y, opts)
+    num_components = opts[:num_components]
+    {x, y, x_mean, y_mean, x_std, y_std} = center_scale_x_y(x, y, opts)
+
+    c =
+      Nx.transpose(x)
+      |> Nx.dot(y)
```
polvalente, line 117:
```suggestion
    c = Nx.dot(x, [1], y, [1])
```

polvalente: @josevalim I think we could add an in-place rewrite in Nx.Defn.Expr that checks if the immediate child arg is Nx.transpose, and aplies the permutation to the contract and axes for dot/2.

This is a really common pattern that I tend to find in reviews

lib/scholar/cross_decomposition/pls_svd.ex
```diff
@@ -0,0 +1,356 @@
...
+          y,
+          opts
+        ) do
+    {x, y} = check_x_y(x, y, opts)
+
+    xr = (x - x_mean) / x_std
+    x_scores = Nx.dot(xr, x_weights)
+
+    yr = (y - y_mean) / y_std
+    y_scores = Nx.dot(yr, y_weights)
```
polvalente, line 217:
```suggestion
    y_scores = Nx.dot(yr, [1], y_weights, [1])
```

if you remove the Nx.transpose from the definition in `y_weights` you can use this suggestion to achieve the same result

norm4nn: Is it more efficient to call `Nx.dot` like you suggest? I’d like to understand why you're insisting on doing it this way.

polvalente: It is because you avoid the explicit transpose when producing the tensor.

Nx.dot(a, Nx.transpose(b))

is equivalent to

Nx.dot(a, [1], b, [1])

lib/scholar/cross_decomposition/pls_svd.ex
```diff
@@ -0,0 +1,353 @@
...
+              """
+
+      true ->
+        nil
+    end
+
+    {x, y}
+  end
+
+  defnp center_scale_x_y(x, y, opts) do
```
msluszniak, line 329: In this function, all the ops are the same for both x and y. I think it will be better to write this to a single tensor and apply function twice

polvalente (approved):

msluszniak (approved):

josevalim:

PR #304, opened 2024-10-26
Added Ledoit Wolf shrinkage covariance estimator

norm4nn:
Added Ledoit Wolf shrinkage covariance matrix estimator. 

[Scikit-learn docs](https://scikit-learn.org/1.5/modules/generated/sklearn.covariance.LedoitWolf.html)

PR #309, opened 2024-11-08
Added ShrunkCovariance

norm4nn:
What has been added:
- ShrunkCovariance module ([scikit-learn context](https://scikit-learn.org/stable/modules/generated/sklearn.covariance.ShrunkCovariance.html)) 
- Covariance.Utilites module with common functions
- tests and docs for ShrunkCovariance

Following up on #304: I added the `?` sign to the `assume_centered` option, but I'm unsure about replacing `empirical_covariance/1` with `Nx.covariance/3` because tests don’t pass when I make that change. I believe the tests should stay as they are, since I’ve already verified the results against scikit-learn.


lib/scholar/covariance/shrunk_covariance.ex
```diff
@@ -0,0 +1,120 @@
...
+  defnp shrunk_covariance(emp_cov, shrinkage) do
+    num_features = Nx.axis_size(emp_cov, 1)
+    shrunk_cov = (1.0 - shrinkage) * emp_cov
+    emp_cov_trace = Scholar.Covariance.Utils.trace(emp_cov)
+    mu = Nx.sum(emp_cov_trace) / num_features
+
+    mask = Nx.iota(Nx.shape(shrunk_cov))
+    selector = Nx.remainder(mask, num_features + 1) == 0
+
+    Nx.select(selector, shrunk_cov + shrinkage * mu, shrunk_cov)
```
msluszniak, line 118:
```suggestion
    shrunk_cov + shrinkage * mu * selector
```

msluszniak: This is less readable as `Nx.select` indicates the explicit intention. But these vectorized ops might be a little bit faster. I don't have a strong preference here, so If you feel that `Nx.select` fits better here and reflect your coding style, then feel free to ignore it.

lib/scholar/covariance/shrunk_covariance.ex
```diff
@@ -0,0 +1,121 @@
...
+          [0.1276099532842636, 0.41163527965545654]
+        ]
+      >
+      iex> model.location
+      #Nx.Tensor<
+        f32[2]
+        [0.18202415108680725, -0.09216632694005966]
+      >
+
+
```
polvalente, line 80:
```suggestion
```

polvalente (approved):

lib/scholar/covariance/utils.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Scholar.Covariance.Utils do
+  @moduledoc false
+  import Nx.Defn
+  require Nx
+
+  defn center(x, assume_centered \\ false) do
```
krstopro, line 6:
```suggestion
defn center(x, assume_centered? \\ false) do
```

lib/scholar/covariance/utils.ex
```diff
@@ -0,0 +1,39 @@
...
+      if assume_centered do
+        0
+      else
+        Nx.mean(x, axes: [0])
+      end
+
+    {x - location, location}
+  end
+
+  defn empirical_covariance(x) do
```
krstopro, line 23: You should be able to use `Nx.covariance/2` instead.

krstopro (commented):
Few nitpicks.

lib/scholar/covariance/utils.ex
msluszniak: I don't think so, look at the PR description

krstopro: Right, I forgot to ask: @norm4nn did you try setting `ddof: 0` when calling `Nx.covariance/2`?

krstopro: Sorry, `ddof: 0` is default. It's strange that the tests are failing, because `Nx.covariance/2` does exactly what is implemented here. Could it be the case that the data in your test is not centered and you are using `Nx.covariance/2`?

norm4nn: Yeah, you are right! This is the case, I will fix this on Thursday.

lib/scholar/covariance/ledoit_wolf.ex
```diff
@@ -182,9 +149,9 @@ defmodule Scholar.Covariance.LedoitWolf do
 
   defnp ledoit_wolf_shrinkage_complex(x) do
     {num_samples, num_features} = Nx.shape(x)
-    emp_cov = empirical_covariance(x)
+    emp_cov = Nx.covariance(x)
```
krstopro, line 152: Is `x` centered here? If yes, maybe reverting it to `empirical_covariance` might be better. 😅

norm4nn: Hmm, I don't think `empirical_covariance/1` was centering `x,` if that's what you're asking about. However, I'm still a bit confused as to why, when I used `empirical_covariance/1` instead of `Nx.covariance/2,` the `Covariance.*.fit/2` functions produced the same output as the scikit-learn versions when I passed a non-centered `x` and set `assume_centered?` to `true.` While I feel this case is invalid and can't think of any practical use case for passing such arguments, it still feels odd that the output in this edge case differs when using `Nx.covariance/2` compared to the scikit-learn version. It seems `empirical_covariance` might not operate exactly the same way as `Nx.covariance/2` , so I also like the idea of reverting it to `empirical_covariance`.

krstopro: Lemme have a look.

polvalente: I think you need 
```suggestion
    emp_cov = Nx.covariance(x, Nx.reshape(0, {1}))
```

for the code to be totally equivalent.

empirical_covariance was blindly assuming the mean is 0

krstopro: > empirical_covariance was blindly assuming the mean is 0

Yes, because the data should be centered before. That is why using `empirical_covariance` might be better idea. Like this the data won't be centered twice.

norm4nn: So you want me to change it back to `empirical_covariance` version?

josevalim: @norm4nn let's revert to `empirical_covariance` for now so we can merge this and get unstuck. :)

msluszniak (approved):

lib/scholar/covariance/ledoit_wolf.ex
norm4nn: Sure, done

josevalim:

PR #311, opened 2024-11-16
Update to latest Nx

### LINKED ISSUE #310: Bug Scholar.Manifold.MDS.fit
```elixir
Mix.install([:nx, :scholar])

defmodule T do
  def fit do
     key = Nx.Random.key(42)

     Nx.tensor([1, 2, 3, 4])
     |> Nx.reshape({2, 2})
     |> Scholar.Manifold.MDS.fit(key: key)
  end
end
```

Not happy with the type of `i`  ?

https://github.com/elixir-nx/scholar/blob/473060fc9bac44b377d1b5e32786325af3de371b/lib/scholar/linear/isotonic_regression.ex#L477

```
** (CompileError) deps/scholar/lib/scholar/linear/isotonic_regression.ex:469: the do-block in while must return tensors with the same shape, type, and names as the initial arguments.

{#Nx.Tensor<
   f32[1]
 >, #Nx.Tensor<
   f32[1]
 >,
 <<<<< Body (do-block) <<<<<
 #Nx.Tensor<
   s64
 >
 ==========
 #Nx.Tensor<
   s32
 >
 >>>>>     Initial     >>>>>
 , #Nx.Tensor<
   s64[1]
 >, #Nx.Tensor<
   f32
 >, #Nx.Tensor<
   f32
 >, #Nx.Tensor<
   f32
 >, #Nx.Tensor<
   s64
 >, #Nx.Tensor<
   s32
 >, #Nx.Tensor<
   s32
 >}
```

josevalim:
Nx defaults to 32 precision everywhere, hence the changes.

Closes #310.


lib/scholar/manifold/trimap.ex
```diff
@@ -188,13 +188,19 @@ defmodule Scholar.Manifold.Trimap do
     final_samples = Nx.broadcast(Nx.s64(0), shape)
 
     {final_samples, key, _, _} =
-      while {final_samples, key, rejects, i = Nx.s64(0)}, i < elem(shape, 0) do
-        {samples, key} = Nx.Random.randint(key, 0, opts[:maxval], shape: {elem(shape, 1)})
+      while {final_samples, key, rejects, i = Nx.u32(0)}, i < elem(shape, 0) do
+        # TODO: See if we can relax the samples to u32
```
josevalim, line 192: @msluszniak I kept the type here as s64 because if we change the type, we change the samples, which changes the result of the tests. Can you please tell me if you copied the tests from Python's trimap? Because if not, it is ok for us to change the values here and then update the tests. :)

josevalim:
I will merge this. If you folks have feedback, I will tackle them later.

josevalim:

PR #313, opened 2024-12-22
Add CategoricalNB

ksew1:
Add CategoricalNB :)


krstopro:
Hi @ksew1 and thanks for the pull request. Looks great on the first look and I will be having a detailed review soon.

lib/scholar/naive_bayes/categorical.ex
```diff
@@ -0,0 +1,556 @@
...
+  end
+
+  defnp joint_log_likelihood(
+          %__MODULE__{
+            feature_log_probability: feature_log_probability,
+            class_log_priors: class_log_priors
+          },
+          x
+        ) do
+    {_, _, _, jll} =
```
msluszniak, line 479: As you don't use the first three variables, you can wrap them in a tuple, and then use more convenient pattern match `{_, jil}`

lib/scholar/naive_bayes/categorical.ex
```diff
@@ -0,0 +1,556 @@
...
+          Nx.slice_along_axis(feature_log_probability, i, 1, axis: 0)
+          |> Nx.squeeze()
+          |> Nx.take(indices, axis: 1)
+          |> Nx.transpose()
+          |> Nx.add(jll)
+
+        {i + 1, feature_log_probability, x, jll}
+      end
+
+    total_jll = jll + class_log_priors
```
msluszniak, line 494: This can be simply returned

lib/scholar/naive_bayes/categorical.ex
```diff
@@ -0,0 +1,556 @@
...
+    total_jll
+  end
+
+  defnp count(x, y_weighted, num_features, num_classes, num_categories, num_samples) do
+    class_count = Nx.sum(y_weighted, axes: [0])
+
+    feature_count = Nx.broadcast(0.0, {num_features, num_classes, num_categories})
+
+    {_, _, _, _, feature_count} =
```
msluszniak, line 503: The same as above

msluszniak (approved):
I've Dropped some minor comments, but LGMT. There are some part that are common for all of NB algorithms. I think that it will be cool to separate them into `utils.ex` inside `naive_bayes` dir. But it's not necessary as a scope of this PR.

ksew1:
> I've Dropped some minor comments, but LGMT. There are some part that are common for all of NB algorithms. I think that it will be cool to separate them into `utils.ex` inside `naive_bayes` dir. But it's not necessary as a scope of this PR.



I’ve addressed all the comments 😄. I agree that it would be a good idea to extract the common parts into a separate file. I’ll do this in a separate PR.

josevalim (approved):

krstopro:
Any reason why `num_categories` isn't a required option?

ksew1:
> Any reason why `num_categories` isn't a required option?

`num_categories` isn’t needed as an option since it’s just an implementation detail. We can easily calculate it from `min_categories`, which the user can already specify, so i think adding it would just be redundant

krstopro:
@josevalim @msluszniak I am still having a look. Looks very good, but please give me some time before merging. I might have some improvements to suggest (e.g. maybe the code can be vectorised; this could be valuable for @ksew1 to learn in case they don't know it already).

Happy New Year everyone!

krstopro:
@josevalim Let's merge this one; I might submit another PR in case I see a convenient way to vectorise parts of the procedure. I apologise for the delay, in particular to @ksew1.

josevalim:

PR #315, opened 2025-01-11
Add CountVectorizer

ksew1:
Hi I added `CountVectorizer` as it would be useful in pipeline with Naive Bayes algorithms that we already implemented (for example in spam detection)

### DIFF
--- a/lib/scholar/feature_extraction/count_vectorizer.ex
+++ b/lib/scholar/feature_extraction/count_vectorizer.ex
@@ -1,10 +1,10 @@
 defmodule Scholar.FeatureExtraction.CountVectorizer do
   @moduledoc """
-    A `CountVectorizer` converts a collection of text documents to a matrix of token counts.
+  A `CountVectorizer` converts a collection of text documents to a matrix of token counts.
 
-    Each row of the matrix corresponds to a document in the input corpus, and each column corresponds to a unique token from the vocabulary of the corpus.
+  Each row of the matrix corresponds to a document in the input corpus, and each column corresponds to a unique token from the vocabulary of the corpus.
 
-    Supports also already indexed tensors.
+  Supports also already indexed tensors.
   """
   import Nx.Defn
 
@@ -35,8 +35,11 @@ defmodule Scholar.FeatureExtraction.CountVectorizer do
   * creation of a count tensor
 
   ## Options
+
   #{NimbleOptions.docs(@binarize_schema)}
+
   ## Examples
+
       iex> corpus = ["Elixir is amazing!", "Elixir provides great tools."]
       iex> Scholar.FeatureExtraction.CountVectorizer.fit_transform(corpus)
       %Scholar.FeatureExtraction.CountVectorizer{
--- a/lib/scholar/feature_extraction/count_vectorizer.ex
+++ b/lib/scholar/feature_extraction/count_vectorizer.ex
@@ -4,6 +4,18 @@ defmodule Scholar.FeatureExtraction.CountVectorizer do
   """
   import Nx.Defn
 
+  opts_schema = [
+    max_token_id: [
+      type: :pos_integer,
+      required: true,
+      doc: ~S"""
+      Maximum token id in the input tensor.
+      """
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(opts_schema)
+
   @doc """
   Generates a count matrix where each row corresponds to a document in the input corpus, and each column corresponds to a unique token in the vocabulary of the corpus.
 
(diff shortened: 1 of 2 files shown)


lib/scholar/feature_extraction/count_vectorizer.ex
```diff
@@ -0,0 +1,147 @@
+defmodule Scholar.FeatureExtraction.CountVectorizer do
+  @moduledoc """
+    A `CountVectorizer` converts a collection of text documents to a matrix of token counts.
```
josevalim, line 3: Don't indent the docs by two spaces. They should be aligned with the closing `"""`, as that's the one that controls indentation. :)

lib/scholar/feature_extraction/count_vectorizer.ex
```diff
@@ -0,0 +1,147 @@
...
+
+  This function performs:
+
+  * tokenization of input text (splitting by whitespace and removing punctuation)
+  * vocabulary construction
+  * creation of a count tensor
+
+  ## Options
+  #{NimbleOptions.docs(@binarize_schema)}
+  ## Examples
```
josevalim, line 39:
```suggestion
  ## Options

  #{NimbleOptions.docs(@binarize_schema)}

  ## Examples

```

lib/scholar/feature_extraction/count_vectorizer.ex
```diff
@@ -0,0 +1,147 @@
...
+      if opts[:indexed_tensor] do
+        {corpus, %{}}
+      else
+        preprocessed_corpus = preprocess(corpus)
+        vocabulary = create_vocabulary(preprocessed_corpus)
+        tensor = create_tensor(preprocessed_corpus, vocabulary)
+        {tensor, vocabulary}
+      end
+
+    max_index = tensor |> Nx.reduce_max() |> Nx.add(1) |> Nx.to_number()
```
josevalim, line 82: Under `jit` mode, you cannot convert a tensor to a number. You can try adding this test:

  fun = &Scholar.FeatureExtraction.CountVectorizer.fit_transform(&1, indexed_tensor: true)
  Nx.Defn.jit(fun).(Nx.tensor([[0, 1, 2], [1, 3, 4]])

The correct solution is to compute the default value for `max_index` inside `defn`.

josevalim (commented):
The `fit_transform` function currently has an issue in that it accepts either tensors (which can be JITed) or a list of strings, which cannot. Generally speaking, our `fit_transform` functions can be jitted, so this would be breaking the tradition.

It seems passing a list of strings is implementing rather a simple tokenization algorithm? I think we should either move it elsewhere OR simply remove the functionality from here. My suggestion would be to remove it :)

lib/scholar/feature_extraction/count_vectorizer.ex
ksew1: The problem is that it needs to be a number in order to create a tensor with this shape. Inside `defn`, I'm afraid it is not possible to dynamically obtain a number. We can make this a required option, so we do not need to use `Nx.to_number()`.

lib/scholar/feature_extraction/count_vectorizer.ex
ksew1: Fixed 😄

lib/scholar/feature_extraction/count_vectorizer.ex
ksew1: Fixed 😄

ksew1:
> The `fit_transform` function currently has an issue in that it accepts either tensors (which can be JITed) or a list of strings, which cannot. Generally speaking, our `fit_transform` functions can be jitted, so this would be breaking the tradition.
> 
> It seems passing a list of strings is implementing rather a simple tokenization algorithm? I think we should either move it elsewhere OR simply remove the functionality from here. My suggestion would be to remove it :)

Removed this feature, now it operates only on tensor

lib/scholar/feature_extraction/count_vectorizer.ex
josevalim: I see. So we should not use `deftransform`, because this certainly cannot be invoked inside a `defn`, as we don't yet support dynamic shapes. For now, it should be a regular `def` and maybe it should be called something else.

I believe this topic appeared in the past and we maybe discussed a possible contract for keeping code that doesn't work inside `defn` but I don't recall it right now. :) Maybe @msluszniak does?

msluszniak: Usually, if the shape was based on computations we either droped the options that force it (like percent of variation preserved in PCA) or develop some heuristics that roughly assesses the upper bound of the problematic shape, and then make computations on bigger tensor with some "padding".

msluszniak: Or make an option that is required via NimbleOptions

josevalim: Maybe an option is the way to go then and we could even have a helper function like `CountVectorizer.size(n)` that they could use to compute it.

ksew1: I made this required option and added helper function 😄

josevalim:

PR #320, opened 2025-06-15
RANSAC estimator

santiago-imelio:
Hello! This is work in progress but I would love to get some feedback on how this looks so far. 

Specifically on the `while` code, not sure if I'm passing things correctly (`loss_fn` and `min_samples` are being passed as tensor shapes because otherwise they turn into `Nx.Defn.Expr` 0_o).

Ideally, RANSAC could take any fit-predict model as base estimator, so I'd love to hear suggestions on how to go about implementing that (could be something similar to the `loss_fn` pattern, but not sure).

### DIFF
--- a/lib/scholar/linear/ransac_regression.ex
+++ b/lib/scholar/linear/ransac_regression.ex
@@ -11,26 +11,14 @@ defmodule RANSACRegression do
 
   import Nx.Defn
   alias Scholar.Metrics.Regression, as: Metrics
-
-  alias Scholar.Linear.{
-    LinearRegression,
-    PolynomialRegression
-  }
+  alias Scholar.Linear.PolynomialRegression
 
   @derive {Nx.Container, containers: [:model, :inliers_indices, :n_inliers]}
-  defstruct [:estimator, :model, :inliers_indices, :n_inliers]
+  defstruct [:model, :inliers_indices, :n_inliers]
 
-  @models [linear: 1, polynomial: 2]
   @losses [mae: 1, mse: 2]
 
   opts = [
-    estimator: [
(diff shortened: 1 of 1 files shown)


lib/scholar/linear/ransac_regression.ex
```diff
@@ -0,0 +1,159 @@
+defmodule RANSACRegression do
```
josevalim, line 1: Remember to namespace the module :)

lib/scholar/linear/ransac_regression.ex
```diff
@@ -0,0 +1,159 @@
...
+  Fits a robust regressor using RANSAC algorithm, using
+  a polynomial regression as base estimator.
+
+  #{NimbleOptions.docs(@opts_schema)}
+  """
+  def fit(x, y, opts) do
+    opts = NimbleOptions.validate!(opts, @opts_schema)
+
+    inliers_mask = fit_n(x, y, opts)
+    n_inliers = Nx.to_number(Nx.sum(inliers_mask))
```
josevalim, line 77: You want to avoid Nx.to_number because it means `fit` then cannot be called from another `defn`. The best way is to write some tests calling this function directly and also from another `defn`.

santiago-imelio: @josevalim Hey! thanks for the suggestions. I've been looking for a way to use `n_inliers` without converting to number but failed to do so. Since it is a `Nx.Defn.Expr`, I cannot use it for indexing. Any suggestion on how to approach this? The objective is to extract from `x` and `y` the inliers, meaning the indices from `inliers_mask` such that the value is `1`.

PR #329, opened 2026-01-11
Logistic regression updates

### LINKED ISSUE #301: Bug in Cross-Entropy Implementation inside `Scholar.Linear.LogisticRegression`?
I believe there is a reduction missing (probably `Nx.mean/2`) in the implementation of cross-entropy inside `Scholar.Linear.LogisticRegression`.
https://github.com/elixir-nx/scholar/blob/975938a4e57704ad2400aa89ec18761cc0e45ce4/lib/scholar/linear/logistic_regression.ex#L188
`xs` is a tensor of shape `{num_samples, num_features}`, `ys` is a tensor of shape `{num_samples, num_classes}` (one-hot encoded). Hence, the expression above evaluates to a tensor of shape `{num_samples}`. I don't know how exactly does `Polaris` handle vectors, but I believe a loss should be a scalar.

krstopro:
Changes:
- Fixes loss computation by adding average over sample losses
- Fixes convergence criteria
- Adds input validation in `predict` and `predict_proba`
- Polishes the code slightly
- Adds more unit tests

Fixes #301


lib/scholar/linear/logistic_regression.ex
```diff
@@ -139,11 +171,20 @@ defmodule Scholar.Linear.LogisticRegression do
...
+          x,
+          y,
+          w,
+          b,
+          alpha,
+          l1_ratio,
+          tol,
+          w_optimizer_state,
+          b_optimizer_state,
+          opts
```
josevalim: Btw, when you have too many arguments like this, you can wrap the arguments that don't change (are fixed in the loop) in a map and pass it down. Helps clean up while loops and Elixir signatures!

josevalim (approved):

lib/scholar/linear/logistic_regression.ex
krstopro: This works for tensors as well, right?
Apart from that, CI is failing. I am working on it.

krstopro: Sorry, I am not sure I understood. 😅

Would having a while loop as `{result, state} = while ...` where `state` is a map work? Is this something new in `Nx`?

josevalim: Yes, it works. It is quite old i think :)

krstopro:
Alright, this was way more work than I thought :)
I managed to remove the optimizer completely by explicitly computing the gradients and using line search for adapting the learning rate.
@josevalim If this gets merged I would suggest releasing a new version. Logistic regression is a fundamental algorithm and there have been important changes to others (e.g. DBSCAN).

josevalim:
It is your decision to merge it! I can ship a new version afterwards!

josevalim (approved):

krstopro:
> It is your decision to merge it! I can ship a new version afterwards!

Alright then, merged. :)

krstopro:
P.S. I am having issues compiling EXLA 0.9.2 on my M1 Mac. Perhaps we should bump the version?

josevalim:
Released (and feel free to bump Nx as well!)

PR #332, opened 2026-03-15
Fixing KaTeX in docs

akoutmos:
While going through the Scholar docs I noticed that the KaTeX expressions were not being rendered properly. My guess is that the event listener has changed and the handler was never being invoked. Either way I followed the ExDocs instructions and updated the `before_closing_body_tag` function:

Before:
image

After:
image


josevalim:

PR #335, opened 2026-07-11
Add monotonic cubic spline (PCHIP) interpolation

RicardoSantos-99:
Adds Scholar.Interpolation.MonotonicCubicSpline, a piecewise cubic Hermite interpolant whose knot derivatives are chosen with the Fritsch-Carlson method so the curve stays monotonic where the data is monotonic and does not overshoot around local extrema (equivalent to scipy.interpolate.PchipInterpolator).

The power-basis coefficient layout and predict/3 are shared with CubicSpline, so only the derivative computation differs. Reference values in the tests were taken from SciPy.

Closes #322


josevalim (approved):

josevalim:

PR #337, opened 2026-07-12
Add Kernel PCA decomposition

RicardoSantos-99:
Adds `Scholar.Decomposition.KernelPCA`, the kernel extension of PCA that runs
the decomposition in a reproducing kernel Hilbert space, capturing non-linear
structure that ordinary PCA cannot. Addresses the Kernel PCA item of #246.

Supports the `:linear`, `:poly`, `:rbf`, `:sigmoid` and `:cosine` kernels,
with `fit/2`, `transform/2` and `fit_transform/2` mirroring the existing PCA
API. The kernel matrix is double-centered, decomposed with `Nx.LinAlg.eigh`,
and the eigenvectors are sign-flipped for deterministic output.

A few numerical decisions worth calling out:

* The centered kernel is explicitly symmetrized before the decomposition,
  since centering can leave asymmetries of a few ULPs and `eigh` requires
  exact symmetry.
* `eigh` can stop before the eigenvalues fully converge while the
  eigenvectors are already accurate, so the eigenvalues are recomputed as
  the Rayleigh quotient of the renormalized eigenvectors. On a 25-sample
  dataset this brings them from a few percent off to within 1.0e-10 of
  scikit-learn (covered by a regression test).
* The `eigh` tolerance follows the input precision (f32/f64): a tolerance
  below the floating-point resolution makes the iteration accumulate
  rounding noise past convergence.
* Zero and negative eigenvalues (floating-point noise, or indefinite
  kernels such as sigmoid) are clipped and their components project to
  zero, as in scikit-learn.

## Verification

Results match `sklearn.decomposition.KernelPCA` on all five kernels for
`eigenvalues`, `fit_transform` and `transform`, including custom
gamma/degree, zero-norm rows with the cosine kernel, and
`num_components == num_samples`. f32 agrees to roughly 1.0e-4 and f64 to
roughly 1.0e-5 or better. The reference values in the tests were generated
with scikit-learn 1.6.1.


josevalim (approved):

josevalim:

Issue #135, opened 2023-07-10
Add clustering algorithms


josevalim:
* [x] K-means
* [x] Affinity Propagation
* [x] Mean Shift
* [x] Spectral clustering
* [x] Hierarchical clustering
* [x] DBSCAN
* [x] HDBSCAN
* [x] OPTICS

grevgeny:
I'm interested in contributing to the project and would like to begin by implementing the Mean Shift algorithm.

josevalim:
Please go ahead @grevgeny! You can see the PRs from other clustering algorithms, to get an idea how of these contributions usually work (and the project goals).

paulsullivanjr:
I'm going to work on implementing BIRCH

RicardoSantos-99:
Sorry, the Closes #135 in my spectral clustering squash commit closed this by mistake. It's a tracking issue and Mean Shift, HDBSCAN and BIRCH are still open, so it was closed in error. Could we reopen it? @josevalim @krstopro

paulsullivanjr:
I committed to doing BIRCH a while back and life got in the way. I can get started on it in the next few days if that works.

RicardoSantos-99:
@josevalim the Spectral clustering is done, it landed in [#339](https://github.com/elixir-nx/scholar/pull/339). Mean Shift and HDBSCAN are the only ones left (I cannot edit issue)

RicardoSantos-99:
With Mean Shift and HDBSCAN merged, every algorithm on this list is now implemented.

CC: @josevalim

josevalim:
:green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart: thanks everyone!

PR #339, opened 2026-07-13
Add spectral clustering

RicardoSantos-99:
Add spectral clustering

Implements Scholar.Cluster.SpectralClustering via the normalized graph
Laplacian. The affinity matrix is built with an RBF kernel (gamma
configurable) or accepted as precomputed. Self-loops are discarded before
computing degrees, matching scipy's csgraph convention. Isolated vertices
keep a unit diagonal, matching scikit-learn's _set_diag behavior.

The spectral embedding takes the k eigenvectors of the smallest eigenvalues
of the Laplacian (eigh output is not sorted, so an explicit argsort is
needed), rescales them as u = D^-1/2 v per the relaxed Ncut formulation,
and applies a deterministic sign flip. K-means then runs on this embedding.

The struct exposes both labels and the embedding itself.

Related to #135.

### DIFF
--- a/lib/scholar/cluster/spectral_clustering.ex
+++ b/lib/scholar/cluster/spectral_clustering.ex
@@ -1,0 +1,210 @@
+defmodule Scholar.Cluster.SpectralClustering do
+  @moduledoc """
+  Spectral clustering.
+
+  Applies clustering to a projection of the normalized graph Laplacian [1].
+  Instead of clustering the points in their original space, the algorithm
+  builds an affinity (similarity) matrix between all pairs of samples,
+  computes the eigenvectors associated with the smallest eigenvalues of its
+  normalized graph Laplacian, and runs k-means on that spectral embedding.
+
+  In practice spectral clustering is very useful when the individual
+  clusters are non-convex (e.g. nested circles), a setting where plain
+  k-means on the original space performs poorly [2].
+
+  The eigendecomposition of the dense `{num_samples, num_samples}` Laplacian
+  gives a time complexity of $O(N^3)$ and a memory complexity of $O(N^2)$,
+  where $N$ is the number of samples.
+
+  References:
+
+  * [1] - [A Tutorial on Spectral Clustering, Ulrike von Luxburg](https://arxiv.org/abs/0711.0189)
+  * [2] - [On Spectral Clustering: Analysis and an algorithm, Ng, Jordan & Weiss](https://proceedings.neurips.cc/paper_files/paper/2001/file/801272ee79cfde7fa5960571fee36b9b-Paper.pdf)
+  """
+  import Nx.Defn
+
+  @derive {Nx.Container, containers: [:labels, :embedding]}
+  defstruct [:labels, :embedding]
+
+  opts = [
+    num_clusters: [
+      required: true,
+      type: :pos_integer,
+      doc: "The number of clusters to form, as well as the dimension of the spectral embedding."
+    ],
+    affinity: [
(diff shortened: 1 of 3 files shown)


krstopro:
I remember giving this one a try long time ago, but one of the issues was that there were no efficient solvers for top-k eigenvalues in Nx. Is it still the case?

RicardoSantos-99:
> I remember giving this one a try long time ago, but one of the issues was that there were no efficient solvers for top-k eigenvalues in Nx. Is it still the case?

Yeah, I think that's still the case. As far as I can tell Nx only has the
dense eigh, I couldn't find any partial or iterative solver in there.

So right now this just runs the full eigh and takes the k smallest
eigenvectors, which ends up O(N^3) / O(N^2) (I tried to be upfront about
that in the moduledoc). My thinking was that it matches sklearn's dense
eigen_solver and the values line up with it, and it felt in line with the
other exact/dense stuff already in Scholar like exact t-SNE. But I might be
missing something.

Would you be ok merging the dense version for now and revisiting if a
partial eigensolver ever lands in Nx? Or do you think it's better to hold
off until there's a proper top-k solver? Happy to go either way.

krstopro:
> > I remember giving this one a try long time ago, but one of the issues was that there were no efficient solvers for top-k eigenvalues in Nx. Is it still the case?
> 
> Yeah, I think that's still the case. As far as I can tell Nx only has the dense eigh, I couldn't find any partial or iterative solver in there.
> 
> So right now this just runs the full eigh and takes the k smallest eigenvectors, which ends up O(N^3) / O(N^2) (I tried to be upfront about that in the moduledoc). My thinking was that it matches sklearn's dense eigen_solver and the values line up with it, and it felt in line with the other exact/dense stuff already in Scholar like exact t-SNE. But I might be missing something.
> 
> Would you be ok merging the dense version for now and revisiting if a partial eigensolver ever lands in Nx? Or do you think it's better to hold off until there's a proper top-k solver? Happy to go either way.

Yes, this seems fine for the first version, especially since you documented the running time.

lib/scholar/cluster/spectral_clustering.ex
```diff
@@ -0,0 +1,210 @@
...
+        eigh_eps: eigh_eps
+      )
+
+    kmeans =
+      Scholar.Cluster.KMeans.fit(embedding,
+        num_clusters: opts[:num_clusters],
+        num_runs: opts[:num_runs],
+        max_iterations: opts[:max_iterations],
+        key: key
+      )
```
josevalim, line 175: Ideally we call these two from within `defnp`, so they are both inlined into the same GPU computation. We just returns `{labels, embedding}` from said that `defnp` which we return below.

josevalim (approved):
One comment and we can ship it (as long as @krstopro agrees!)

krstopro:
> One comment and we can ship it (as long as @krstopro agrees!)

Of course! I don't know why am I not able to approve the pull requests though.

josevalim:
@krstopro weird, you should have permissions to do so!

krstopro:
> @krstopro weird, you should have permissions to do so!

Maybe I should have assigned myself as a reviewer or so. Will figure it out. :)

lib/scholar/cluster/spectral_clustering.ex
RicardoSantos-99: Done! Moved both calls into a single defnp that returns {labels, embedding}, so everything compiles into one computation now.

josevalim:

PR #340, opened 2026-07-14
Fix mean Tweedie deviance for power outside 0, 1 and 2

RicardoSantos-99:
I was evaluating a regression model using `mean_tweedie_deviance` with power 1.5 (compound Poisson-gamma) and the deviance looked way off compared to what sklearn gave for the same predictions. Traced it down to the formula for power outside 0, 1 and 2: it was split across multiple lines with the minus and plus operators at the start of each line, which Elixir parses as separate expressions inside parentheses rather than a continued one, so only the last term was actually being returned. Same issue affected d2_tweedie_score.

Joined it into a single expression and added tests comparing against sklearn for power 1.5, 3 and -1, plus a check that deviance is 0 for perfect predictions.

### DIFF
--- a/lib/scholar/metrics/regression.ex
+++ b/lib/scholar/metrics/regression.ex
@@ -275,11 +275,9 @@ defmodule Scholar.Metrics.Regression do
       cond do
         power < 0 ->
           2 *
-            (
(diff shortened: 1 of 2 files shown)


josevalim (approved):

josevalim:

Issue #338, opened 2026-07-13
Add Linear and Quadratic Discriminant Analysis (LDA / QDA)


RicardoSantos-99:
Noticed Scholar doesn't have discriminant analysis yet, would be nice to add:

- [x] LinearDiscriminantAnalysis (LDA)
- [ ] QuadraticDiscriminantAnalysis (QDA)

Refs:
- https://scikit-learn.org/stable/modules/lda_qda.html
- https://scikit-learn.org/stable/modules/generated/sklearn.discriminant_analysis.LinearDiscriminantAnalysis.html
- https://scikit-learn.org/stable/modules/generated/sklearn.discriminant_analysis.QuadraticDiscriminantAnalysis.html

krstopro:
Yes, these are somewhat popular algorithms so it would be nice to have them here. Pull requests are accepted. :)

PR #341, opened 2026-07-16
Add linear discriminant analysis classifier

RicardoSantos-99:
Adds `Scholar.DiscriminantAnalysis.Linear`, the Linear Discriminant Analysis
classifier. Addresses the LDA item of #338.

LDA fits a Gaussian to each class under a shared within-class covariance,
which gives a linear decision boundary between every pair of classes. The
model exposes `fit/3`, `decision_function/2`, `predict/2` and
`predict_probability/2`, and stores `coefficients`, `intercept`, `means`,
`priors` and `classes`.

A few decisions worth calling out:

* The pooled within-class covariance is normalized by
  `num_samples - num_classes` and the scores are parameterized around the
  overall mean, matching scikit-learn's `svd` solver. That parameterization
  makes `decision_function` differ from the bare textbook formula by a
  class-constant term, which cancels in `predict` and `predict_probability`.
* The coefficients are obtained by solving the pooled covariance system
  directly (`Nx.LinAlg.solve`) rather than via a double SVD. This is
  algebraically equivalent for full-rank data and matches scikit-learn to
  about 1.0e-13, but it assumes an invertible covariance (documented). A
  singular covariance raises instead of silently returning garbage.
* `predict_probability` applies a softmax to the discriminant scores. For
  two classes this reduces exactly to scikit-learn's binary sigmoid.

## Verification

Results match `sklearn.discriminant_analysis.LinearDiscriminantAnalysis`
(solver `svd`) for `coefficients`, `intercept`, `decision_function`,
`predict` and `predict_probability`, on balanced and unbalanced classes,
binary and multiclass, and 2 and 3 features, all to roughly 1.0e-13 in f64.
The reference values in the tests were generated with scikit-learn 1.6.1.

### DIFF
--- a/lib/scholar/discriminant_analysis/linear.ex
+++ b/lib/scholar/discriminant_analysis/linear.ex
@@ -1,0 +1,187 @@
+defmodule Scholar.DiscriminantAnalysis.Linear do
+  @moduledoc ~S"""
+  Linear Discriminant Analysis (LDA) for classification.
+
+  LDA fits a Gaussian density to each class, assuming that all classes share the
+  same covariance matrix. The shared covariance leads to a linear decision
+  boundary between every pair of classes. For a sample $x$ the score of class $k$
+  is
+
+  $$ \delta\_{k}(x) = x^{T} \Sigma^{-1} \mu\_{k} - \frac{1}{2} \mu\_{k}^{T} \Sigma^{-1} \mu\_{k} + \log \pi\_{k} $$
+
+  where $\mu\_{k}$ is the class mean, $\pi\_{k}$ the class prior and $\Sigma$ the
+  pooled within-class covariance. The predicted class is the one with the
+  highest score.
+
+  The pooled covariance is assumed to be invertible, which holds when the number
+  of samples is larger than the number of features and no feature is a linear
+  combination of the others.
+
+  Reference:
+
+  * [1] - [The Elements of Statistical Learning, Hastie, Tibshirani & Friedman, Section 4.3](https://hastie.su.domains/ElemStatLearn/)
+  """
+  import Nx.Defn
+  import Scholar.Shared
+
+  @derive {Nx.Container, containers: [:coefficients, :intercept, :means, :priors, :classes]}
+  defstruct [:coefficients, :intercept, :means, :priors, :classes]
+
+  opts_schema = [
+    num_classes: [
+      type: :pos_integer,
+      required: true,
+      doc:
+        "Number of different classes used in training. Labels must be integers in `[0, num_classes)`."
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(opts_schema)
+
+  @doc """
+  Fits a Linear Discriminant Analysis model for sample inputs `x` and target
+  labels `y`.
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+  The function returns a struct with the following parameters:
+
+    * `:coefficients` - Weight vector of each class, shape `{num_classes, num_features}`.
+
+    * `:intercept` - Intercept term of each class, shape `{num_classes}`.
+
+    * `:means` - Class-wise mean of the samples, shape `{num_classes, num_features}`.
+
+    * `:priors` - Class prior probabilities, shape `{num_classes}`.
+
+    * `:classes` - Labels seen during fit, shape `{num_classes}`.
+
+  ## Examples
+
+      iex> x = Nx.tensor([[-2.0, -1.0], [-1.0, -1.0], [-1.0, -2.0], [1.0, 1.0], [1.0, 2.0], [2.0, 1.0]])
+      iex> y = Nx.tensor([0, 0, 0, 1, 1, 1])
+      iex> model = Scholar.DiscriminantAnalysis.Linear.fit(x, y, num_classes: 2)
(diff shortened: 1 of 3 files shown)


josevalim (approved):

lib/scholar/discriminant_analysis/linear.ex
```diff
@@ -0,0 +1,192 @@
...
+
+  Reference:
+
+  * [1] - [The Elements of Statistical Learning, Hastie, Tibshirani & Friedman, Section 4.3](https://hastie.su.domains/ElemStatLearn/)
+  """
+  import Nx.Defn
+  import Scholar.Shared
+
+  @derive {Nx.Container, containers: [:coefficients, :intercept, :means, :priors, :classes]}
+  defstruct [:coefficients, :intercept, :means, :priors, :classes]
```
krstopro: Do you need to store classes given that they are `Nx.iota({num_classes})`? Would it be better to store `num_classes`?

I am not sure about this myself.

lib/scholar/discriminant_analysis/linear.ex
```diff
@@ -0,0 +1,192 @@
...
+      iex> model = Scholar.DiscriminantAnalysis.Linear.fit(x, y, num_classes: 2)
+      iex> Scholar.DiscriminantAnalysis.Linear.predict(model, Nx.tensor([[-1.5, -1.5], [1.5, 1.5]]))
+      #Nx.Tensor<
+        s32[2]
+        [0, 1]
+      >
+  """
+  defn predict(%__MODULE__{classes: classes} = model, x) do
+    scores = decision_function(model, x)
+    Nx.take(classes, Nx.argmax(scores, axis: 1))
```
krstopro: Do you need `Nx.take/2` here given that your classes are `Nx.iota({num_classes})`?

lib/scholar/discriminant_analysis/linear.ex
```diff
@@ -0,0 +1,192 @@
...
+      iex> probs = Scholar.DiscriminantAnalysis.Linear.predict_probability(model, Nx.tensor([[1.5, 1.5]]))
+      iex> Nx.sum(probs) |> Nx.round()
+      #Nx.Tensor<
+        f32
+        1.0
+      >
+  """
+  defn predict_probability(%__MODULE__{} = model, x) do
+    scores = decision_function(model, x)
+    scores = scores - Nx.reduce_max(scores, axes: [1], keep_axes: true)
```
krstopro: You might wanna apply `stop_grad/1` after `Nx.reduce_max/2` as in
https://github.com/elixir-nx/scholar/blob/3619c28c45c760d98c20d81941736f0e7da66183/lib/scholar/linear/logistic_regression.ex#L223

krstopro: Though thinking of it, this function might need to be defined somewhere else, perhaps in `Nx` itself.

krstopro (commented):
Few remarks.

lib/scholar/discriminant_analysis/linear.ex
RicardoSantos-99: You're right, since the module requires labels to be integers in [0, num_classes), classes is always `Nx.iota({num_classes})` and carries nothing beyond num_classes. Dropping it from the struct. predict then just returns the argmax directly (your next comment).

lib/scholar/discriminant_analysis/linear.ex
RicardoSantos-99: Right, `Nx.take(Nx.iota(...), idx)` is just idx. predict now returns `Nx.argmax(scores, axis: 1)` directly.

lib/scholar/discriminant_analysis/linear.ex
RicardoSantos-99: Done, wrapped the `reduce_max` in `stop_grad` to match the softmax helpers in logistic regression.

RicardoSantos-99: > Though thinking of it, this function might need to be defined somewhere else, perhaps in `Nx` itself.

Agreed the stable-softmax shows up in a few places now. I left it here for this PR, happy to open a follow-up if you think it belongs in `Scholar.Shared` or upstream in Nx.

krstopro (approved):
Alright, looks good to me.

@josevalim @polvalente We have softmax several times in `Scholar` and it is implemented in `Axon`. It can be implemented in `Scholar.Shared`, but perhaps it might be better to have it in `Nx`?

josevalim:

polvalente:
@krstopro I think it's ok for us to add softmax in Nx, but we have to think about what types of shapes we want to accept in it (for instance, do we want to always force rank-1 inputs, treat leading axes as batch dims, accept an axes option to calculate the softmax against?). 

Do you wanna submit a PR to Nx with one that can solve both the Axon and Scholar use cases?
I think we should do it as an Nx.Block so that backends can optionally override the implementation.

josevalim:
My .02 cents is that it is a small function that is ok to repeat. The only reason to add it to Nx is only if there are indeed cases where we wnat to override the implementation.

polvalente:
I checked that at least MLX has a fast activation, and we already have sigmoid which I think falls im a similar class of "small and could be copied over" but is still in the Nx API. And softmax is very common overall

josevalim:
So it sounds good to me. And I guess we can look at sigmoid as an example of the API?

krstopro:
> we already have sigmoid which I think falls im a similar class of "small and could be copied over" but is still in the Nx API

Same for `logsumexp` and there are probably others.

> And softmax is very common overall

This was my main argument for it.

PR #343, opened 2026-07-20
Fix KDTree pruning to test against the splitting node

### LINKED ISSUE #342: KDTree.predict returns wrong nearest neighbors
## Problem

`Scholar.Neighbors.KDTree.predict/2` returns incorrect nearest neighbors on a
noticeable fraction of inputs. Checked against `Scholar.Neighbors.BruteKNN`
(exact by definition) on 20 random small datasets ({12, 2}, k=3, uniform in
[0, 10]): 7 out of 20 disagreed, with distance errors up to ~0.4 on that
scale. Also diverges from `sklearn.neighbors.NearestNeighbors` with the
default euclidean metric and with chebyshev on a larger dataset ({30, 3}).

## Reproduction

```elixir
key = Nx.Random.key(4)
{x, key2} = Nx.Random.uniform(key, 0.0, 10.0, shape: {12, 2}, type: :f64)
{q, _} = Nx.Random.uniform(key2, 0.0, 10.0, shape: {3, 2}, type: :f64)

tree = Scholar.Neighbors.KDTree.fit(x, num_neighbors: 3)
{kd_idx, kd_dist} = Scholar.Neighbors.KDTree.predict(tree, q)

brute = Scholar.Neighbors.BruteKNN.fit(x, num_neighbors: 3)
{b_idx, b_dist} = Scholar.Neighbors.BruteKNN.predict(brute, q)

kd_idx
# #Nx.Tensor<s64[3][3]> [[3, 9, 11], [10, 2, 9], [5, 6, 0]]

b_idx
# #Nx.Tensor<s64[3][3]> [[3, 9, 11], [0, 10, 2], [5, 6, 0]]
```
Row 1 disagrees: KDTree returns [10, 2, 9], BruteKNN returns [0, 10, 2].
kd_dist for that row is larger than b_dist, so KDTree is missing an
actual nearer neighbor, not just breaking a tie differently.

Why it matters
KDTree exists to be a faster drop-in for exact nearest-neighbor search, so
its whole point is to match brute force, only quicker. It's also a backend
option for Scholar.Neighbors.KNNClassifier, KNNRegressor, and
Scholar.Cluster.OPTICS (via algorithm: :kd_tree), so wrong neighbors here
propagate silently into wrong predictions and wrong cluster assignments
upstream, with nothing to signal that anything went wrong.

Scope
I haven't traced the exact bug in the traversal logic yet (the backtracking
walk in predict_n is fairly involved), just confirmed the wrong output and
its impact. Happy to dig into the traversal itself if useful, but wanted to
flag this first since it's silent and affects several other modules.

RicardoSantos-99:
Closes #342.

The backtracking step in `predict_n` decided whether to visit the sibling subtree by measuring the distance from the query to the child node's point along the split axis, instead of to the splitting hyperplane of the current node. Since the child lies inside the subtree, this overestimates the minimum possible distance and prunes subtrees that still contain closer points. On random 12x2 datasets with k=3, more than half of the queries disagreed with brute-force search.

The fix uses `indices[node]`, the point defining the cut, in both pruning tests.

Checked against `BruteKNN` across dimensions, k, and metrics (Minkowski p=1,2,3 and Chebyshev), and against scikit-learn's kd_tree. Added two regression tests comparing with brute force.

### DIFF
--- a/lib/scholar/neighbors/kd_tree.ex
+++ b/lib/scholar/neighbors/kd_tree.ex
@@ -461,7 +461,7 @@ defmodule Scholar.Neighbors.KDTree do
                   if Nx.any(
                        metric.(
                          point[[coord_indicator]],
-                         data[[indices[right_child(node)], coord_indicator]]
+                         data[[indices[node], coord_indicator]]
                        ) <
                          distances
                      ) do
@@ -489,7 +489,7 @@ defmodule Scholar.Neighbors.KDTree do
                   if Nx.any(
                        metric.(
                          point[[coord_indicator]],
-                         data[[indices[left_child(node)], coord_indicator]]
+                         data[[indices[node], coord_indicator]]
                        ) <
                          distances
                      ) do
--- a/test/scholar/neighbors/kd_tree_test.exs
+++ b/test/scholar/neighbors/kd_tree_test.exs
@@ -130,5 +130,39 @@ defmodule Scholar.Neighbors.KDTreeTest do
         ])
       )
     end
+
+    test "agrees with brute-force search" do
+      for seed <- 1..10 do
+        key = Nx.Random.key(seed)
+        {x, key} = Nx.Random.uniform(key, 0.0, 10.0, shape: {12, 2}, type: :f64)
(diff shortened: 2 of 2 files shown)


josevalim (approved):

josevalim:

PR #347, opened 2026-08-06
BIRCH clustering

paulsullivanjr:
Adds `Scholar.Cluster.BIRCH`.

  BIRCH makes a single pass over the data, compressing it into a
  height-balanced tree of Clustering Features — additive `(n, linear_sum,
  squared_sum)` triples that summarize a group of points without storing
  them. The leaf centroids are then reduced to `:num_clusters`, and each
  input point is labelled by its nearest subcluster. 2000 points compress
  to ~34 subclusters before the global step runs.

  The CF-tree is built host-side in Elixir since insertion is sequential;
  the labelling and global steps are tensor work.

  ```elixir
  model = Scholar.Cluster.BIRCH.fit(x, num_clusters: 3, key: key)
  model.labels
  Scholar.Cluster.BIRCH.predict(model, new_x)
  ```

  Options mirror scikit-learn's defaults: `:threshold` (0.5),
  `:branching_factor` (50), `:num_clusters` (3), plus `:key`,
  `:num_runs`, `:max_iterations` for the global step.

  **Validation.** The CF-tree is independent of the global step, so it was
  differential-tested against scikit-learn 1.9.0 across 650 random cases
  (varying n, dimensionality, branching factor, threshold; half on an
  integer grid to force ties): **248/250** and **398/400** exact agreement
  on leaf centroids at 1e-9. Residuals are last-ULP boundary decisions in
  deep trees. 18 tests plus doctests, including three sklearn reference
  tests and a regression test pinning subcluster ordering on splits.

  **One deviation:** the global step uses `Scholar.Cluster.KMeans` rather
  than agglomerative clustering, following `SpectralClustering`'s
  precedent. `Scholar.Cluster.Hierarchical` can't be used — it reorders
  dendrogram rows without remapping the clade ids stored inside them, so
  `labels_list` crashes on 25/25 random 50-point datasets. Pre-existing
  and unrelated to BIRCH; left alone here, happy to file separately.

  Not included: `n_clusters=None`, `transform/2`, `partial_fit`.

### DIFF
--- a/lib/scholar/cluster/birch.ex
+++ b/lib/scholar/cluster/birch.ex
@@ -1,0 +1,384 @@
+defmodule Scholar.Cluster.BIRCH do
+  @moduledoc """
+  BIRCH (Balanced Iterative Reducing and Clustering using Hierarchies).
+
+  BIRCH incrementally builds a height-balanced tree of Clustering Features (CFs).
+  Each point is inserted into the closest leaf subcluster if the resulting subcluster
+  radius stays below `:threshold`; otherwise a new subcluster is created. Nodes that
+  exceed `:branching_factor` are split. The leaf subcluster centroids are then reduced
+  to the requested number of clusters with agglomerative (hierarchical) clustering.
+
+  Because the tree is built by inserting one sample at a time, the CF-tree is
+  constructed on the host in Elixir while the numeric work (distances, radii, the
+  global clustering step) is done with `Nx`. Each insertion reads a scalar back from
+  the tensor backend, so the tree-building phase does not benefit from a compiled
+  backend the way the rest of Scholar does.
+
+  Unlike scikit-learn's `Birch`, which reduces the leaf subclusters with agglomerative
+  clustering, the global step here uses `Scholar.Cluster.KMeans`. Labels may therefore
+  differ from scikit-learn on data where the two disagree, and the step is seeded by
+  the `:key` option.
+
+  Building the CF-tree takes a single pass over the data: inserting one sample costs
+  $O(BDH)$, where $B$ is `:branching_factor`, $D$ is the number of features and $H$ is
+  the height of the tree, for $O(NBDH)$ overall with $N$ samples. Space complexity is
+  $O(MD)$ where $M$ is the number of leaf subclusters, which is typically far smaller
+  than $N$ — this is what makes BIRCH suitable for large datasets. Labelling the input
+  at the end is $O(NMD)$.
+
+  Reference:
+
+  * [BIRCH: An Efficient Data Clustering Method for Very Large Databases](https://www.cs.sfu.ca/CourseCentral/459/han/papers/zhang96.pdf)
+  """
+
+  import Nx.Defn
+  import Scholar.Shared
+
+  @derive {Nx.Container, containers: [:subcluster_centers, :subcluster_labels, :labels]}
+  defstruct [:subcluster_centers, :subcluster_labels, :labels]
+
+  opts = [
+    threshold: [
+      type: {:custom, Scholar.Options, :positive_number, []},
+      default: 0.5,
+      doc: """
+      The radius of the subcluster obtained by merging a new sample and the closest subcluster
+      must be less than this threshold, otherwise a new subcluster is started.
+      """
+    ],
+    branching_factor: [
+      type: :pos_integer,
+      default: 50,
+      doc: """
+      The maximum number of subclusters in each node. If a node exceeds this value,
+      it is split into two nodes.
+      """
+    ],
+    num_clusters: [
+      type: :pos_integer,
+      default: 3,
+      doc: "The number of clusters to form from the leaf subclusters."
+    ],
+    num_runs: [
+      type: :pos_integer,
+      default: 10,
+      doc: """
+      Number of times the k-means algorithm will be run on the leaf subcluster centroids
+      with different centroid seeds. The final result is the best output in terms of inertia.
+      """
+    ],
+    max_iterations: [
+      type: :pos_integer,
+      default: 300,
+      doc: "Maximum number of iterations of the k-means algorithm for a single run."
+    ],
+    key: [
+      type: {:custom, Scholar.Options, :key, []},
+      doc: """
+      Determines random number generation for centroid initialization of the global
+      clustering step. If the key is not provided, it is set to `Nx.Random.key(System.system_time())`.
+      """
+    ]
+  ]
+
+  @opts_schema NimbleOptions.new!(opts)
+
+  @doc """
+  Fits a BIRCH model for sample inputs `x`.
+
+  ## Options
+
+  #{NimbleOptions.docs(@opts_schema)}
+
+  ## Return Values
+
+  The function returns a struct with the following parameters:
+
+  * `:subcluster_centers` - Centroids of all the leaf subclusters.
+
+  * `:subcluster_labels` - The cluster each leaf subcluster is assigned to.
+
+  * `:labels` - Cluster label of each point in `x`.
+
+  ## Examples
+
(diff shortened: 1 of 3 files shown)


josevalim:
Thank you @paulsullivanjr! Unfortunately this pull request does a lot of work outside of Nx, which may:

1. Indicate this algorithm does not fit Scholar
2. We need to refactor the algorithm

@RicardoSantos-99 can you take a look if this could mostly written as `defn`, as in your previous PRs?

> Scholar.Cluster.Hierarchical can't be used — it reorders dendrogram rows without remapping the clade ids stored inside them, so labels_list crashes on 25/25 random 50-point datasets. Pre-existing and unrelated to BIRCH; left alone here, happy to file separately.

I am submitting a fix for that, thanks for filing!

RicardoSantos-99:
Nice work on the sklearn parity, @paulsullivanjr.

> @RicardoSantos-99 can you take a look if this could mostly written as `defn`, as in your previous PRs?


@josevalim, Not with the CF-tree in place, I think. defn wants static shapes, and the tree is dynamic in node count, subcluster count and depth. Preallocating for the worst case means roughly one node per sample, which gives up the memory advantage that makes BIRCH worth using.

It also isn't buying much here. The tree exists to avoid scanning every subcluster, but scanning them all is one matvec for us.

Without it, the first phase is a single while over the samples with preallocated buffers, same shape as `parallel_nearest_neighbor/3` in Hierarchical. The global step, labelling and predict are already fine.

The cost is the sklearn parity, and `:branching_factor` stops meaning anything.

RicardoSantos-99:
> Not included: n_clusters=None, transform/2, partial_fit.

@paulsullivanjr You mention partial_fit as out of scope, which is fair for a first PR, though it's the piece that usually justifies BIRCH over KMeans, and PCA already has incremental_fit/2 as a precedent. Worth noting the flat version makes it easier rather than harder: the state is just the CF buffers and a counter, which is close to resumable already.

RicardoSantos-99:
@paulsullivanjr here's what I had in mind, concretely, in case it's useful.

What comes out is everything from `insert_root/4` down through `closest_index/2`, so the tree build and the CF helpers around it. `global_labels/3`, `assign_labels/3`, `predict/2`, the struct and the options schema stay as they are.

In its place, phase 1 becomes one `while` over the samples. The CF buffers are preallocated and a counter tracks how many rows are live:

```elixir
while {ls, ss, counts, centers, m = 0, i = 0, x}, i < n do
  p = x[i]
  active = Nx.iota({max_m}) < m
  dist = Nx.select(active, Nx.sum((centers - p) ** 2, axes: [1]), inf)
  j = Nx.argmin(dist)

  n_j = counts[j] + 1
  ls_j = ls[j] + p
  ss_j = ss[j] + Nx.sum(p * p)
  radius2 = ss_j / n_j - Nx.sum((ls_j / n_j) ** 2)
  fits? = m > 0 and radius2 <= threshold ** 2

  idx = Nx.select(fits?, j, m)
  ls = Nx.put_slice(ls, [idx, 0], Nx.reshape(Nx.select(fits?, ls_j, p), {1, d}))
  # same for ss, counts and centers

  {..., m + Nx.select(fits?, 0, 1), i + 1, x}
end
```

`Nx.put_slice/3` takes a dynamic start index, which is what lets the same line either merge into row `j` or open a new row at `m` without the shapes changing. You'd want a `:max_subclusters` option to size the buffers, defaulting to the sample count, and `:branching_factor` would go.

Worth being clear that this is still sequential over the samples, and nothing makes it otherwise, since each one depends on the ones before it. The gain is that the loop compiles into a single program instead of reading scalars back per sample. `parallel_nearest_neighbor/3` in `Hierarchical` is in the same position, and `dbscan_inner/2` has a comment explaining why they swapped a sequential DFS for label propagation.

The awkward part is the tests. The reference values won't survive, since a flat scan picks a different subcluster than the descent does. I'd rewrite them around properties instead: the radius bound holds for every subcluster, every point lands in its nearest one, results stay order dependent. Happy to help with that.

Two unrelated things I noticed while reading. The moduledoc says agglomerative clustering on line 9 and KMeans further down. And `assign_labels/3` could use `pairwise_squared_euclidean`, since the square root doesn't change the `argmin`.

josevalim:
@krstopro so I am thinking this one may not be a good fit for Scholar after all. :(

krstopro:
> @krstopro so I am thinking this one may not be a good fit for Scholar after all. :(

@josevalim Quite possible, lemme have a look...

krstopro:
> @krstopro so I am thinking this one may not be a good fit for Scholar after all. :(

@josevalim I agree, there are a lot of operations being performed outside of `Nx` which makes it not suitable for `Scholar`.

@paulsullivanjr I don't see a convenient way of doing this within `defn`. Do you maybe see one? If not, I would suggest closing this one.

paulsullivanjr:
@RicardoSantos-99  Good point, and PCA.incremental_fit/2 is a precedent I'd missed. Agreed
  the flat form makes it much easier.

paulsullivanjr:
@krstopro I don’t see an easy way either. I’m happy to close it.

paulsullivanjr:
> @paulsullivanjr here's what I had in mind, concretely, in case it's useful.
> 
> What comes out is everything from `insert_root/4` down through `closest_index/2`, so the tree build and the CF helpers around it. `global_labels/3`, `assign_labels/3`, `predict/2`, the struct and the options schema stay as they are.
> 
> In its place, phase 1 becomes one `while` over the samples. The CF buffers are preallocated and a counter tracks how many rows are live:
> 
> ```elixir
> while {ls, ss, counts, centers, m = 0, i = 0, x}, i < n do
>   p = x[i]
>   active = Nx.iota({max_m}) < m
>   dist = Nx.select(active, Nx.sum((centers - p) ** 2, axes: [1]), inf)
>   j = Nx.argmin(dist)
> 
>   n_j = counts[j] + 1
>   ls_j = ls[j] + p
>   ss_j = ss[j] + Nx.sum(p * p)
>   radius2 = ss_j / n_j - Nx.sum((ls_j / n_j) ** 2)
>   fits? = m > 0 and radius2 <= threshold ** 2
> 
>   idx = Nx.select(fits?, j, m)
>   ls = Nx.put_slice(ls, [idx, 0], Nx.reshape(Nx.select(fits?, ls_j, p), {1, d}))
>   # same for ss, counts and centers
> 
>   {..., m + Nx.select(fits?, 0, 1), i + 1, x}
> end
> ```
> 
> `Nx.put_slice/3` takes a dynamic start index, which is what lets the same line either merge into row `j` or open a new row at `m` without the shapes changing. You'd want a `:max_subclusters` option to size the buffers, defaulting to the sample count, and `:branching_factor` would go.
> 
> Worth being clear that this is still sequential over the samples, and nothing makes it otherwise, since each one depends on the ones before it. The gain is that the loop compiles into a single program instead of reading scalars back per sample. `parallel_nearest_neighbor/3` in `Hierarchical` is in the same position, and `dbscan_inner/2` has a comment explaining why they swapped a sequential DFS for label propagation.
> 
> The awkward part is the tests. The reference values won't survive, since a flat scan picks a different subcluster than the descent does. I'd rewrite them around properties instead: the radius bound holds for every subcluster, every point lands in its nearest one, results stay order dependent. Happy to help with that.
> 
> Two unrelated things I noticed while reading. The moduledoc says agglomerative clustering on line 9 and KMeans further down. And `assign_labels/3` could use `pairwise_squared_euclidean`, since the square root doesn't change the `argmin`.

@RicardoSantos-99   Makes sense, better fit for Scholar than what I did, and agreed on moving the tests to properties.

  Good catches on both, line 9 is a leftover from when the global step moved to KMeans and squared is right since the argmin is unchanged.

PR #348, opened 2026-08-06
Fix linear interpolation predict for unsorted input

### LINKED ISSUE #346: Linear interpolator `predict` doesn't handle unsorted input
(Scholar version 0.4.2)
(Nx version 0.13.0)
(Elixir version 1.20.1)

Linear interpolator's `predict` method doesn't return interpolated values in the same order as the input.

```
alias Scholar.Interpolation.Linear

# Fit a trivial line: f(x) = 10 * x
model = Linear.fit(Nx.tensor([1.0, 2.0, 3.0]), Nx.tensor([10.0, 20.0, 30.0]))

run = fn label, target_list ->
  target_x = Nx.tensor(target_list)
  result = Nx.to_list(Linear.predict(model, target_x))
  expected = Enum.map(target_list, &(&1 * 10.0))

  IO.puts("\n#{label}")
  IO.puts("  target_x : #{inspect(target_list)}")
  IO.puts("  result   : #{inspect(result)}")
  IO.puts("  expected : #{inspect(expected)}   (f(x) = 10x)")
  IO.puts("  correct? : #{result == expected}")
end

run.("Sorted input", [1.0, 2.0, 3.0])
run.("Unsorted input", [3.0, 1.0, 2.0])
```

Running yields

```
Sorted input
  target_x : [1.0, 2.0, 3.0]
  result   : [10.0, 20.0, 30.0]
  expected : [10.0, 20.0, 30.0]   (f(x) = 10x)
  correct? : true

Unsorted input
  target_x : [3.0, 1.0, 2.0]
  result   : [20.0, 30.0, 10.0]
  expected : [30.0, 10.0, 20.0]   (f(x) = 10x)
  correct? : false
```

RicardoSantos-99:
Closes #346.

`Scholar.Interpolation.Linear.predict/2` sorts `target_x` so its sequential scan can walk the bins in order, then maps the results back at the end. It applied the sort permutation where it needed the inverse, so unsorted input came back shuffled:

```elixir
model = Linear.fit(Nx.tensor([1.0, 2.0, 3.0]), Nx.tensor([10.0, 20.0, 30.0]))
Linear.predict(model, Nx.tensor([3.0, 1.0, 2.0]))
#=> [20.0, 30.0, 10.0]   expected [30.0, 10.0, 20.0]
```

### Why this reaches further than interpolation

`Scholar.Linear.IsotonicRegression.predict/2` calls into it, so isotonic predictions stopped being monotonic in x whenever the input wasn't sorted. Fitting with `increasing: true` and predicting on unsorted x returned `x=1.0 -> 4.5` alongside `x=5.0 -> 2.5`.

Non-metric MDS then calls isotonic with unsorted dissimilarities, so it has been optimizing against shuffled disparities.

### On the MDS test values

The reference values in `mds_test.exs` encoded the old behaviour, so they change here. Three things say the new ones are the correct values rather than merely different:

| | normalized stress | ordering recovered (Kendall tau vs `[0..9]`) |
|---|---|---|
| scikit-learn 1.6.1 (best of 20 seeds) | 0.090 | +1.000 |
| this PR | 0.136 | +0.867 |
| before | 0.249 | +0.467 |

Stress is what MDS minimizes, and it drops in both affected tests (1.288 to 0.197, and 0.249 to 0.136). The fixture input is `Nx.iota({10, 50})`, whose rows are collinear, so a correct 2-D embedding should lay the points out in order along one axis. The remaining gap to scikit-learn is that it takes the best of 20 random inits on a non-convex problem while Scholar runs a single init from `key(42)`.

Isotonic itself is deterministic, and it now agrees with scikit-learn to within f32 precision on unsorted input.

### Tests

Regression tests for unsorted `target_x` in `Linear`, including values outside the training bounds so the reorder is covered across all three loops, plus an order-independence test for `IsotonicRegression`.

Worth noting the existing isotonic `predict` test already used unsorted input and still passed. Its `argsort` is `[1, 0, 2]`, a swap of two entries, which is its own inverse and so cannot distinguish a permutation from its inverse. The new tests use a 3-cycle.

Also removed a stray `{res, i}` expression in `predict_n` that had no effect. Happy to drop that from the diff if you would rather keep this to the fix.


josevalim (approved):

RicardoSantos-99:
@josevalim CI failed here, probably because of this Actions incident: https://www.githubstatus.com/incidents/qcvjkzcs7j74

I don't have write access to re-run it. Could you re-run it?

Should be green now.

josevalim:

PR #349, opened 2026-08-08
Add quadratic discriminant analysis classifier

### LINKED ISSUE #338: Add Linear and Quadratic Discriminant Analysis (LDA / QDA)

RicardoSantos-99:
Closes #338. LDA landed in #341, so this completes it.

Adds `Scholar.DiscriminantAnalysis.Quadratic`. Each class keeps its own covariance, stored as its eigendecomposition so both the log determinant and the quadratic form become elementwise work over the eigenvalues. Options are `:num_classes` (required, as in LDA) and `:reg_param`.

### Two implementation notes

Covariances are built and decomposed one class at a time. That keeps the working set at O(num_samples * num_features), and it avoids `Nx.LinAlg.eigh/2` over a stack of matrices, which on Nx 0.9.x under EXLA decomposes only the first one and returns zeros for the rest. Fixed in Nx 0.13, but Scholar still allows `~> 0.9`.

`eigh` is called with a tighter `eps` than its default. QDA divides by the eigenvalues and takes their log, and the default `1.0e-4` leaves them about three digits short in the pure Nx implementation. Compiled backends replace that routine and are unaffected.

### Validation against scikit-learn 1.6.1

Twelve scenarios (2 to 5 classes, up to 10 features, unbalanced, overlapping, correlated features, mixed scales, `reg_param`, singular covariance), EXLA in f64: means, eigenvalues and scores within 6.0e-8, identical predictions in eleven of twelve.

On a 30% stratified holdout, every prediction agrees:

| dataset | this PR | scikit-learn | LDA |
|---|---|---|---|
| iris | 1.0000 | 1.0000 | 1.0000 |
| wine | 0.9815 | 0.9815 | 1.0000 |
| breast_cancer | 0.9532 | 0.9532 | 0.9532 |
| digits (61 features, 10 classes) | 0.9704 | 0.9704 | 0.9593 |

### One deliberate divergence

The twelfth scenario is a class with fewer samples than features. scikit-learn keeps only `min(class_size, num_features)` components and works in the subspace the class spans; this keeps all directions and relies on `:reg_param`. Noted in the moduledoc.

Tests include the `jit_apply` case AGENTS.md asks for, and two classes that share a mean and differ only in spread, where LDA cannot separate them.

### DIFF
--- a/lib/scholar/discriminant_analysis/quadratic.ex
+++ b/lib/scholar/discriminant_analysis/quadratic.ex
@@ -20,8 +20,22 @@ defmodule Scholar.DiscriminantAnalysis.Quadratic do
 
   Every class covariance is assumed to be invertible, which requires at least as
   many samples as features in each class and no feature being a linear
-  combination of the others. Use `:reg_param` to shrink the eigenvalues towards
-  one when that does not hold.
+  combination of the others. When that does not hold, a covariance is singular
+  and its smallest eigenvalues come out as zero, which the score then divides by.
+  Use `:reg_param` to shrink the eigenvalues towards one in that case.
+
+  Note that scikit-learn handles rank deficiency differently: it decomposes the
+  centered class data and keeps only `min(class_size, num_features)` components,
+  so it works in the subspace the class actually spans. This implementation keeps
+  all `num_features` directions and relies on `:reg_param` instead, so results
+  diverge from scikit-learn's for a class with fewer samples than features.
+
+  Forming the covariance squares its condition number, so features on wildly
+  different scales cost accuracy in the eigendecomposition well before the
+  matrix is singular. A compiled backend absorbs this, but the pure-Nx
+  eigendecomposition does not, and features spanning several orders of magnitude
+  can lose the smallest eigenvalues there entirely. Standardizing the features
+  beforehand, with `Scholar.Preprocessing.StandardScaler`, avoids it either way.
 
   Reference:
 
@@ -33,11 +47,13 @@ defmodule Scholar.DiscriminantAnalysis.Quadratic do
   @derive {Nx.Container, containers: [:means, :priors, :scalings, :rotations]}
   defstruct [:means, :priors, :scalings, :rotations]
 
-  # `Nx.LinAlg.eigh/2` defaults to `eps: 1.0e-4`, which leaves the eigenvalues
-  # about three digits short. QDA divides by them and takes their log, so the
-  # error is amplified rather than absorbed. Measured against LAPACK on the same
-  # f64 covariance, 1.0e-8 lands the eigenvalues within ~1.0e-7; tightening
-  # further makes it worse, since the iteration then runs into `:max_iter`.
+  # Only matters for the pure-Nx eigendecomposition, which compilers like EXLA
+  # replace with a native routine. There `Nx.LinAlg.eigh/2` runs a Jacobi
+  # iteration whose default `eps: 1.0e-4` leaves the eigenvalues about three
+  # digits short, and QDA divides by them and takes their log, so that error is
+  # amplified rather than absorbed. Measured against LAPACK on the same f64
+  # covariance, 1.0e-8 lands the eigenvalues within ~1.0e-7; tightening further
+  # makes it worse, since the iteration then runs into `:max_iter`.
   @eigh_eps 1.0e-8
 
   opts_schema = [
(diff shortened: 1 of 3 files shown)


RicardoSantos-99:
@josevalim while working on this I hit a bug in the Nx the lockfile pins. On 0.9.2, `Nx.LinAlg.eigh/2` over a stack of matrices decomposes only the first one under EXLA and returns zeros for the rest:

```elixir
one = Nx.tensor([[2.0, 1.0], [1.0, 2.0]])

Nx.Defn.jit_apply(
  fn t -> Nx.LinAlg.eigh(t) |> elem(0) end,
  [Nx.stack([one, one])],
  compiler: EXLA
)
#=> [[1.0, 3.0], [0.0, 0.0]]
``` 

It is nondeterministic, sometimes exiting the VM silently instead of returning. Both routes into EXLA's native routine are affected, the compiler and EXLA.Backend on its own. BinaryBackend is fine, and it is already fixed in Nx 0.13, which I confirmed on a clean project.

Worth bumping? It fails silently with wrong numbers rather than raising, so anything in Scholar reaching a batched decomposition is exposed, not just my case. I worked around it by decomposing one class at a time, but that workaround only exists because of 0.9.

josevalim:
> Fixed in Nx 0.13, but Scholar still allows ~> 0.9.

We can bump the requirement to v0.13 if that is a better solution.

lib/scholar/discriminant_analysis/quadratic.ex
```diff
@@ -0,0 +1,262 @@
...
+  end
+
+  defnp fit_n(x, y, opts) do
+    x = to_float(x)
+    num_classes = opts[:num_classes]
+    reg_param = opts[:reg_param]
+    {num_samples, num_features} = Nx.shape(x)
+
+    one_hot =
+      (Nx.new_axis(y, 1) == Nx.new_axis(Nx.iota({num_classes}), 0)) |> Nx.as_type(Nx.type(x))
```
josevalim, line 141: Don't we already have a function that does this somewhere in the lib?

RicardoSantos-99: I think the `Scholar.Preprocessing.one_hot_encode/2` is the closest one, but it runs the labels through `OrdinalEncoder` first, which compacts them to the distinct values actually present. With `num_classes: 3`:

| y | inline | `one_hot_encode` |
|---|---|---|
| `[0, 2, 2, 0]` | counts `[2, 0, 2]` | counts `[2, 2, 0]` |
| `[1, 1, 2, 2]` | counts `[0, 2, 2]` | counts `[2, 2, 0]` |

So a class missing from the training data shifts every class after it, and the means, priors and covariances would silently line up with the wrong class. That conflicts with the contract documented for `:num_classes` here, which is that labels are already integers in `[0, num_classes)`.

RicardoSantos-99: But the line is duplicated, because `DiscriminantAnalysis.Linear` has the same one, and `LogisticRegression`, `NaiveBayes.Complement` and `Metrics.Classification` have it in a slightly different spelling. 

@josevalim do you think it's better extract to `Scholar.Shared` and use it here and in LDA?, leaving the other three for a follow up.

josevalim: Nah, it is fine to have it duplicated, it is a single line of code anyway!

josevalim:
Btw, if we are bumping Nx, let's do that on another PR and then we rebase this one!

RicardoSantos-99:
> Btw, if we are bumping Nx, let's do that on another PR and then we rebase this one!

@josevalim I tried the bump locally to see how big it is, and it is not a small change.

Two things came up. Nx and EXLA 0.13 declare elixir: `"~> 1.17"`, so Scholar's own requirement moves off `~> 1.14` and the older leg of the CI matrix moves off 1.15.8. That drops 1.14 through 1.16 in one go

The suite also goes to 157 failures on 0.13. Most of it is harmless: 139 are doctests, and 0.13 prints `f32` with the shortest form that round trips, so `7.071067810058594` is now `7.071068`. The rest needs judgment: around 26 cases have values that actually moved, in TSNE, TruncatedSVD, PLSSVD, etc.

For this PR the bump changes nothing structurally. The only thing it buys is dropping the `@eigh_eps` constant, since on 0.13 passing `:eps` makes no difference and the default is already accurate. So I would rather not hold this one on it. Happy to pick the bump up separately whenever you want, and it probably wants to be two PRs, the version plus the formatting churn first, then the fixtures that moved.

josevalim (approved):

josevalim:
Approved this one! A PR for the upgrade would be fantastic as well, thanks!

josevalim:

PR #350, opened 2026-08-12
Fix BayesianRidgeRegression never converging (state rotation bug in the iteration loop)

RicardoSantos-99:
## Bug

`BayesianRidgeRegression.fit/3` never converges on real data. `alpha` and
`lambda` oscillate in a 2-cycle depending on the parity of `:iterations`
instead of converging to the values scikit-learn's `BayesianRidge` finds on
the same data.

Reproduced on the diabetes dataset already used by this test file
(`test/support/diabetes_data_raw.csv`), comparing against scikit-learn 1.6.1:

| iterations | alpha (before) | alpha (after) | sklearn |
|---:|---:|---:|---:|
| 1   | 2.51e-4 | 2.51e-4 | 2.51e-4 |
| 50  | 3.25e-4 | 3.51e-4 | 3.51e-4 |
| 51  | 1.53e+5 | 3.51e-4 | 3.51e-4 |
| 299 | 8.70e+4 | 3.51e-4 | 3.51e-4 |
| 300 | 5.75e-4 | 3.51e-4 | 3.51e-4 |

Adjacent iteration counts can disagree by orders of magnitude (299 vs 300
above), and this holds all the way out to the max iteration count, it never
settles.

Root cause: inside the `while` loop in `fit_n`, the accumulator is declared
as `{coef, rmse, alpha, lambda, ...}` but the loop body returns
`{coef_new, alpha, lambda, rmse, ...}`. Since Nx.Defn's `while` rebinds
loop-carried values by position, not by name, this silently rotates
`rmse`/`alpha`/`lambda` through the wrong slots on every iteration after the
first. `rmse` (order of magnitude ~1e5 on this dataset) ends up used in
place of `lambda` in the `gamma` update, which is what produces the
oscillation.

Two smaller, related bugs are fixed alongside it:
- `default_alpha` used `Var(x)` (pooled variance of the whole design matrix)
  instead of the documented `Var(y)`.
- The loop condition (`iter <= iterations`) ran one extra iteration beyond
  what `:iterations` requests.

Neither of these two causes the oscillation on its own, verified in
isolation; the tuple ordering is the actual cause.

## Fix

Reorder the loop body's return tuple (and the destructuring pattern that
reads the loop's final state) to match the declared accumulator order.
Fix `default_alpha` to use `Var(y)`. Fix the off-by-one in the loop bound.

## Testing

Verified against scikit-learn 1.6.1 `BayesianRidge` on the diabetes dataset.
Added three tests: convergence matches the sklearn reference across a range
of iteration counts, adjacent iteration counts no longer disagree, and the
default `alpha_init` matches the documented `1/Var(y)`.

The two moduledoc doctests changed: the old expected values were the result
of the bug. For the toy `y = x` dataset in the doctest, the model now
converges to the mathematically exact answer (coefficient 1.0, predictions
matching y exactly) instead of a slightly-off fit.

### DIFF
--- a/lib/scholar/linear/bayesian_ridge_regression.ex
+++ b/lib/scholar/linear/bayesian_ridge_regression.ex
@@ -214,12 +214,12 @@ defmodule Scholar.Linear.BayesianRidgeRegression do
       iex> model.coefficients
       #Nx.Tensor<
         f32[1]
-        [0.9932512044906616]
+        [0.9999999403953552]
       >
       iex> model.intercept
       #Nx.Tensor<
         f32
-        0.03644371032714844
+        4.76837158203125e-7
       >
   """
   deftransform fit(x, y, opts \\ []) do
@@ -241,7 +241,7 @@ defmodule Scholar.Linear.BayesianRidgeRegression do
     # handle vector types
     # handle default alpha value, add eps to avoid division by 0
     eps = Nx.Constants.smallest_positive_normal(x_type)
-    default_alpha = Nx.divide(1, Nx.add(Nx.variance(x), eps))
+    default_alpha = Nx.divide(1, Nx.add(Nx.variance(y), eps))
     alpha = Keyword.get(opts, :alpha_init, default_alpha)
     alpha = Nx.tensor(alpha, type: x_type)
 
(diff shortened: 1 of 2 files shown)


RicardoSantos-99:
Visual before/after on the same sweep from the table above (alpha vs iterations, log scale on Y).

**Before** (this branch's parent commit): alpha oscillates between the two branches of the 2-cycle and never settles.

**After** (this PR): alpha and the sklearn reference overlap almost exactly.

before:
image

After:
image

josevalim (approved):

josevalim:

PR #351, opened 2026-08-14
Support precomputed dissimilarities in hierarchical clustering

RicardoSantos-99:
`dissimilarity: :precomputed` was already sketched out in `Scholar.Cluster.Hierarchical`
but left commented out since the module was added in #187. This enables it.

With it, `fit/2` accepts a `{num_obs, num_obs}` dissimilarity matrix directly instead of
computing one from coordinates. That covers dissimilarities that are expensive to
recompute, come from outside Scholar, or are not derived from coordinates at all.

My motivation is HDBSCAN, which clusters a mutual reachability matrix rather than the
raw points. I plan to send that as a follow up and would rather reuse the existing
single linkage implementation than duplicate it.

## Validation

The check is that a precomputed fit on the pairwise euclidean matrix produces
a model identical to a euclidean fit on the raw data, asserted for all five linkages.

Beyond that, a dissimilarity matrix that cannot come from coordinates (it violates the
triangle inequality) is checked against `scipy.cluster.hierarchy.linkage` for single,
complete, average and weighted.

Also covered: the diagonal is ignored, labels can be derived from a precomputed model,
`jit_apply` works, and a non square input raises.

Symmetry is not validated, since checking it would mean inspecting the tensor inside
`deftransform`. It is documented as a caller requirement, along with a note that
`:ward` assumes euclidean dissimilarities.

### DIFF
--- a/lib/scholar/cluster/hierarchical.ex
+++ b/lib/scholar/cluster/hierarchical.ex
@@ -12,7 +12,7 @@ defmodule Scholar.Cluster.Hierarchical do
 
   Due to the requirements of the current implementation, only these options are supported:
 
-    * `dissimilarity: :euclidean`
+    * `dissimilarity: :euclidean | :precomputed`
     * `linkage: :average | :complete | :single | :ward | :weighted`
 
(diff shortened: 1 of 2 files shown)


josevalim:

PR #352, opened 2026-08-14
Add HDBSCAN clustering

RicardoSantos-99:
Implements HDBSCAN, from the clustering tracking issue #135. Unlike `Scholar.Cluster.DBSCAN` it does not need an `eps` radius, so it recovers clusters of differing densities.

Builds on the `dissimilarity: :precomputed` support added in #351: the mutual reachability is fed straight into `Scholar.Cluster.Hierarchical` with `linkage: :single`, so there is no second single linkage implementation here.

The interesting part is step 4, condensing the hierarchy and picking clusters by excess of mass, which is normally written as a recursive tree walk with variable length queues. It is expressible under `defn` because a child cluster's node id in the linkage matrix is always greater than its parent's, so iterating node ids in ascending order is a bottom up traversal and descending order is top down. That turns the whole step into four `while` passes with static trip counts.

## Validation

The condense and excess of mass steps are identical to scikit-learn's internal extractor across 3424 (tree, `min_cluster_size`) combinations, which isolates that step completely. End to end, labels match `sklearn.cluster.HDBSCAN` on any data whose merge heights are distinct. On data with tied heights the dendrogram is genuinely ambiguous and implementations differ without either being wrong, which the moduledoc explains.

Tests cover the metric options, degenerate input (duplicates, all identical, collinear, single feature, smallest accepted size), every numeric dtype, `jit_apply`, determinism, f32 vs f64, and the validation errors.

### DIFF
--- a/lib/scholar/cluster/hdbscan.ex
+++ b/lib/scholar/cluster/hdbscan.ex
@@ -24,6 +24,16 @@ defmodule Scholar.Cluster.HDBSCAN do
   The time complexity is $O(N^2 \\log N)$ for $N$ samples. The space complexity is $O(N^2)$,
   since the mutual reachability is held as a dense matrix.
 
+  ## Comparing against other implementations
+
+  The mutual reachability contains many exactly equal values by construction: whenever a
+  core distance dominates a pair, every pair it dominates collapses onto that same value.
+  Single linkage is then genuinely ambiguous, and different implementations resolve those
+  ties differently while producing dendrograms of identical height. So the labels here can
+  differ from `sklearn.cluster.HDBSCAN` on such data, without either being wrong. On data
+  where the merge heights are all distinct, and therefore the dendrogram is unique, the two
+  agree.
+
   ## Not yet supported
 
   Compared to `sklearn.cluster.HDBSCAN`, the following are not implemented yet:
--- a/lib/scholar/cluster/hdbscan.ex
+++ b/lib/scholar/cluster/hdbscan.ex
@@ -52,6 +52,7 @@ defmodule Scholar.Cluster.HDBSCAN do
       Campello, Moulavi and Sander, 2013.
   """
   import Nx.Defn
+  import Scholar.Shared
 
   @derive {Nx.Container, containers: [:labels]}
   defstruct [:labels]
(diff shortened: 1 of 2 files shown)


RicardoSantos-99:
## Remaining work

Keeping this as a draft while I work through the options scikit-learn supports and this does not yet. The condensed tree they all read from is already built, so these are additions on top rather than rewrites.

- [ ] `max_cluster_size`
- [ ] `cluster_selection_method: :leaf`
- [ ] `allow_single_cluster`, so the root can be returned as a cluster
- [ ] Membership probabilities
- [ ] `cluster_selection_epsilon`

Ordered by how self contained they are. The last two are the ones I would most like a second opinion on: probabilities add a field to the struct, and `cluster_selection_epsilon` has edge cases in how far it walks up the tree that I want to check against scikit-learn carefully. If either looks better as its own PR, happy to split.

CC: @josevalim



Side note: I have a task at work that HDBSCAN fits perfectly. I am going to take the chance to try Scholar on it instead of scikit-learn, and I maybe write an article about that and bring it here soon.

josevalim:
@RicardoSantos-99 if this is suitable as is, my suggestion is to remove the "Not yet supported" from the docs, and we ship this version as soon after you test it at your job. We can open up an issue for the remaining options!

RicardoSantos-99:
@josevalim we tested this exhaustively.

Against a real dataset it matches `sklearn.cluster.HDBSCAN` exactly, same labels including the cluster numbering.

We also put it through a larger validation grid, 1172 scenarios over 26 datasets covering separated and overlapping blobs, noise, varying density, non convex shapes, duplicates, collinear and identical points, and the n = 3 minimum, crossed with both float types and a range of `min_cluster_size` and `min_samples`:

- no violation of the labelling invariants (contiguous ids from zero, nothing smaller than `min_cluster_size`, noise marked)
- where the single linkage tree over the mutual reachability is unique, so the answer is well defined, the partition matches scikit-learn in 178 of 178 scenarios
- where merge heights tie, so there is no single correct answer, it agrees in 64% and returns an equally valid grouping otherwise

Comparing against scikit-learn alone cannot separate a bug from a tie resolved differently, so the dendrograms are also validated against the definition of the linkage: replay the merge sequence through the Lance-Williams recursion and check that each reported height is the real distance, that the pair was a closest pair among the live clades, that sizes add up, and that no clade is consumed twice.

That is what surfaced #356, which fixes a clade being merged twice on tied distances. It is a prerequisite for this, since HDBSCAN builds its hierarchy through `Hierarchical`.

RicardoSantos-99:
@josevalim rebased on top of [#356](https://github.com/elixir-nx/scholar/pull/356), which merged. Ready whenever you have time to look.

Once this is in, I'll open a tracking issue for the remaining `sklearn.cluster.HDBSCAN` options we intentionally left out (`max_cluster_size`, `cluster_selection_method: :leaf`, `alpha`, `allow_single_cluster`, `membership` `probabilities`, `cluster_selection_epsilon`, `medoids`)

josevalim:

PR #353, opened 2026-08-15
Fix Hierarchical hanging on non-finite dissimilarities

RicardoSantos-99:
## Bug

With NaN or infinity in the dissimilarity matrix, `fit/2` can hang forever.

Root cause: once only two clades remain and their distance is infinite (or NaN), argmin's
tie-breaking points them at each other asymmetrically (one points to itself, the other to
it), so neither is ever picked as the other's mutual nearest neighbor. No merge happens and
the loop that drives the algorithm never terminates.

This is otherwise impossible for finite input: the globally closest pair of clades is
always mutually nearest to each other, guaranteeing at least one merge every round. Caught
while working on HDBSCAN, which can produce infinite mutual reachability values.

## Fix

Track which clades are still alive. If a round makes zero progress, force a merge between
the two lowest-indexed ones still alive. This never triggers on finite input, so it's a
no-op there.

## Testing

4 new regression tests, each wrapped in a `Task` with an explicit timeout so a future
regression fails fast instead of hanging the suite: infinite distance between the last two
clades, a NaN coordinate, an all-NaN input, and a check that results on finite data are
byte-for-byte unchanged across every linkage. Full suite passes, 0 failures.

### DIFF
--- a/test/scholar/cluster/hierarchical_test.exs
+++ b/test/scholar/cluster/hierarchical_test.exs
@@ -280,6 +280,89 @@ defmodule Scholar.Cluster.HierarchicalTest do
     end
   end
 
+  describe "non-finite dissimilarities" do
+    # Regression tests for a real hang: with a non-finite dissimilarity (from :nan or
+    # :infinity), argmin's tie-breaking can point two clades at each other without either
+    # being picked as the other's mutual nearest neighbor, so no merge ever happens. This is
+    # impossible for finite dissimilarities, where the globally closest pair of clades is
+    # always mutually nearest to each other, guaranteeing at least one merge every round.
+    # When it happens the loop now stops instead of running forever, and the merges it could
+    # not make are reported as NaN dissimilarities with clade -1 and size 0.
+    #
+    # Every case is wrapped in a Task with an explicit timeout so a regression fails fast
+    # with a clear message instead of hanging the whole test run.
+    defp fit_within(data, opts, ms \\ 5_000) do
+      task = Task.async(fn -> Hierarchical.fit(data, opts) end)
+
+      case Task.yield(task, ms) || Task.shutdown(task, :brutal_kill) do
+        {:ok, result} -> result
+        nil -> flunk("Hierarchical.fit did not terminate within #{ms}ms")
+      end
+    end
+
+    test "an infinite dissimilarity between the last two clades reports an incomplete merge" do
+      # Point 3 is infinitely far from everyone. Once points 0, 1 and 2 (mutually close)
+      # have merged, only two clades remain: {0, 1, 2} and {3}, at distance infinity, and
+      # there is no finite distance left to justify merging them over any other pairing.
+      d =
+        Nx.tensor([
+          [0.0, 1.0, 1.0, :infinity],
+          [1.0, 0.0, 1.4142135623730951, :infinity],
+          [1.0, 1.4142135623730951, 0.0, :infinity],
+          [:infinity, :infinity, :infinity, 0.0]
+        ])
+
+      model = fit_within(d, dissimilarity: :precomputed, linkage: :single)
+
+      assert model.num_points == 4
+      # The two finite merges are made and kept, in ascending order.
+      assert Nx.to_flat_list(model.dissimilarities) |> Enum.take(2) == [1.0, 1.0]
+      # The merge that could not be made is reported instead of guessed.
+      assert Nx.to_number(Nx.is_nan(model.dissimilarities[-1])) == 1
+      assert Nx.to_flat_list(model.clades[-1]) == [-1, -1]
+      assert Nx.to_number(model.sizes[-1]) == 0
+    end
+
+    test "a NaN coordinate still makes every merge" do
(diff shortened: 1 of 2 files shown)


josevalim:
Tracking which ones is alive requires additional work though, no? I wonder if instead of making it work we could abort? For example, if the thing argmin points to is non-finite, we abort the loop?

I also assume the :tie_break option in argmin won't help us...

RicardoSantos-99:
> Tracking which ones is alive requires additional work though, no? I wonder if instead of making it work we could abort? For example, if the thing argmin points to is non-finite, we abort the loop?
> 
> I also assume the :tie_break option in argmin won't help us...

Might be overengineering this, so maybe simplify is better.

And switching argmin's tie_break to :high doesn't help, the same stuck pair just moves 


I agree aborting is the better fix. Tracking alive only exists to force a merge when we're stuck, and that merge doesn't actually mean anything (the two clades are stuck precisely because they're both non-finite from each other or from everything else). Better to stop the loop and raise than hand back a dendrogram whose top merge silently carries a meaningless height.

@josevalim I'll rework the PR to abort and raise instead, and drop the alive tracking entirely.

RicardoSantos-99:
Pushed, with one change from what I said: no raise, because it can't work here.

`fit/2` is a `deftransform`, so under `Nx.Defn.jit_apply` the whole thing is traced and "did the loop stall" is only known at runtime. Reading it with `Nx.to_number` to decide whether to raise blows up with "cannot invoke to_binary/2 on Nx.Defn.Expr", and breaks the two existing jit_apply tests. Hooks don't fit either since they're opt-in per call, so a plain `fit/2` wouldn't fire it.

So the loop still aborts as agreed, but the merges it couldn't make are reported instead: clades of `[-1, -1]`, sizes of 0, NaN dissimilarity, sorted to the end. alive is gone entirely and merge_clades is back to what it was.

test/scholar/cluster/hierarchical_test.exs
```diff
@@ -280,6 +280,89 @@ defmodule Scholar.Cluster.HierarchicalTest do
...
+    # Regression tests for a real hang: with a non-finite dissimilarity (from :nan or
+    # :infinity), argmin's tie-breaking can point two clades at each other without either
+    # being picked as the other's mutual nearest neighbor, so no merge ever happens. This is
+    # impossible for finite dissimilarities, where the globally closest pair of clades is
+    # always mutually nearest to each other, guaranteeing at least one merge every round.
+    # When it happens the loop now stops instead of running forever, and the merges it could
+    # not make are reported as NaN dissimilarities with clade -1 and size 0.
+    #
+    # Every case is wrapped in a Task with an explicit timeout so a regression fails fast
+    # with a clear message instead of hanging the whole test run.
```
josevalim: We don't need this. ExUnit tests have a built-in timeout and those will be trigger anyway!

test/scholar/cluster/hierarchical_test.exs
```diff
@@ -280,6 +280,89 @@ defmodule Scholar.Cluster.HierarchicalTest do
...
+
+    test "does not change the result on finite data" do
+      # Aborting only ever triggers when a round makes zero progress, which cannot happen
+      # for finite dissimilarities, so this must match plain `fit/2` exactly.
+      data = Nx.tensor([[1, 5], [2, 5], [1, 4], [4, 5], [5, 5], [5, 4], [1, 2], [1, 1], [2, 1]])
+
+      for linkage <- [:average, :complete, :single, :ward, :weighted] do
+        assert fit_within(data, linkage: linkage) == Hierarchical.fit(data, linkage: linkage)
+      end
+    end
```
josevalim: This test is not relevant I beleive.

test/scholar/cluster/hierarchical_test.exs
```diff
@@ -280,6 +280,89 @@ defmodule Scholar.Cluster.HierarchicalTest do
...
+
+    test "works with jit_apply" do
+      x = Nx.tensor([[0.0, 0.0], [1.0, 0.0], [0.0, 1.0], [:nan, :nan], [5.0, 5.0]])
+
+      assert Nx.Defn.jit_apply(&Hierarchical.fit/1, [x]) == Hierarchical.fit(x)
+    end
+  end
+
```
josevalim, line 327: Nor this one! jit only need to be tested once.

test/scholar/cluster/hierarchical_test.exs
```diff
@@ -280,6 +280,89 @@ defmodule Scholar.Cluster.HierarchicalTest do
...
+    test "an all-NaN input still makes every merge" do
+      x = Nx.broadcast(Nx.tensor(:nan), {6, 2})
+
+      model = fit_within(x, [])
+
+      assert model.num_points == 6
+      assert Nx.to_number(Nx.all(Nx.not_equal(model.clades, -1))) == 1
+      assert Nx.to_number(Nx.all(Nx.greater(model.sizes, 0))) == 1
+    end
```
josevalim: This test is probably redundant too.

josevalim:

PR #354, opened 2026-08-16
Fix precision loss and overflow in pairwise_squared_euclidean

RicardoSantos-99:
## Bug

`pairwise_squared_euclidean/1,2` compute `||x-y||^2` via the dot-product expansion `||x||^2 + ||y||^2 - 2x.y`. Once coordinates are large relative to the actual distance between points, this expansion suffers catastrophic cancellation: on f32, points ~1e5 apart in magnitude with a real distance of 1.0 come back as an outright 0.0. On f64, coordinates above ~1e150 overflow the intermediate squared norm and raise `ArithmeticError` instead of returning a distance. This affects every caller of the shared metric, including clustering and manifold algorithms.

## Fix

Shift `x` and `y` by the mean of their finite rows before the expansion. Squared distance is translation invariant, so the result is unchanged, but the intermediate norms and dot product now stay at the scale of the data spread instead of the raw coordinate magnitude, which removes the cancellation and the overflow. Integer inputs are left untouched since they're already exact. The mean is computed over finite rows only, so a placeholder non-finite row (e.g. the `:infinity` exemplar rows `AffinityPropagation` uses internally before pruning) can't poison every other row's distance through a shared Inf/NaN mean.

Caught while validating an unrelated HDBSCAN implementation on data with large coordinate magnitudes.

## Testing

Added regression tests for the cancellation and overflow cases, the non-finite sentinel row case, integer exactness, and `jit_apply`. Full test suite passes. Several other modules (KernelPCA, MDS, RadiusNN, Clustering, TSNE) pin exact float literals in their doctests/tests derived from this function; those are updated to the corrected values.


josevalim:

PR #355, opened 2026-08-20
Replace Hierarchical's round-based merge with nearest-neighbor chain

RicardoSantos-99:
`Hierarchical.fit` scales close to $O(n^3)$ in practice rather than the documented $O(\frac{n^2}{p} \cdot \log n)$.

The round-based merge assumes $O(\log n)$ rounds, which only holds while many clades are simultaneously each other's nearest neighbor. Once few enough clades remain, nearest-neighbor relationships form long chains instead of isolated mutual pairs, so it settles into one merge per round while every round still runs `Nx.argmin` over the whole matrix. That is $O(n)$ rounds at $O(n^2)$ each.

The practical effect is that the module is unusable past a few hundred points. `Scholar.Cluster.HDBSCAN` (#352) calls `Hierarchical.fit` internally and inherits the same ceiling.

This replaces the round-based merge with the nearest-neighbor-chain algorithm (Müllner 2011), which SciPy uses for this same family of Lance-Williams-updatable linkages. It extends a chain of nearest neighbors and merges as soon as the last two entries are each other's nearest, so each step scans one row instead of the whole matrix, and the number of steps is amortized $O(n)$ over the run. Measured with EXLA on random data:

| n | before | after |
|---|--------|-------|
| 400 | 1.5s | 0.13s |
| 800 | 28.8s | 0.13s |
| 1600 | 545.9s | 0.17s |

Reaching that in practice also depends on the pairwise matrix being updated in place across loop iterations, and two things quietly defeat it and bring the cubic behaviour back. Both are commented where they appear: reading a single cell of the matrix into another loop variable, and writing the merged clade's row before its column rather than after.

This also fixes a latent bug where remapping clade ids could leave a `clades` row in descending order when two merges tie, since the remapping is a permutation of ids and is not monotonic on ties.

Merge order at tied dissimilarities changes, as it has to with a different algorithm. The heights are the same; what differs is which of several equally valid merges happens first, which can shift clade ids and, where a cut falls inside a run of tied merges, the partition `labels_list/2` returns.

Verified against `scipy.cluster.hierarchy.linkage` over 600 configurations (5 linkages, 3 data shapes, 8 seeds, 3 sizes, f32 and f64): f64 matches exactly, and f32 deviates only in the cases where the implementation this replaces already deviates.

### DIFF
--- a/lib/scholar/cluster/hierarchical.ex
+++ b/lib/scholar/cluster/hierarchical.ex
@@ -15,11 +15,10 @@ defmodule Scholar.Cluster.Hierarchical do
     * `dissimilarity: :euclidean | :precomputed`
     * `linkage: :average | :complete | :single | :ward | :weighted`
 
-  Our current algorithm is $O(\\frac{n^2}{p} \\cdot \\log(n))$ where $n$ is the number of data points
-  and $p$ is the number of processors.
-  This is better than the generic algorithm which is $O(n^3)$.
-  It is also parallel, which means that runtime decreases in direct proportion to the number of
-  processors.
+  We use the nearest-neighbor-chain algorithm, which is $O(n^2)$ in both time and memory, where
+  $n$ is the number of data points.
+  This is better than the generic algorithm, which is $O(n^3)$, and is the best known bound for
+  this family of linkages.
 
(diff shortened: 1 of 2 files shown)


josevalim:

PR #356, opened 2026-08-21
Fix tied nearest neighbors merging a clade twice

RicardoSantos-99:
`Hierarchical.fit` can merge a clade twice when nearest neighbor distances tie. The result is a malformed tree: a clade ends up as the child of two different parents, the sizes stop adding up, and the root no longer gathers every point.

The nearest neighbor chain is only guaranteed to terminate on a mutual pair while the distance along it strictly decreases, and that is also what stops it from ever revisiting a clade it already holds. Ties break the guarantee. The chain ended only when `argmin` happened to return the previous entry, so on a tie it could walk past that entry and back onto a clade already on the chain. The duplicate then outlives the merge that consumes its other occurrence and is merged a second time, after it is already dead.

The chain now ends whenever the previous entry is *a* nearest neighbor of the tip, not only when it is the one `argmin` returns, which restores the strict decrease. Comparing the distances alone would miss a `NaN`, which is never equal to itself, so either condition ending the chain is enough. Writing past the end of the chain buffer is guarded too, since `Nx.indexed_put` clamps an out of range index rather than raising.

Introduced in #355. Measured over 200,000 random fits on integer coordinates, where distances tie constantly: 38 malformed trees before, 0 after. The regression test fails on the current implementation.

`linkage: :ward` also raises on the binary backend, and is fixed here too. Merged clades are masked out of the matrix rather than blanked, so their rows keep stale values that ward's update still reads, and its Lance-Williams term can come out negative for a dead column. Those columns are never used, but the square root of a negative raises where EXLA gives a silent NaN, so the term is clamped at zero. Also introduced in #355: 17 crashes in 600 random fits before, 0 after.

The chain guard also covers a third case from #355: an infinite coordinate puts both infinity and NaN in the distance matrix, the chain then extends without ever finding a mutual pair, and writing past the end of its buffer raised `index 4 is out of bounds` on every linkage. It now reports the merges it could not make, as the moduledoc describes.

### DIFF
--- a/lib/scholar/cluster/hierarchical.ex
+++ b/lib/scholar/cluster/hierarchical.ex
@@ -295,10 +295,12 @@ defmodule Scholar.Cluster.Hierarchical do
         chain_length = Nx.select(needs_start, 1, chain_length)
 
         # Extend the chain until its last two entries are mutual nearest neighbors.
-        # Bounded by n: a real chain can never revisit a clade before terminating,
-        # so exceeding n extensions only happens with a non-finite dissimilarity
-        # (from NaN or infinite input), where argmin's tie-breaking can land on an
-        # already-dead clade and loop without ever finding a genuine mutual pair.
+        # Bounded by n: the distance along a chain strictly decreases, so a chain
+        # can never revisit a clade before terminating, provided ties end it rather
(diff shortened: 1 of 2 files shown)


josevalim:

PR #358, opened 2026-08-22
Bump Nx to 0.13

RicardoSantos-99:
Bumps Nx and EXLA to 0.13

Draft while the doctests and test fixtures that shifted with the newer numerics get verified one by one against a reference implementation.


josevalim:
> Draft while the doctests and test fixtures that shifted with the newer numerics get verified one by one against a reference implementation.

FWIW, you only need to do so for the values that actually changed, as most changes are pretty printing related.

RicardoSantos-99:
> > Draft while the doctests and test fixtures that shifted with the newer numerics get verified one by one against a reference implementation.
> 
> FWIW, you only need to do so for the values that actually changed, as most changes are pretty printing related.

Yeah, that’s what I did for most of them. A handful looked a bit off though, so I dug into those specifically and ended up finding a few real bugs, unrelated to the bump itself. They were just numerical issues that got exposed by the newer SVD path. I fixed those too (`LinearRegression`, `PLSSVD`, `RidgeRegression`) and verified the results against sklearn and Scholar’s own solvers. The PR is out of draft now.

RicardoSantos-99:
While reviewing some of these modules, I found a few other bugs as well, but I didn’t include them in this PR since they’re unrelated to the bump. I still need to investigate them a bit further, and I’ll either open a separate PR or create an issue for them.

josevalim:

PR #359, opened 2026-08-23
Fix NaiveBayes prediction paths and prior dtype

RicardoSantos-99:
Five bugs in the NaiveBayes modules, four on the prediction path.

- BernoulliNB never applies `:binarize` at predict time. The threshold is not kept on the model, so every `predict*` passes raw `x` into a likelihood that assumes binary features. `:binarize` defaults to `0.0`, so this hits the default configuration.
- CategoricalNB's `check_dim` reads axis 1 of `feature_count`, which is the class count here (the tensor is `{num_features, num_classes, num_categories}`, not 2-D as in the other modules). It rejects valid input and accepts invalid input whenever `num_features != num_classes`.
- CategoricalNB's `joint_log_likelihood` sizes its accumulator from `x` rather than from the class axis, so predict raises for any dataset where the two differ. Every doctest in the module uses `num_features == num_classes`, the one shape where this is invisible.
- CategoricalNB's `:min_categories` is dead code. `Keyword.pop` rebinds `opts` before the value is read, so the category count always falls back to being inferred from the data.
- `class_log_priors` is computed in f32 regardless of input type, while `class_count` and `feature_log_probability` follow the input. The block is identical in all four modules, so all four are fixed.

`:alpha` also gets a non-negative check through the existing `Scholar.Options.non_negative_number`. A negative value previously produced NaN with no error.

Behaviour changes: `%Bernoulli{}` gains a `:binarize` field; BernoulliNB predictions change for input that is not already binary, and its doctests are updated; `fit/3` raises on a negative `:alpha`; CategoricalNB models built with `:min_categories` change shape.

### DIFF
--- a/lib/scholar/naive_bayes/bernoulli.ex
+++ b/lib/scholar/naive_bayes/bernoulli.ex
@@ -10,13 +10,14 @@ defmodule Scholar.NaiveBayes.Bernoulli do
   import Scholar.Shared
 
   @derive {Nx.Container,
+           keep: [:binarize],
            containers: [
              :feature_count,
              :class_count,
              :class_log_priors,
              :feature_log_probability
            ]}
-  defstruct [:feature_count, :class_count, :class_log_priors, :feature_log_probability]
+  defstruct [:feature_count, :class_count, :class_log_priors, :feature_log_probability, :binarize]
 
   opts_schema = [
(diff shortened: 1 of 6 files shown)


josevalim:

PR #363, opened 2026-08-26
Fix StandardScaler discarding the mean on a constant feature

RicardoSantos-99:
`StandardScaler` zeroed out the mean for any zero-variance feature instead of keeping it, so a constant feature came out uncentered instead of 0.0. Worse on new data: transforming an unseen value through a constant training column just passes it through raw instead of centering it, which is silent data leakage into a pipeline that assumes standardized input.

sklearn only clamps the scale to 1 for a constant feature, it never touches the mean. Removed the line that zeroed it, the existing division-by-1 guard was already enough.


josevalim:

PR #364, opened 2026-08-27
Fix KNNImputer wrong donors for multi-NaN rows and f64 crash

RicardoSantos-99:
`KNNImputer` computes distance to a candidate donor by zeroing out only the single column currently being filled. Any other missing value in that same row is left as NaN, so it poisons the distance sum for every donor. Every candidate ends up at infinity, and picking "nearest" from a constant array just returns the first `num_neighbors` rows in the tensor. So any row with two or more missing values gets imputed from the first rows in the data instead of its real nearest neighbors, silently.

Fixed by excluding any column where either row has NaN, not just the one being filled, matching the present-features weighting `sklearn.impute.KNNImputer`'s `nan_euclidean_distances` uses. Verified against it directly: on a case where the true nearest donors sit far from the ones the old code always picked, the old code returns `[1.05, 2.05, 5.0]`, this returns `[55.0, 85.0, 5.0]`, matching sklearn exactly.

Also fixed a separate bug found while testing: the distance accumulator and the NaN placeholder both hardcoded f32, so an f64 tensor crashed as soon as a value needed imputing. Both now take their type from the input.

### DIFF
--- a/lib/scholar/impute/knn_imputter.ex
+++ b/lib/scholar/impute/knn_imputter.ex
@@ -199,17 +199,12 @@ defmodule Scholar.Impute.KNNImputter do
     # minus nan column
     coordinates = coordinates - 1
 
-    # a donor with no value in nan_col cannot be used to fill it - this also
-    # rules out the row itself as its own neighbor, since the row's value at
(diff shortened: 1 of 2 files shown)


josevalim:

PR #366, opened 2026-08-28
Expose OPTICS's reachability graph

RicardoSantos-99:
`Scholar.Cluster.OPTICS.fit/2` computes the full reachability graph but only keeps the labels. `compute_optics_graph` builds `core_distances`, `reachability`, `predecessor` and `ordering`, then `fit_p` feeds them into `cluster_optics_dbscan` and discards them at the end of the function.

That graph is not an internal detail. `ordering` against `reachability` is the reachability plot, the standard way OPTICS results are read and the way clusters are told apart from noise visually, and it is what `predecessor` and `core_distances` support (tracing how a point was reached, and at what radius it became a core point). None of it survives past `fit/2` today, so nothing built on Scholar can draw one.

This adds `:reachability`, `:core_distances`, `:ordering` and `:predecessor` to the struct, populated the same way `:labels` already is.

### DIFF
--- a/lib/scholar/cluster/optics.ex
+++ b/lib/scholar/cluster/optics.ex
@@ -10,8 +10,29 @@ defmodule Scholar.Cluster.OPTICS do
   """
   import Nx.Defn
 
-  @derive {Nx.Container, containers: [:labels, :min_samples, :max_eps, :eps, :algorithm]}
-  defstruct [:labels, :min_samples, :max_eps, :eps, :algorithm]
+  @derive {Nx.Container,
+           containers: [
+             :labels,
+             :min_samples,
+             :max_eps,
+             :eps,
+             :algorithm,
+             :reachability,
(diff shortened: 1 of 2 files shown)


RicardoSantos-99:
Would it make sense to cut a new release soon? maybe v0.5.0

josevalim:
We can do a new release soon.

For this PR, can we improve the tests? They are converting too many values and asserting on shapes. Ideally we want to assert on values and rely Nx.all_close if possible.

josevalim:
Sorry, assert_all_close!

RicardoSantos-99:
> We can do a new release soon.
> 
> For this PR, can we improve the tests? They are converting too many values and asserting on shapes. Ideally we want to assert on values and rely Nx.all_close if possible.

Done

josevalim:

PR #367, opened 2026-08-29
Fix bugs in Scholar.Manifold.Trimap

RicardoSantos-99:
- **`in1d/2` returns its mask permuted.** It applies the sort permutation again
  instead of inverting it. Wrong on 167 of 300 random inputs.
- **Three or four points raise.** The local scale reads a fixed
  `distances[[.., 3..5]]` slice, assuming at least six neighbor columns.
- **Anything wider than 100 features raises.** The PCA branch passed
  `full_matrices:` to a private SVD that takes `full_matrices?:`. Behind the
  crash it also reconstructed instead of projecting, kept 101 components instead
  of 100, and rebound the result inside the `case` so it never reached the
  initialization.
- **`num_random: 0` is rejected.** `generate_triplets/3` branches on
  `num_random > 0`, but the option is typed `:pos_integer`.
- **Impossible outlier sampling hangs forever.** `rejection_sample/4` never
  repeats a draw and has no iteration bound. The defaults want 50 outliers, so
  any dataset under 61 points can hang with no output.
- **`init_embeddings` silently overrides `num_components`.** Three columns while
  asking for two returned a three column embedding.
  
  
One behaviour change worth calling out: some configurations that returned a
result on main now raise. `rejection_sample/4` fills `num_inliers *
num_outliers` slots from the pool of points outside an anchor's neighbourhood,
never repeating a value, so when the request is wider than that pool it is
unsatisfiable by pigeonhole. Those runs only completed by emitting duplicate
outliers, and with `in1d/2` fixed they hang instead, which is why the guard
rejects them up front.

The PCA projection matches an exact numpy SVD to 2.9e-13 in f64. Every fix has a
test that fails when only that fix is reverted. Full suite: 343 doctests, 662
tests, 0 failures

### DIFF
--- a/lib/scholar/manifold/trimap.ex
+++ b/lib/scholar/manifold/trimap.ex
@@ -498,6 +498,13 @@ defmodule Scholar.Manifold.Trimap do
               "outside an anchor's own neighborhood, got: #{opts[:num_inliers] * opts[:num_outliers]}"
     end
 
+    if Nx.rank(init_embeddings) != 0 and
+         Nx.axis_size(init_embeddings, 1) != opts[:num_components] do
+      raise ArgumentError,
+            "init_embeddings must have num_components (#{opts[:num_components]}) columns, " <>
+              "got: #{Nx.axis_size(init_embeddings, 1)}"
+    end
+
     unless (Nx.rank(triplets) == Nx.rank(weights) and Nx.rank(triplets) == 0) or
              (Nx.rank(triplets) == 2 and Nx.rank(weights) == 1 and
                 Nx.axis_size(triplets, 0) == Nx.axis_size(weights, 0)) do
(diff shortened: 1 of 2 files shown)


RicardoSantos-99:
One thing I deliberately left out of this PR is an issue around the interaction
between Trimap's `metric` and `knn_algorithm` options.

While testing the different combinations, I found that some combinations exposed
by the current schemas cannot run successfully:

| metric | `:brute` | `:nndescent` | `:large_vis` |
| --- | --- | --- | --- |
| `:euclidean` | ok | ok | requires sufficiently large n |
| `:squared_euclidean` | ok | ok | requires sufficiently large n |
| `:cosine` | ok | ok | rejected |
| `:manhattan` | ok | ok | rejected |
| `:chebyshev` | rejected | ok | rejected |

`:auto` appears to avoid these cases because it selects the KNN implementation
automaticaly.

I have not included a fix for this in the PR because I still need to investigate
why these combinations behave differently and where the validation or
normalization should live.

josevalim:

PR #368, opened 2026-09-01
Fix bugs in Scholar.Linear.IsotonicRegression

RicardoSantos-99:
- **A decreasing fit can come back non-monotonic.** `contiguous_isotonic_regression/4`
  writes each pooled value back over the block it covers, but that pass stops at
  `max_size`, which is only where the valid range ends when fitting increasing.
  Fitting decreasing reverses the input, so the placeholder slots left by tied x
  values move to the front and the range ends at `y_size` instead. Part of the
  last block keeps its unpooled value:
  `fit([0, 0, 1, 1, 2], [4, 4, 5, 5, 3], increasing: false)` returned thresholds
  `[4.0, 4.5, 3.0]`, which rises where the caller asked for a decreasing fit.
- **A zero weight poisons the whole fit.** A sample with weight zero still opens
  or joins a group, so a group made only of those divides by zero. The NaN then
  spreads through the pooling: `sample_weights: [1, 0, 1, 1, 1]` returned
  thresholds of `-infinity` and predicted NaN everywhere. scikit-learn masks
  non-positive weights out before fitting.
- **`out_of_bounds` was never read.** It appears once in the whole module, in the
  schema. `predict/2` always clipped, so `:nan`, the documented default, behaved
  exactly like `:clip` and the default behaviour did not exist.
- **`preprocess/1` raises on models with fewer than three thresholds.** The
  duplicate trim slices `y[1..-2//1]`, which needs three. Constant targets or a
  two point fit produce two, and the range `1..0` raises.
- **`cutoff_index` goes stale.** The trim drops thresholds but the index keeps
  pointing into the untrimmed tensors, so running `preprocess/1` on its own
  output slices past the end and raises.
- **`fit/3` rejects the column shape its own check allows.** `check_input_shape/1`
  documents `{n, 1}` as valid and `predict/2` accepts it, but `fit/3`
  destructures the shape as `{n_samples}` and raises a MatchError first.
- **`increasing: :auto` follows a least squares slope.** One outlier flips it.
  scikit-learn uses the sign of Spearman's rho, which reads ranks and does not
  move with the size of an outlier. On `x = 1..10` with `y = 1..9` then `-100`,
  the slope is -5 and picks decreasing while rho stays at 0.45 and picks
  increasing.

### DIFF
--- a/lib/scholar/linear/isotonic_regression.ex
+++ b/lib/scholar/linear/isotonic_regression.ex
@@ -548,10 +548,27 @@ defmodule Scholar.Linear.IsotonicRegression do
     if increasing, do: y, else: Nx.reverse(y)
   end
 
+  # sign of Spearman's rho, matching sklearn; a constant input makes it undefined
+  # and the comparison then picks decreasing
   defnp check_increasing(x, y) do
-    x = Nx.new_axis(x, -1)
-    y = Nx.new_axis(y, -1)
-    model = Scholar.Linear.LinearRegression.fit(x, y)
-    model.coefficients[0] >= 0
+    rank_x = average_rank(Nx.flatten(x))
+    rank_y = average_rank(Nx.flatten(y))
+
+    centered_x = rank_x - Nx.mean(rank_x)
+    centered_y = rank_y - Nx.mean(rank_y)
+
+    rho =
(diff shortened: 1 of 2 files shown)


josevalim:

PR #369, opened 2026-09-01
Add mean shift clustering

RicardoSantos-99:
Closes the Mean Shift item on #135.

Mean shift moves every seed towards the mean of the samples within one
bandwidth of it until the seed settles, then keeps the strongest of the seeds
that landed on the same mode. The number of clusters comes out of the data
rather than being given, which is what the bandwidth controls.

```elixir
iex> x = Nx.tensor([[1, 1], [1.2, 1.1], [8, 8], [8.1, 8.2]])
iex> model = Scholar.Cluster.MeanShift.fit(x, bandwidth: 1.0)
iex> Scholar.Cluster.MeanShift.prune(model).labels
#Nx.Tensor<
  s32[4]
  [1, 1, 0, 0]
>
```

## Shape of the implementation

Every seed moves at once rather than one at a time, so an iteration is a single
pairwise matrix instead of one per seed. That costs `O(seeds * samples)` of
space, the same shape `DBSCAN` already pays through `radius_neighbors`. A
k-d tree would not help here: the query mean shift needs is "every sample
within the bandwidth", whose result size is not known until it runs, and
`Scholar.Neighbors.KDTree` only answers k nearest anyway.

`fit/2` keeps one row per seed so the shapes stay static, marking the centers
that lost with `:infinity` and reporting how many survived, and `prune/1` drops
them. This follows `Scholar.Cluster.AffinityPropagation`, which has the same
problem of not knowing the cluster count until it has run. `prune/1` raises
rather than building an empty tensor when nothing survived.

## Validation

Checked against scikit-learn 1.6.1 on 131 datasets: 11 hand-written edge cases
and 120 random ones spanning 2 to 25 samples, 1 to 4 features, bandwidths from
0.3 to 6.0, and 1 to 21 resulting clusters. Labels, centers and cluster counts
match on all of them, on the default backend and under EXLA.

That includes the order scikit-learn puts the centers in, which sorts by how
many samples a center gathered and breaks ties on the coordinates. The edge
cases cover a bandwidth wide enough to hold everything, one narrower than the
closest pair, samples that are all the same point, a single sample, one
feature, samples sitting exactly at the bandwidth, and seeds that never reach a
sample, which scikit-learn drops and so does this.

`max_iterations` and `cluster_all` each have a test where the option changes
the answer, and there is a `jit_apply` test.

## Left out on purpose

`bin_seeding` and `estimate_bandwidth` are not here. Binning produces a number
of seeds that is only known at runtime, which does not fit a static shape, so
the cheap way in is the `:seeds` option, the way `Trimap` accepts precomputed
triplets. Both are worth a follow-up issue.

I added `HDBSCAN` and `OPTICS` because were never
added to `groups_for_modules`, so ex_doc filed them outside the Models group.


josevalim:

PR #370, opened 2026-09-15
chore: Move to Nx 1.0

georgeguimaraes:
Nx 1.0.0 is out and `nx ~> 0.13` keeps it out of the resolver. This widens it to `~> 0.13 or ~> 1.0` and moves the lock to 1.0.0.

Nx gets `override: true` because polaris 0.1.0 pins `nx ~> 0.5` and there's no polaris release that allows 1.0 yet. Only affects Scholar's own build. It can go once polaris ships a release with a wider nx requirement.

Six doctests had exact float values from SVD, ridge regression and truncated SVD that move in the sixth digit with the linalg fixes in Nx 1.0, so they're updated to the new outputs.


josevalim:
@georgeguimaraes stb_image is out, can we perhaps remove override: true?

georgeguimaraes:
Tried it, and it's actually polaris that blocks it, not stb_image. Dropping the override fails resolution because polaris 0.1.0 pins `nx ~> 0.5`, and polaris main still has the same pin. stb_image is only an optional dep of scidata and isn't even in Scholar's lock, so my description got the reason wrong. Sorry about that.

So the override stays until there's a polaris release that allows `~> 1.0`. Happy to drop it here the moment that's out.

polvalente (approved):

